PR #51, opened 2012-06-02
asynchronous git prompt updates

robertream:
these changes speed up the git prompt, with the side effect that the prompt will be eventually consistent

Start-Prompt turns the asynchronous prompt on
Stop-Prompt turns the asynchronous prompt off


dahlbyk:
Between #319 and #208 we should be in a better spot performance-wise. I'm going to go ahead an close this, but do keep an eye out for an upcoming release with those improvements and leave a note on #78 if you're still having issues.

I do appreciate the effort you put into this several years ago!

PR #58, opened 2012-06-08
set path prompt maximum size=30

rismoney:
when you dive deep into a repo the prompt can get infinitely larger by default.  this puts a cap of 30 chars on the prompt to retain readability and avoid posh wraparound.

### DIFF
--- a/profile.example.ps1
+++ b/profile.example.ps1
@@ -14,7 +14,12 @@ function prompt {
 
     # Reset color, which can be messed up by Enable-GitColors
     $Host.UI.RawUI.ForegroundColor = $GitPromptSettings.DefaultForegroundColor
-
+	$maxpromptsize=30
+	$promptpath=$pwd.path
+	if ($promptpath.length -ge $maxpromptsize)
+	{
+		$pwd = "..." + $promptpath.substring($promptpath.length - $maxpromptsize + 4)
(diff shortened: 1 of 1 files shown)


dahlbyk:
@rismoney so now that I've let this sit forever... are you still interested in this or should I close? My attempt to make this a little smarter is still on a branch: https://github.com/dahlbyk/posh-git/compare/gh58. In particular I think we would have a better experience if the max size were to kick in at directory boundaries, e.g. `C:\...\my-really-long-project-directory-name` or `C:\...\work\project-directory-name`.

The other thing I didn't try to solve is handling UNC paths (see #75) reasonably, e.g. `\\server\share\...\project-directory`.

-----------------------------------

> @dunckr: If anyone prefers more minimalist, I use only current directory as my prompt:

This in particular seems like it might be a reasonable settings option. Or...what if we just added this default:

```powershell
$GitPromptSettings.PromptPath = { $pwd }
```

And then you would just have to add this to your `$PROFILE`:

```powershell
$GitPromptSettings.PromptPath = { (Get-Item $pwd).Name }
```

RandomlyKnighted:
I'd love to see this finished. Most of my directories where my repos are stored have rather long directory names so it'd be nice have this limit.

PR #67, opened 2012-08-22
Adding svn prefix to branch name if svn clone

cromwellryan:
Hard check for .git\svn propogates the existance of that directory to Write-GitStatus using IsSvn.

I can imagine this evolving to support git-tf/git-tfs with a folder/prefix map.  If you'd rather go all in on that pattern, let me know.


dahlbyk:
This obviously slipped way off my radar and hopefully most SVN/TFS users have been pushed into Git anyway. Thanks for the effort, though!

PR #80, opened 2013-02-13
Added shortcut aliases for commonly used git commands.

codito:


rkeithhill:
I think the aliases would be best provided by a new module say `git-aliases`.  While the aliases are less to type there are a lot that are fairly cryptic.  It would easier to just use tab completion than to learn all of these.  And some of these aliases are likely going to interfere with folks' personal aliases.  There are a couple that would stomp over my aliases.

PR #94, opened 2013-04-04
Example of integrating help

wekempf:
This pull request is not complete, but provides an example of how PowerShell help can be integrated directly in the code. I would volunteer to do all the work, but I'm unfamiliar with the code and frankly don't understand what some of the commands are ore why they are publicly exported? With this example merged you can do the following:

PS> help about_posh-git
PS> help Add-SshKey -Full

### DIFF
--- a/GitUtils.ps1
+++ b/GitUtils.ps1
@@ -189,6 +189,10 @@ function InDisabledRepository {
     return $false
 }
 
+<#
+.SYNOPSIS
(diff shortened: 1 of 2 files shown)


rkeithhill:
An about topic was added back in May but we haven't shipped an update since then.  I have merged your Add-SshKey help into GitUtils.ps1.

Issue #78, opened 2013-02-09
posh-git is still too slow...


tgoeke:
I'm trying out posh-git, but it's still too slow.  How do I uninstall it?

ghostsquad:
I'm guessing that this still has not been implemented?

dahlbyk:
> I'm guessing that this still has not been implemented?

You guess correctly. However, #208 just dropped and it's looking very promising!

edclement:
removing this line:
`$stashCount = $null | git stash list 2>$null | measure-object | select -expand Count`
from GitUtils.ps1 solved the issue for me. It cut 3 seconds off the prompt return time. Considering that git stash status is disabled by default with posh git, I don't understand why this command is always run anyway....

ghostsquad:
the pipeline is slow. Any areas that can be converted to use standard loops and .NET classes would significantly improve performance.

Mellbourn:
@edclement excellent workaround! That increased the performance for me from unacceptable to no problem at all.

DoCode:
posh-git is still slow on tab completion!
Guys, we must investing in libgit2sharp integration!

Tavriets:
Adding ```--untracked-files=no``` (or just ```-uno```) to the ```git stash list``` mentioned by @edclement makes it much faster for me.

ghostsquad:
@edclement - not sure if your workaround is even valid. if StashStatus is disabled, that line won't execute.

edclement:
@ghostsquad - removing that line from the 0.4.0 release (the latest tagged version) definitely fixed the issue for me. That was the only thing I changed.
looking at the code in that tagged version:
```
if($settings.EnableFileStatus -and !$(InDisabledRepository)) {
    dbg 'Getting status' $sw
    $status = git -c color.status=false status --short --branch 2>$null
    #$stashCount = $null | git stash list 2>$null | measure-object | select -expand Count
} else {
    $status = @()
}
```
I don't see why that command wouldn't execute...

ghostsquad:
oh oh, sorry, we were looking at different versions.

serbrech:
@edclement :+1:  worked for me, this makes it good enough to work with for me too

dahlbyk:
Between #319 and #208 we should be in a better spot performance-wise. I'm going to go ahead an close this, but do keep an eye out for an upcoming release with those improvements and leave a note if you're still having issues.

Ramin2c:
[...]

Issue #50, opened 2012-06-01
Posh-git is too slow


ghost:
I installed GitHub for Windows, which include posh-git. When I use the included GitHub Powershell link, it loads posh-git.

Just loading the posh-git part in a new Powershell window is very slow, taking 8 seconds. Pressing the "enter" key on an empty line takes 3 seconds to show a new prompt.

Here's a video (needs Flash, no audio):
https://www.Scolab.com/Public/Jing/VideoPlayer/Jing.aspx?filename=../CM/slowposhgit2012-06-01_1507.mp4&width=1224&height=672

The machine is quite fast (Intel i7 laptop, dual-core with SSD hard disk). The project  weight 726 MB and has 2500 folders and 16300 files, excluding the ".git" folder.

I can live with the slow loading, but not with the wait. Could it disable itself or limit itself if it detects that it's taking more than 500ms to show?

Thanks a lot,
Carl

PR #104, opened 2013-06-20
initial code to show how async 'git status' prompt can work

jamesmanning:
This is intended more for discussion than being merged 'as-is' - if/when it were to be merged, the common bits of functionality in the sync and async code paths should be factored out (and maybe the sync case being the async code path with an added wait on the process to exit, but that's not really relevant right now).

The approach taken here intentionally doesn't use any of the other work (like caching status, filesystemwatcher, etc) - not that it couldn't be used with those, but more to keep it as simple as possible for now for a hopefully smaller diff and easier discussion about whether it's a good fit for posh-git.

The logic/data flow isn't much different than the existing synchronous prompt - it's controlled in this diff by a new bool in the GitPromptSettings (EnableAsyncFileStatus) for easier testing and comparison.

The key 'breakthrough' WRT earlier discussions in issues like #52 was finding that the RawUI interface didn't actually require cursor location manipulation in order to write to the screen buffer in a place other than wherever the cursor was already located.  Instead, it supports writing a 2d rectangle to the buffer at a specified coordinate.  This means we don't have to worry about the various bad states we might get into if we were manually moving the cursor and writing strings out to the buffer while the user was typing at their prompt (yay!)

[PSHostRawUserInterface.SetBufferContents Method (Coordinates, BufferCell[,])](http://msdn.microsoft.com/en-us/library/windows/desktop/ms568430.aspx)

The main new/entry function to look at it is Write-GitStatusPromptAsync in the new file GitAsyncPrompt.ps1

in the case of EnableAsyncFileStatus -eq $true (and we're in a repo that hasn't been disabled), it does:
- saves the current cursor position to know where it will write the multi-colored status string once it's done being calculated
- writes out a 'placeholder' string to the host on its own line (not strictly necessary, just to give the user an idea taht there's an ongoing process happening, even if it's only going to take a few seconds)
- run the same "git status" cmdline as usual, just doing so as a System.Diagnostics.Process
- gathers stdout lines via OutputDataReceived event on the process
- when Exited event happens on the process:
  - takes the stdout lines, converts to the same object already being stored in $global:GitStatus by the existing prompt code, and also writes the object to that global variable to maintain that behavior
  - calls Get-GitStatusBufferCells to convert that object into the set of buffer cells (multi-colored string, currently always 1 row tall just like the existing synchronous code)
  - calls Write-BufferCellsToPosition to write those buffer cells to the screen buffer (correct location was grabbed/stored earlier)

IME it runs 'fast enough' that even if it hit enter ~10 times, resulting in ~10 git status calls running, they all still complete very quickly (and they each write to their correct screen buffer location).

Does the idea/approach make sense for posh-git?

Thanks!

Related issues about perf of git status:
- #78 - posh-git is still too slow...
- #52 - Cache Get-GitStatus results to speed up the prompt
- #51 - asynchronous git prompt updates
- #50 - Posh-git is too slow

### DIFF
--- a/GitAsyncPrompt.ps1
+++ b/GitAsyncPrompt.ps1
@@ -1,0 +1,220 @@
+function Write-GitStatusPromptAsync([string]$gitDir = (Get-GitDirectory)) {
+    $processStartInfo = New-Object Diagnostics.ProcessStartInfo
+    $processStartInfo.FileName = "git"
+    $processStartInfo.Arguments = "-c color.status=false status --short --branch"
+    $processStartInfo.CreateNoWindow = $true
+    $processStartInfo.RedirectStandardOutput = $true
+    $processStartInfo.RedirectStandardError = $false
+    $processStartInfo.UseShellExecute = $false
+    $processStartInfo.WorkingDirectory = $pwd
+
+    $process = New-Object Diagnostics.Process
+    $process.StartInfo = $processStartInfo
+    
+    $processOutputLines = @()
+    $positionToWriteStatus = $Host.UI.RawUI.CursorPosition
+    
+    # Output a placeholder line where the status will be written to.  This makes sure the status line doesn't overwrite the actual prompt
+    Write-Host " [...processing...]" 
+
+    $messageData = new-object psobject -property @{
+        gitDir = $gitDir
+        processOutputLines = $processOutputLines
+        positionToWriteStatus = $positionToWriteStatus
+    }
+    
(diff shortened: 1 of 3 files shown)


dahlbyk:
Between #319 and #208 we should be in a better spot performance-wise. I'm going to go ahead an close this, but do keep an eye out for an upcoming release with those improvements and leave a note on #78 if you're still having issues.

I do appreciate the effort you put into this several years ago!

Issue #108, opened 2013-07-23
Tab completion for tags


Haacked:
I went to checkout a tag and typed:

```
git checkout tags/
```

And started hitting `{tab}` hoping it would cycle through my tags but it didn't. Would be a nice touch to add that.

joshgo:
Also, tag completion doesn't seem to work with the `push` command:

`git push origin v`   then hit {tab key} and nothing

Note, the **v** is there because all our version tags start with a **v** ... semver.

rkeithhill:
@Haacked  This works for checkout but you don't specify `tags/`.  In fact, when I try `git checkout tags/<checkout` in Bash, that doesn't auto-complete either.  What does work in both Bash and posh-git is: `git checkout v<tab>` for the semver approach where you prefix with a `v`.

@joshgo `git push origin v<tab` works for me with the latest bits.

I'm going close this.  Feel free to re-open if you still think there is an issue here.

PR #128, opened 2014-01-02
Implement tab completion for tags.

jsenich:
Here's my initial attempt at implementing tab completion for tags (Issue #108).  It doesn't yet recognize trying to complete using "tags/" like described in the issue, but cycles through a combined list of branches and tags similar to how this is handled in the bash prompt implementation.

### DIFF
--- a/GitTabExpansion.ps1
+++ b/GitTabExpansion.ps1
@@ -74,6 +74,12 @@ function script:gitBranches($filter, $includeHEAD = $false) {
         foreach { $prefix + $_ }
 }
 
+function script:gitTags($filter) {
+    git tag |
+        where { $_ -like "$filter*" } |
+        foreach { $_ }
+}
+
 function script:gitFeatures($filter, $command){
 	$featurePrefix = git config --local --get "gitflow.prefix.$command"
     $branches = @(git branch --no-color | foreach { if($_ -match "^\*?\s*$featurePrefix(?<ref>.*)") { $matches['ref'] } }) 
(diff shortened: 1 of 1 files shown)


GitTabExpansion.ps1
```diff
@@ -74,6 +74,12 @@ function script:gitBranches($filter, $includeHEAD = $false) {
         foreach { $prefix + $_ }
 }
 
+function script:gitTags($filter) {
+    git --git-dir="$(Get-GitDirectory)" for-each-ref --format='%(refname:short)' refs/tags |
```
dahlbyk: As we celebrate the anniversary of this comment... do you want to take a crack at updating this or should I?

jsenich:
Modified my commit to use the simpler git tag rather than manually reading the refs.

GitTabExpansion.ps1
```diff
@@ -74,6 +74,12 @@ function script:gitBranches($filter, $includeHEAD = $false) {
         foreach { $prefix + $_ }
 }
 
+function script:gitTags($filter) {
+    git tag |
+        where { $_ -like "$filter*" } |
+        foreach { $_ }
```
JayBazuzi: What is this line for?

jsenich: It provides tab completion based on tag names and is called from GitTabExpansion().

JayBazuzi: I'm asking specifically about the line `foreach { $_ }`.

Try this: remove that line, try TAB-completion on tags, and see if it still works.

jsenich: Good call. That line wasn't  necessary.

dahlbyk:
Thank you!

For future reference, watchers won't get a notification on `push` to a PR, so you may want to bump a thread to which you've pushed a fix. :sparkles:

PR #139, opened 2014-05-08
Limit length of long branch name

dbuckingham:
I wished to limit the length of the branch name (in the prompt) for branches with very long (verbose) names.  I arbitrarily chose the default to be 20 characters because it seemed to show enough of the branch name when using a Git-Flow like convention, with a JIRA id:  "feature/ABC-123-name...".


dahlbyk:
Not sure I ever noticed you had pushed the ellipsis fix! Sigh.

Thanks for this!

dbuckingham:
No problem!  Thanks for the merge!

PR #140, opened 2014-05-13
Add function to create shortcut alias's that still support tab expansion

bradphelan:


dahlbyk:
Similar to #80, I'm inclined to leave stuff for supporting PowerShell aliases out of this module. It doesn't look like this specifically depends on anything internal to posh-git.

Issue #114, opened 2013-09-16
Try to avoid ssh-agent warning


dahlbyk:
Resolving "ssh-agent not found" is a FAQ and there's really no reason not to try to avoid it... the vast majority of the time its location can be resolved from the location of `git.exe` (specifically, "..\bin\ssh-agent.exe").

theaquamarine:
Implemented in #141, closing.

PR #141, opened 2014-05-28
Guess agent location if not on path

theaquamarine:
Implements #114 for ssh-agent and pageant

### DIFF
--- a/GitUtils.ps1
+++ b/GitUtils.ps1
@@ -262,6 +262,16 @@ function Get-SshAgent() {
     return 0
 }
 
+# Attempt to guess Pageant's location
+function Guess-Pageant() {
+    Write-Verbose "Pageant not in path. Trying to guess location."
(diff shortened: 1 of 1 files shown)


giggio:
@dahlbyk Do you have an idea of when you plan to work on this?

rpavlik:
Will this get merged soon? After adding aliases to my profile as a workaround:

```powershell
# Aliases to find ssh-agent/ssh-add to quiet down posh-git - http://stackoverflow.com/a/24368751/265522
Set-Alias ssh-agent "${env:ProgramFiles(x86)}\git\bin\ssh-agent.exe"
Set-Alias ssh-add "${env:ProgramFiles(x86)}\git\bin\ssh-add.exe"
```

posh itself seems to start a lot faster now - was taking several seconds before with that error message.

giggio:
This PR will have its first birthday soon.

giggio:
Tks!

PR #148, opened 2014-07-04
Small updates to SSH agent setup

Ventajou:
- Saving SSH Agent environment to user level so other apps may use it such
  as cmd.exe.
- Adding all keys found in the user's .ssh folder instead of just the
  default one.

Makes it easy to use several git servers with different ssh keys.


dahlbyk:
Closing in deference to #338 and the other changes that this now conflicts with.

PR #150, opened 2014-07-14
Added support for pipelining to Get-Git* cmdlets

luolong:
I needed to get the status of multiple git projects, but found out that posh-git codlets were woefully unable to be pipelined, so I added few attributes and updated the implementations to support pipelining of `Get-GitDirectory` and `Get-GitStatus` cmdlets.

Also, I started a rudimentary documentation of the cmdlets.

Please take a look at this patch and see if I have not obviously broken anything.

### DIFF
--- a/GitUtils.ps1
+++ b/GitUtils.ps1
@@ -1,11 +1,45 @@
 # Inspired by Mark Embling
 # http://www.markembling.info/view/my-ideal-powershell-prompt-with-git-integration
 
+<#
+.SYNOPSIS
+    Get the path of the .git directory
+
+.DESCRIPTION
+    Returns the location of the .git directory. If environment variable GIT_DIR 
+    has been defined, the return value will be the content of GIT_DIR.
+    This command can be piped and it will return 
+    If one or more 
+
+.PARAMETER Path
+    Path of the location to start searching from.
+
+.LINK
+    http://git-scm.com/blog/2010/04/11/environment.html
+
+#>
 function Get-GitDirectory {
-    if ($Env:GIT_DIR) {
-        $Env:GIT_DIR
-    } else {
-        Get-LocalOrParentPath .git
+    [CmdletBinding()]
+    param(
+        [Parameter(ValueFromPipeline=$true,Position=0)]
+        [AllowEmptyCollection()]
+        [String[]] $Path
+    )
+    
+    PROCESS{
+        if ($Env:GIT_DIR) {
+            return Get-Item $Env:GIT_DIR
+        }
+
+        if (!$Path) {
+            return Get-LocalOrParentPath .git
+        }
+
+        $all = @()
+        foreach ($l in $Path) {
+            $all = $all + @(Get-LocalOrParentPath .git $l)
+        }
+        return $all | Get-Unique
     }
 }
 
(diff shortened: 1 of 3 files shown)


GitUtils.ps1
```diff
@@ -1,11 +1,79 @@
...
-        Get-LocalOrParentPath .git
+    [CmdletBinding()]
+    param(
+        [Parameter(ValueFromPipeline=$true,Position=0)]
+        [AllowEmptyCollection()]
+        [String[]] $Path
+    )
+    
+    PROCESS{
+        if ($Env:GIT_DIR) {
```
dahlbyk: Should values passed from the pipeline take precedence over `GIT_DIR`?

GitUtils.ps1
```diff
@@ -1,11 +1,79 @@
...
+#>
+function Test-GitWorkingTree {
+    [CmdletBinding()]
+    param(
+        [Parameter(ValueFromPipeline=$true,Position=0)]
+        [AllowEmptyCollection()]
+        [String[]] $Path,
+
+        [Parameter()]
+        [String] $GitDir = $Env:GIT_DIR
```
dahlbyk: Looks like this isn't used?

GitUtils.ps1
```diff
@@ -1,11 +1,79 @@
...
+    )
+    PROCESS {
+        $p = ?? $Path @(Get-Location)
```
dahlbyk: Looks like `$p` isn't used?

GitUtils.ps1
```diff
@@ -94,14 +162,30 @@ function Get-GitBranch($gitDir = $(Get-GitDirectory), [Diagnostics.Stopwatch]$sw
...
+    process {
+        $GitDir | foreach {
+            return GitStatusOf($_)
+        }
+    }
+}
+
+function GitStatusOf($GitDir) {
     $settings = $Global:GitPromptSettings
     $enabled = (-not $settings) -or $settings.EnablePromptStatus
```
dahlbyk: Maybe YAGNI, but it might be reasonable to add a `-Force` parameter to override `$GitPromptSettings.EnablePromptStatus`.

GitUtils.ps1
luolong: yeah, seems so...

GitUtils.ps1
luolong: True again. It must have been an artifact of some experiments I tried... probably safe to delete...

GitUtils.ps1
luolong: Probably yes.

GitUtils.ps1
luolong: I don't know. What does it do?

luolong:
All in all, I can probably make these changes in my repo and update the pull request, but I have no way of testing/verifying any of it as I have since moved away from Windows OS. 

Or, you can take these changes and incorporate them in your code as you see fit...

dahlbyk:
If you've moved on I can take this and run with it, no worries.

luolong:
Yeah, it might be the best way to go forward

dahlbyk:
Closing in deference to #278, which seems similar.

PR #151, opened 2014-07-25
update prompt (PSDrive)

nickvdyck:
When you have custom PSDrives set to acces long UNC or file paths, posh git will always override the prompt to the original path (ProviderPath).

Instead it would be nice that posh git behaves like powershell and updates the prompt to the path from the psprovider.

Example:
When you have the following PSDrive set:

```
Name           Used (GB)     Free (GB) Provider      Root                                                                             CurrentLocation
----           ---------     --------- --------      ----                                                                             ---------------
Projects                         62.41 FileSystem    C:\Users\<UserName>\Documents\Projects
LongPath                               FileSystem    \\Verry\\long\\unc\path
```

Posh git always updates the prompt to the original path:

```
PS C:\Users\<UserName>\Documents\Projects\Dir [master]>
```

This updates fixes this problem and makes posh-git behave as default powershell.

```
PS Projects:\Dir [master]>
```

Because at work I have many PSDrives mapping to UNC paths, which  keep clogging up my prompt.

Maybe there is a reason why you wouldn't do it this way, but I couldn't find anything against it in the commits and issues.

### DIFF
--- a/profile.example.ps1
+++ b/profile.example.ps1
@@ -15,7 +15,7 @@ function global:prompt {
     # Reset color, which can be messed up by Enable-GitColors
     $Host.UI.RawUI.ForegroundColor = $GitPromptSettings.DefaultForegroundColor
 
-    Write-Host($pwd.ProviderPath) -nonewline
+    Write-Host("PS $($pwd.Path)") -NoNewline
 
     Write-VcsStatus
 


rkeithhill:
This has been addressed in PR #322

PR #152, opened 2014-08-15
Git k tab completion

coridrew:
Nothing like pair-programming with the repo owner at ThatConference to make one submit a pull request :) Fun!


dahlbyk:
> @dahlbyk is spreading the love.

And then letting it sit for months. :broken_heart: But no longer!

Issue #130, opened 2014-01-11
Case sensitivity with file paths


nycdotnet:
Posh-git displays strange behavior if you navigate to a directory using the wrong case.

Imagine a directory called "code" with a subdirectory called "Model".  The posh-git prompt is in code.  User types "cd model" (all lower-case, so technically the incorrect case).  PowerShell handles this fine and will show the prompt as "code\model", but if you do git status (for example) it will show any changes in the Model folder as "..\Model\changedfile.txt" instead of "changedfile.txt".  This is technically correct if Git is assuming that both the model and Model directories exist. Other commands such as git mv OLD NEW, don't work at all as Git will warn that the file is not under source control.

Is there any way to get posh-git to intercept and fix cases on cd commands (or at least error/warn if a case is invalid)?  I'm willing to look into this a bit if it sounds like this is something that you're interested to address.  Thanks!

dahlbyk:
Not sure when this was fixed, but I can no longer reproduce this in git 2.10.2.

nycdotnet:
ty

PR #158, opened 2014-09-26
Implements #130 Case Sensitive Path

sclarson:
implements #130


dahlbyk:
@sclarson are you interested in revisiting this to resolve conflicts and address the feedback? If not, can you [let us edit this PR](https://github.com/blog/2247-improving-collaboration-with-forks)?

PR #159, opened 2014-10-02
Remove merged branches

### LINKED ISSUE #79: Remove remote merged branches command
So we (GHfW) team have a PowerShell script that will list all remote branches that are already merged into a specified branch (it defaults to origin/master if no branch is supplied) and offer to delete them for you.

I'd love to clean it up and add it as a command to `GitUtils.ps1`. Would you be interested in such a thing as part of Posh-Git?

tkirill:
Add the cmdlets for removing merged branches:
-  Remove-MergedLocalGitBranches -- removes local merged branches.
-  Remove-MergedRemoteGitBranches -- removes remote merged branches.
-  Remove-MergedGitBranches -- removes both local and remote merged branches.

Implementation placed in the separate file GitRemoveMergedBranches.ps1 although most of other exported cmdlets sit in GitUtils.ps1.  Hope it's not a code style rule to place everything inside GitUtils.ps1.

This is how these cmdlets work.

**Remove-MergedLocalGitBranches**
1.  Call `git branch --merged`.
2.  For each line trim whitespaces from both sides.
3.  Filter out line starting with "*" -- it is the current branch.
4.  Filter out ignored branches -- master and develop by default.
5.  If no branch remains then print message and return.
6.  For each remaining branch call `git branch -d $item`.

**Remove-MergedRemoteGitBranches**

Works same as Remove-MergedLocalGitBranches except three things:
1.  List of merged branches is retrieved by `git branch -r --merged`.
2.  Slightly more complex parsing and filtration.  We need to filter out the line with "/HEAD ->" and cut off `<remote-name>/` part from a beginning of each line.
3.  Removes branch with `git push origin :$item`.

**Remove-MergedGitBranches**
This one simply calls previous two cmdlets.

All three cmdlets support -WhatIf and -Confirm ([link1](http://blogs.technet.com/b/heyscriptingguy/archive/2012/10/02/build-your-own-powershell-cmdlet-part-4-of-9.aspx), [link2](http://becomelotr.wordpress.com/2013/05/01/supports-should-process-oh-really/)).  Also, they all support -Ignore parameter for overriding list of ignored branches.

I haven't tested it enough yet to my shame but I think it is a good idea to create PR to receive feedback as soon as possible.

fixes #79

### DIFF
--- a/GitRemoveMergedBranches.ps1
+++ b/GitRemoveMergedBranches.ps1
@@ -1,0 +1,86 @@
+function IsNotCurrentOrIgnoredLocal-GitBranch($branch)
+{
+    $ignored = "master", "develop"
+    !$branch.StartsWith("*") -and !$ignored.Contains($branch)
+}
+
+function TrimRemote-GitBranch($branch)
+{
+    $branch -replace "^(.*?)/(.*)$", "`$2"
+}
+
+function IsNotCurrentOrIgnoredRemote-GitBranch($branch)
+{
+    $ignored = "master", "develop"
+    if ($branch.Contains("/HEAD ->"))
+    {
+        return $false
+    }
+
+    $branchName = TrimRemote-GitBranch $branch
+    return !$ignored.Contains($branchName)
+}
+
+function Get-MergedLocalGitBranches
+{
+    $branches = git branch --merged
+    $trimmed = $branches | foreach {$_.Trim()}
+    return $trimmed | where {IsNotCurrentOrIgnoredLocal-GitBranch $_}
+}
+
+function Get-MergedRemoteBranches
+{
+    $branches = git branch -r --merged
+    $trimmed = $branches | foreach {$_.Trim()}
+    $filtered = $trimmed | where {IsNotCurrentOrIgnoredRemote-GitBranch $_}
+    return $filtered | foreach {TrimRemote-GitBranch $_}
+}
+
+function Remove-MergedLocalGitBranches
+{
+    $branches = Get-MergedLocalGitBranches
+    if ($branches.Length -eq 0)
+    {
+        Write-Host "No local merged branches"
+        return
+    }
+
+    Write-Host "Start deleting local merged branches"
+    Write-Host
+    
+    foreach ($item in $branches)
+    {
+        Write-Host "Deleting branch $item..."
+        git branch -d $trimmed
+        Write-Host
+    }
+    Write-Host "Finish deleting local merged branches"
+}
+
+function Remove-MergedRemoteGitBranches
+{
+    $branches = Get-MergedRemoteBranches
+    if ($branches.Length -eq 0)
+    {
+        Write-Host "No remote merged branches"
+        return
+    }
+
+    Write-Host "Start deleting remote merged branches"
+    Write-Host
+    
+    foreach ($item in $branches)
+    {
+        Write-Host "Deleting branch $item..."
+        git push origin :$branch
+        Write-Host
+    }
+    Write-Host "Finish deleting remote merged branches"
+}
+
+function Remove-MergedGitBranches
+{
+    Remove-MergedLocalGitBranches
+    Write-Host
+    Remove-MergedRemoteGitBranches
+}
--- a/GitRemoveMergedBranches.ps1
+++ b/GitRemoveMergedBranches.ps1
@@ -15,7 +15,7 @@ function Get-MergedLocalGitBranches($ignore)
 {
     $branches = git branch --merged
     $trimmed = $branches | foreach {$_.Trim()}
-    return $trimmed | where {(IsNotCurrentLocal-GitBranch $_) -and !$ignore.Contains($_)}
+    return $trimmed | where {IsNotCurrentLocal-GitBranch $_}
 }
 
 function Get-MergedRemoteBranches($ignore)
@@ -26,7 +26,7 @@ function Get-MergedRemoteBranches($ignore)
     return $trimmed `
         | where {!$_.Contains("/HEAD ->")} `
         | TrimRemote-GitBranch `
-        | where {($_ -ne $current) -and !$ignore.Contains($_)}
+        | where {$_ -ne $current}
 }
 
 function Remove-MergedLocalGitBranches
@@ -35,10 +35,10 @@ function Remove-MergedLocalGitBranches
     Param(
         [Parameter()]
         [string[]]
-        $Ignore = @("master", "develop")
+        $Ignore = $Global:GitPromptSettings.MergedBranchesToKeep
     )
 
-    $branches = Get-MergedLocalGitBranches $Ignore
+    $branches = Get-MergedLocalGitBranches | where {$_ -notin $Ignore}
     if ($branches.Length -eq 0)
     {
         Write-Host "No local merged branches"
@@ -66,10 +66,10 @@ function Remove-MergedRemoteGitBranches
     Param(
         [Parameter()]
         [string[]]
-        $Ignore = @("master", "develop")
+        $Ignore = $Global:GitPromptSettings.MergedBranchesToKeep
     )
 
-    $branches = Get-MergedRemoteBranches $Ignore
+    $branches = Get-MergedRemoteBranches | where {$_ -notin $Ignore}
     if ($branches.Length -eq 0)
     {
         Write-Host "No remote merged branches"
@@ -97,7 +97,7 @@ function Remove-MergedGitBranches()
     Param(
         [Parameter()]
         [string[]]
-        $Ignore = @("master", "develop")
+        $Ignore = $Global:GitPromptSettings.MergedBranchesToKeep
     )
 
     Remove-MergedLocalGitBranches @PSBoundParameters
--- a/GitRemoveMergedBranches.ps1
+++ b/GitRemoveMergedBranches.ps1
@@ -3,9 +3,12 @@ function IsNotCurrentLocal-GitBranch($branch)
     !$branch.StartsWith("*")
 }
 
-function TrimRemote-GitBranch($branch)
+function TrimRemote-GitBranch()
 {
-    $branch -replace "^(.*?)/(.*)$", "`$2"
+    PROCESS
+    {
+        $_ -replace "^(.*?)/(.*)$", "`$2"
+    }
 }
 
 function Get-MergedLocalGitBranches($ignore)
@@ -22,7 +25,7 @@ function Get-MergedRemoteBranches($ignore)
     $trimmed = $branches | foreach {$_.Trim()}
     return $trimmed `
         | where {!$_.Contains("/HEAD ->")} `
-        | foreach {TrimRemote-GitBranch $_} `
+        | TrimRemote-GitBranch `
         | where {($_ -ne $current) -and !$ignore.Contains($_)}
 }
 
--- a/GitRemoveMergedBranches.ps1
+++ b/GitRemoveMergedBranches.ps1
@@ -11,14 +11,14 @@ function TrimRemote-GitBranch()
     }
 }
 
-function Get-MergedLocalGitBranches($ignore)
+function Get-MergedLocalGitBranches()
 {
     $branches = git branch --merged
     $trimmed = $branches | foreach {$_.Trim()}
     return $trimmed | where {IsNotCurrentLocal-GitBranch $_}
 }
 
-function Get-MergedRemoteBranches($ignore)
+function Get-MergedRemoteBranches()
 {
     $current = Get-GitBranch
     $branches = git branch -r --merged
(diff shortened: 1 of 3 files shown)


tkirill:
I found a bug.  `Remove-MergedBranches` works bad with multiple remotes because it doesn't respect remote name in the output of `git branch --remote --merged`.  For example:

> C:\Users\t_kirill\workplace\gofra\repo [develop]> git branch --remote --merged
>  gitlab/counter-page-G-1202
>  gitlab/develop
>  gitlab/direct-link-excel
>  origin/HEAD -> origin/develop
>  origin/develop

If we have two remotes then merged branches will be deleted in the `origin` only.  `Remove-MergedBranches` see merged branches in remotes other than `origin` but can't delete them:

> Deleting remote branch counter-page-G-1202...
> error: unable to delete 'counter-page-G-1202': remote ref does not exist
> error: failed to push some refs to 'ssh://git@stash:7999/g/gofra.git'

I'll fix it on this week.

GitRemoveMergedBranches.ps1
```diff
@@ -0,0 +1,95 @@
...
+
+function TrimRemote-GitBranch()
+{
+    PROCESS
+    {
+        $_ -replace "^(.*?)/(.*)$", "`$2"
+    }
+}
+
+function Get-MergedLocalGitBranches($ignore)
```
dahlbyk: Drop unused `$ignore`?

GitRemoveMergedBranches.ps1
```diff
@@ -0,0 +1,95 @@
...
+{
+    $branches = git branch --merged
+    $trimmed = $branches | foreach {$_.Trim()}
+    return $trimmed | where {IsNotCurrentLocal-GitBranch $_}
+}
+
+function Get-MergedRemoteBranches($ignore)
```
dahlbyk: Drop unused `$ignore`?

dahlbyk:
Revisions from six months ago (ugh) look great. Looking forward to the fix for multiple remotes!

Would be simpler/cleaner to export a single `Remove-MergedGitBranches` that accepts a `-Remote` parameter to switch from local deletes to remote (a la `git branch -r`)?

GitRemoveMergedBranches.ps1
tkirill: Completely removed `$ignore` in  47e6c16.

GitRemoveMergedBranches.ps1
tkirill: `$ignore` was completely removed in 47e6c16.

tkirill:
@dahlbyk Your idea about `-Remote` parameter is good!  I implemented it in 31ce862 and it is really better now.  Also, I added `-All` parameter so it is possible to delete both merged and local branches in one call `Remove-MergedGitBranches -All`.

About the multiple remotes bug.  `Remove-MergedGitBranches` works with origin by default now.  There is `-RemoteName` parameter to switch to another remote.  I think an ability to work with multiple remotes in one call by passing array to `-RemoteName` is redundant.  What do you think?

GitRemoveMergedBranches.ps1
```diff
@@ -0,0 +1,90 @@
...
+        | where {$_ -notin $Global:GitPromptSettings.MergedBranchesToKeep}
+    if ($branches.Length -eq 0) {
+        Write-Host "No remote merged branches"
+        return
+    }
+
+    foreach ($item in $branches) {
+        if ($PSCmdlet.ShouldProcess($item, "delete remote branch")) {
+            Write-Host "Deleting remote branch $item..."
+            git push $RemoteName :$item
```
tkirill: No whitespace escaping for `$RemoteName` but I think it is safe since it is not possible to use whitespaces in remote name in Git. @dahlbyk What do you think?

GitRemoveMergedBranches.ps1
```diff
@@ -0,0 +1,90 @@
...
+        | where {$_ -notin $Global:GitPromptSettings.MergedBranchesToKeep}
+    if ($branches.Length -eq 0) {
+        Write-Host "No local merged branches"
+        return
+    }
+
+    foreach ($item in $branches) {
+        if ($PSCmdlet.ShouldProcess($item, "delete branch")) {
+            Write-Host "Deleting branch $item..."
+            git branch -d $item
```
tkirill: No whitespace escaping for `$item` but I think it is [safe](https://github.com/dahlbyk/posh-git/pull/159/files#r30898152) too.

GitRemoveMergedBranches.ps1
dahlbyk: Correct, no need to handle whitespace in remote name.

GitRemoveMergedBranches.ps1
dahlbyk: Correct

dahlbyk:
`-All` is a welcome addition, good call.

On the `-RemoteName` front, it seems like switching to `string[]` would be pretty simple implementation-wise, without changing the single-remote usage at all. I wonder if a more reasonable default behavior would be to remove all remotes' merged branches (I.e. not filter by remote name)?

-------------------

Maybe someone more in tune with the Zen of PowerShell can chime in, but it feels like most of the `Write-Host` here may be better as `Write-Verbose`? Not sure what the correct way is to implement an actual `-Verbose` flag, tho.

tkirill:
@dahlbyk I added support for multiple remotes in c0a821d.  This is really convenient, good that you noticed this!

> I wonder if a more reasonable default behavior would be to remove all remotes' merged branches

The only potential problem I can think of is read-only remote.  If I fork some project on GitHub I have two remotes:

* origin -- my fork.  I have write access.
* upstream -- original repository.  I need this remote for synchronization.  I don't have write access.

I think it is a frequent case and in this case working with all remotes by default doesn't work well.  However, we definitely need ability to work with all remotes if we want.  I thought to add flag `-AllRemotes` but we already have `-All` and these two look confusing together.  What if instead of working with all remotes by default we implement these flags:

* -Remote.  Delete remote branches in origin.
* -Local.  Delete local branches.
* -RemoteName.  Set an array of remotes' names.
* -AllRemote.  Delete remote branches in all remotes.
* -All.  Equals to `-Local -AllRemote`.

Default behavior without any flags equals to `-Local -Remote`.  It used to work like this before I added `-Remote` flag yesterday and it was convenient for me.  What do you think?

dahlbyk:
Remote operations being more "destructive" than their local counterparts, how about we remove any sort of default remote behavior?

* `Delete-MergedGitBranches` = `Delete-MergedGitBranches -Local` = delete merged local branches.
* `Delete-MergedGitBranches -Remote`= error, not enough information
* `Delete-MergedGitBranches -Remote origin`= delete merged remote branches on `origin`
* `Delete-MergedGitBranches -Local -Remote (git remote)` deletes all merged local and remote branches

tkirill:
@dahlbyk I like this idea! :+1: We have only two flags `-Local` and `-Remote` and they allow us to do everything we want.  I'll implement this today or tomorrow.

To be honest I don't find removing merged branches in `origin` dangerous and every time I delete local branches I want to delete them in origin as well.  So `Remove-MergedGitBranches == Remove-MergedGitBranches -Local -Remote origin` would be more convenient for me but I can create alias :ok_hand: 

P.S. You wrote `Delete-MergedGitBranches` instead of `Remove-MergedGitBranches`.  Not sure if it is mistype or not :)

dahlbyk:
Yes, mistyped. Remove is cool.

Dangerousness likely depends on team size and whether or not you use forks - I've dealt with work people doing a "matching" `git push --force`, so I'm pretty tentative about bulk operations against remotes.

tkirill:
@dahlbyk Take a look at new arguments please 27e75e5.

Surprisingly validation for empty set of remotes' names works automatically.  I didn't found this in [documentation](https://technet.microsoft.com/en-us/library/hh847743.aspx).

This is how it looks
parameters

tkirill:
Now after few days I doubt the new way is better.  Your suggestion about `Remote` parameter looks fine for me now.  Yes there is a possibility of deleting not merged remote branches but at least it is simple.  One just need to push to remote before deleting branches.

dahlbyk:
I'm not really a fan of doing any sort of remote manipulation based on local names because upstream branches don't necessarily have to track a local branch of the same name. I suppose you could remove remote branches that have been merged into the remote branch tracked by the current local branch, if applicable, but that feels like a stretch?

Since we don't specify a reference to the commit we're checking is merged, I can really only think of two reasonable behaviors for `Remove-MergedGitBranches -Remote origin`:

1. Remove remote branches that have been merged into local `HEAD` (i.e. `$(git branch -r --merged) -like '  origin/*'`).
2. Remove remote branches that have been merged into `origin/HEAD` (the default branch, typically `master` but configurable).

Between the two I lean toward the latter, with an error if `<remote>/HEAD` does not exist (it's typically created by `git clone` but not `git remote add`).

More generally, earlier I asked:

> General question: would it be useful for any/all of these to accept a parameter (defaulting to `HEAD`) to identify the ref to check mergedness against...?

Especially for the `-Remote` scenario the answer seems to be "Yes".

tkirill:
@dahlbyk I tried to implement `-Target` parameter in 7658e2f and it looks nice.  Here is current interface:

* `-Local` -- remove local branches.
* `-Remote origin, other` -- set remotes
* `-Target branch` -- set target branch.  `HEAD` is default.
* No parameters equals to `-Local`

`-Target` parameter is positional so this `Remove-MergedGitBranches branch1` removes local branches merged to `branch1`.

> ... with an error if <remote>/HEAD does not exist ...
I [write it as warning](https://github.com/tkirill/posh-git/blob/e016be982c5945c1c050bf173a92f5e0d2b15394/GitRemoveMergedBranches.ps1#L67) since it is not fatal.  What do you think?  This is how it looks:

head

Also, I [added a test](https://gist.github.com/tkirill/046845bee297170a4484#file-test-ps1-L322) with multiple remotes and custom target branch.  These cases are currently tested:

1. Do nothing if no branches: [local](https://gist.github.com/tkirill/046845bee297170a4484#file-test-ps1-L44) & [remote](https://gist.github.com/tkirill/046845bee297170a4484#file-test-ps1-L53)
2. Do nothing if branch not merged: [local](https://gist.github.com/tkirill/046845bee297170a4484#file-test-ps1-L64) & [remote](https://gist.github.com/tkirill/046845bee297170a4484#file-test-ps1-L167)
3. Remove single merged branch: [local](https://gist.github.com/tkirill/046845bee297170a4484#file-test-ps1-L92) & [remote](https://gist.github.com/tkirill/046845bee297170a4484#file-test-ps1-L153)
4. Do nothing if branch is merged but it is a current branch: [local](https://gist.github.com/tkirill/046845bee297170a4484#file-test-ps1-L78)
5. Pass HEAD explicitly: [local](https://gist.github.com/tkirill/046845bee297170a4484#file-test-ps1-L105)
6. Custom branch as a target: [local](https://gist.github.com/tkirill/046845bee297170a4484#file-test-ps1-L118) & [remote](https://gist.github.com/tkirill/046845bee297170a4484#file-test-ps1-L322)
7. Multiple merged branches: [local](https://gist.github.com/tkirill/046845bee297170a4484#file-test-ps1-L138) & [remote](https://gist.github.com/tkirill/046845bee297170a4484#file-test-ps1-L239)
8. If branch merged locally and not merged remotely then remove only local branch: [test](https://gist.github.com/tkirill/046845bee297170a4484#file-test-ps1-L181)
9. If branch merged remotely and not merged locally then remove only remote branch: [test](https://gist.github.com/tkirill/046845bee297170a4484#file-test-ps1-L200)
10. If branch merged remotely and locally then remove both local and remote branch: [test](https://gist.github.com/tkirill/046845bee297170a4484#file-test-ps1-L225)
11. Multiple remotes:
    * [no merged branches](https://gist.github.com/tkirill/046845bee297170a4484#file-test-ps1-L255)
    * [only one remote has merged branch](https://gist.github.com/tkirill/046845bee297170a4484#file-test-ps1-L277)
    * [each remote has merged branch](https://gist.github.com/tkirill/046845bee297170a4484#file-test-ps1-L298)

dahlbyk:
No time to review code now, but API seems just right and test coverage is impressive! Will try to revisit over the weekend - bug me next week if I don't.

Thanks for iterating on this!

tkirill:
> bug me next week if I don't.
:wave:

tkirill:
Just discovered one more corner case.

If we have branch which point at the same commit as `develop` then this branch considered merged and the script removes it.  However, it can be just new branch that has no commits yet.  Removing such branch surprised my colleague who created this branch :laughing:

I don't know if we could fix this since the same situation occurs with fast-forward merge.

rkeithhill:
Closing this in favor of PR #663.

PR #168, opened 2014-12-16
Added "Administrator: " prefix to window title

dlwyatt:
Posh-Git was not including this prefix when it set the title, and I keep forgetting which window was launched with "Run As Administrator" and which was not.

### DIFF
--- a/GitPrompt.ps1
+++ b/GitPrompt.ps1
@@ -138,11 +138,20 @@ function Write-GitStatus($status) {
             }
             $repoName = Split-Path -Leaf (Split-Path $status.GitDir)
             $prefix = if ($s.EnableWindowTitle -is [string]) { $s.EnableWindowTitle } else { '' }
-            $Host.UI.RawUI.WindowTitle = "$prefix$repoName [$($status.Branch)]"
+            $adminHeader = Get-AdminTitleHeader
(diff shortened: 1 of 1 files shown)


dahlbyk:
This is a great addition - :disappointed: that I let it sit for so long.

GitPrompt.ps1
```diff
@@ -138,13 +138,22 @@ function Write-GitStatus($status) {
             }
             $repoName = Split-Path -Leaf (Split-Path $status.GitDir)
             $prefix = if ($s.EnableWindowTitle -is [string]) { $s.EnableWindowTitle } else { '' }
-            $Host.UI.RawUI.WindowTitle = "$prefix$repoName [$($status.Branch)]"
+            $adminHeader = Get-AdminTitleHeader
```
dahlbyk: Do you have any concerns about performance? I wonder if it would be sufficient to check admin-ness once when the module loads and use that for the rest of the session?

Or it may be that the security stuff is trivial in cost compared to the Git work we're doing?

dlwyatt: Probably both statements are true.  The cost is trivial compared to every time you spin up a new process to call git.exe, but there's also no reason _not_ to cache the admin status, since it's never going to change within the same session.

dahlbyk:
Thanks!

PR #170, opened 2014-12-22
Adding try/catch logic to Write-VcsStatus

paulmarsy:
Adding try/catch logic to Write-VcsStatus so that if one of the scriptblocks that has been added to the VcsPromptStatuses array fails the other scriptblocks are not affected (are still invoked)

### DIFF
--- a/GitPrompt.ps1
+++ b/GitPrompt.ps1
@@ -37,6 +37,9 @@ $global:GitPromptSettings = New-Object PSObject -Property @{
     UntrackedText             = ' !'
     UntrackedForegroundColor  = [ConsoleColor]::DarkRed
     UntrackedBackgroundColor  = $Host.UI.RawUI.BackgroundColor
+    
+    ErrorForegroundColor      = [ConsoleColor]::Red
+    ErrorBackgroundColor      = $Host.UI.RawUI.BackgroundColor
 
     ShowStatusWhenZero        = $true
 
@@ -167,7 +170,27 @@ function Write-GitStatus($status) {
 if(!(Test-Path Variable:Global:VcsPromptStatuses)) {
     $Global:VcsPromptStatuses = @()
 }
-function Global:Write-VcsStatus { $Global:VcsPromptStatuses | foreach { & $_ } }
+
+function Global:Write-VcsStatus {
+    $Global:VcsPromptStatuses | foreach {
+        $vcsPromptErrors = $null
+        try {
+            Invoke-Command -ScriptBlock $_ -ErrorVariable vcsPromptErrors 2>$null
+        }
+        catch {
+            $vcsPromptErrors = $_
+        }
+        if ($vcsPromptErrors.Length -gt 0) {
+            $vcsPromptErrors | % { Write-Error -ErrorRecord $_ -ErrorAction SilentlyContinue }
+            $s = $Global:GitPromptSettings
+            if ($s) {
+                Write-Prompt $s.BeforeText -BackgroundColor $s.BeforeBackgroundColor -ForegroundColor $s.BeforeForegroundColor
+                Write-Prompt "Error" -BackgroundColor $s.ErrorBackgroundColor -ForegroundColor $s.ErrorForegroundColor
+                Write-Prompt $s.AfterText -BackgroundColor $s.AfterBackgroundColor -ForegroundColor $s.AfterForegroundColor
+            }
+        }
+    }
+}
 
 # Add scriptblock that will execute for Write-VcsStatus
 $PoshGitVcsPrompt = {


dahlbyk:
Sorry for the ridiculous delay in looking at these... I like the `try/catch` to make `Write-VcsStatus` less destrictive, but have concerns about the extra change for #169. If you can tease the changes apart, we can get this in first.

Once we get this nailed down, you might consider proposing an equivalent change to [posh-hg](https://github.com/JeremySkinner/posh-hg/blob/master/HgPrompt.ps1#L113-L122).

GitPrompt.ps1
```diff
@@ -145,13 +145,31 @@ function Write-GitStatus($status) {
     }
 }
 
-if(!(Test-Path Variable:Global:VcsPromptStatuses)) {
-    $Global:VcsPromptStatuses = @()
+$VcsPromptStatusesVariableName = "VcsPromptStatuses"
+
+if (-not (Test-Path Variable:Global:$VcsPromptStatusesVariableName)) {
+    New-Variable -Name $VcsPromptStatusesVariableName -Value @() -Scope Global
```
dahlbyk: Out of curiosity, why `New-Variable ...` versus the simpler `$Global:VcsPromptStatuses = ...`?

GitPrompt.ps1
```diff
@@ -145,13 +145,31 @@ function Write-GitStatus($status) {
...
+}
+
+function Global:Write-VcsStatus {
+    Get-Variable -Name $VcsPromptStatusesVariableName -Scope Global -ValueOnly | foreach { 
+        try {
+            $_.Invoke()
+        }
+        catch {
+            Write-Error "A Write-VcsStatus scriptblock failed with the error: $($_.Exception.InnerException)"
```
dahlbyk: Since this is really meant to be used in a prompt, I wonder if it would be less intrusive to replace `Write-Error` with an inline message that looks like an error (i.e. `Write-Host '[error]' -NoNewLine -ForegroundColor Red -BackgroundColor Black`)?

paulmarsy:
My mistake, I thought I had separated the two as they were done in separate branches.
I'll update this so it is purely the try/catch logic change :smile:

GitPrompt.ps1
paulmarsy: This relates more to the other proposed change, it was so the name of the global variable is controlled & set in one location, currently the $VcsPromptStatusesVariableName variable but potentially it could be set/changed from $GitPromptSettings

GitPrompt.ps1
paulmarsy: How about something like:
PS C:\> $Error
PS C:\> Write-Error -Message "test error" -ErrorAction SilentlyContinue
PS C:\> $Error
Write-Error -Message "test error" -ErrorAction SilentlyContinue : test error
    + CategoryInfo          : NotSpecified: (:) [Write-Error], WriteErrorException
    + FullyQualifiedErrorId : Microsoft.PowerShell.Commands.WriteErrorException

Then the normal Write-Error message isn't displayed but it is still in the error stream & error array so it can be debugged.

You are right as it executed in the prompt so shouldn't be intrusive, following that line of thought If there is an error and it is logged (silently), should doing any Write-Host be avoided as we don't know how it will affect the user's prompt? Or a message just saying [Error] as per your comment?
I am unsure if [error] was intended to be the error message or the actual word '[error]'.

paulmarsy:
The last two commits (b6dd6ea292298ca547b1a704a942dd51b7fa2f29 and 475491551aea5e3237ba03aceeee712ec7c1561e) detect errors written to the error stream as well as terminating exceptions which are thrown.

Let me know if you'd prefer them to be removed, as they detect non-terminating errors it may be better to continue as per normal.

GitPrompt.ps1
```diff
@@ -153,10 +156,31 @@ function Write-GitStatus($status) {
...
+        $vcsPromptErrors = $null
+        try {
+            Invoke-Command -ScriptBlock $_ -ErrorVariable vcsPromptErrors 2>$null
+        }
+        catch {
+            $vcsPromptErrors = $_
+        }
+        if ($vcsPromptErrors.Length -gt 0) {
+            # Log the errors but dont affect the users prompt, splat error objects into Write-Error
+            $vcsPromptErrors | % { Write-Error -ErrorAction SilentlyContinue @_ }
```
dahlbyk: `@_`?

dahlbyk:
New changes look good - can you do me a favor and `rebase` on master without the unrelated commits?

paulmarsy:
I've tweaked the history to remove the WIP & unrelated checkins so it is cleaner now 😀

I have replaced the using of the splat operator with just setting the ErrorRecord parameter to be the object.. Regardless of the type of exception thrown or what's sent to write-error they get changed into ErrorRecord objects so no need to map the object's properties onto the silent write-error call as I was doing.

For now I've left in the functionality which detects and suppresses errors written to the errors stream/use of write-error.. It does complicate the function somewhat so let me know if you'd prefer to simplify it to only catch terminating exceptions.
If that is your preference I'll leave the current commit in the history for reference in the incase it may be useful somewhere else.

Off topic but related.. Posh-git, posh-hg and posh-svn all implement the same functions, has there been any discussion or thought on a provider agnostic base containing these common hooks so each provider repo can focus on its specific implementation details?

paulmarsy:
Oops - I accidentally closed the PR while updating it with the latest changes :blush:

Let me know if you're happy with the proposed implementation - I am in two minds about the detection & suppression of the error stream in addition to handling thrown exceptions..
It complicates the error handling logic & function, but does provide an extra layer of error handling for invoked code which can be external to posh-git.

dahlbyk:
> Posh-git, posh-hg and posh-svn all implement the same functions, has there been any discussion or thought on a provider agnostic base containing these common hooks so each provider repo can focus on its specific implementation details?

Discussion? No. I tried to keep the `Write-VcsStatus` logic simple specifically to keep our shared surface area small, allowing independent evolution as necessary. To that end, my first instinct here is to just mandate that the subscribed status writers guarantee they won't throw so we can keep the copy-pasta simple.

That said, as a user of only posh-git I'm a very poor steward the multi-VCS situation (c.f. #111). If there's a clean way to extract a shared module on which all three could depend, I'm all for it (provided that manual, PsHet and Chocolatey installations keep working with minimal ceremony).

dahlbyk:
@paulmarsy is this still something that you're struggling with? I don't object to the change in principle, but I let it get way out of date. 😞

PR #171, opened 2015-01-09
Update readme.me formatting

JayBazuzi:


dahlbyk:
Took a long time to get through everything, but this looks good.

PR #174, opened 2015-01-12
Adds support for shorthand force-push syntax

### LINKED ISSUE #173: Feat: Support additional short-hand modifiers in branch auto-complete
When force pushing a branch to a remote one can use either:
- git push origin master --force

Or the shorthand version
- git push origin +master

Unfortunately posh-git is unable to perform auto-complete of a branch name when the "+" prefx is used, presumably it is assuming the "+" is the first letter of a branch name.  Likewise the same issue occurs when deleting a remote branch:
- git push origin :master

It would be _amazing_ if posh-git could recognise + and : as simply modifiers and still perform branch auto-complete.

dan-turner:
Git supports a short-hand syntax for force-pushing a branch to a remote by including the "+" modifier before the branch name. However when doing so, Posh-git was no longer able to auto-complete the branch name as it did not understand this syntax.  This PR adds that support.

Closes #173

### DIFF
--- a/GitTabExpansion.ps1
+++ b/GitTabExpansion.ps1
@@ -223,6 +223,11 @@ function GitTabExpansion($lastBlock) {
         "^push.* (?<remote>\S+) (?<ref>[^\s\:]*\:)(?<branch>\S*)$" {
             gitRemoteBranches $matches['remote'] $matches['ref'] $matches['branch']
         }
+        
+        # Handles git push remote +<branch>  (force push)
+        "^push.* (?<remote>\S+) (?<ref>\+)(?<branch>\S*)$" {
+            gitRemoteBranches $matches['remote'] $matches['ref'] $matches['branch']
+        }
 
         # Handles git push remote <branch>
         # Handles git pull remote <branch>


dan-turner:
Any chance of this making it in @dahlbyk?

GitTabExpansion.ps1
```diff
@@ -223,6 +223,11 @@ function GitTabExpansion($lastBlock) {
         "^push.* (?<remote>\S+) (?<ref>[^\s\:]*\:)(?<branch>\S*)$" {
             gitRemoteBranches $matches['remote'] $matches['ref'] $matches['branch']
         }
+        
+        # Handles git push remote +<branch>  (force push)
+        "^push.* (?<remote>\S+) (?<ref>\+)(?<branch>\S*)$" {
```
dahlbyk: I don't think this is quite what we want... This will capture the `+` in the `ref` group and try matching remote branches based on that.

I _think_ it would be sufficient to add `\+?` above/below before we capture `<ref>:<branch>` or just `<branch>` - give that a try?

dahlbyk:
So sorry for the delay in looking at this, @dan-turner. I appreciate you looping back around to bump it, as this is definitely something we should support.

PR #176, opened 2015-01-21
Improve the user experience when ssh-agent.exe isnt in the PATH.

mosoto:
I've run into the problem of ssh-agent not being in my PATH and searching for how to fix it showed a few hits which means other people run into it too.  I figured it might be a good idea to just provide a quick way to fix it for the common case.


dahlbyk:
This should be addressed now that #141 has finally been merged. Give that a try and let me know if you think this is still necessary?

PR #181, opened 2015-03-18
Added stash count to prompt.

fritzh:


dahlbyk:
Sorry for the delay in review, but thanks for the contribution!

A few comments:

1. Thoughts on an `EnableStashStatus` flag that defaults to `$false`? I can see how some might like this feature, but I'm not sure it should be on by default since the prompt itself has been pretty stable for a few years - I'd probably disable it, myself (I've come to only use stashes as a safer `reset --hard`).
2. No idea if people actually use them, but to align with the other path components maybe add `BeforeStashText = ' ('` and `AfterStashText = ')'` options? Should they have their own colors apart from the color of the stash count itself?
3. Can you rebase just your changes on `master` (without the gitprompt merge/revert)?

.gitignore
```diff
@@ -1,2 +1,3 @@
 VERSION
-obj
\ No newline at end of file
+obj
+bin
```
dahlbyk: I don't think we need this.

GitUtils.ps1
```diff
@@ -115,12 +115,15 @@ function Get-GitStatus($gitDir = (Get-GitDirectory)) {
         $filesModified = @()
         $filesDeleted = @()
         $filesUnmerged = @()
+        $stashCount = 0
 
         if($settings.EnableFileStatus -and !$(InDisabledRepository)) {
             dbg 'Getting status' $sw
             $status = git -c color.status=false status --short --branch 2>$null
+            $stashCount = $null | git stash list 2>$null | measure-object | select -expand Count
```
dahlbyk: > Pipe null into git stash command because otherwise it grabs everything in stdin - annoying if you're in the habit of typing before you get the next command prompt.

Piping `$null` into `git stash` is...curious. It seems like the problem might actually be the pager - try `git --no-pager list`?

GitUtils.ps1
```diff
@@ -115,12 +115,15 @@ function Get-GitStatus($gitDir = (Get-GitDirectory)) {
...
         $filesUnmerged = @()
+        $stashCount = 0
 
         if($settings.EnableFileStatus -and !$(InDisabledRepository)) {
             dbg 'Getting status' $sw
             $status = git -c color.status=false status --short --branch 2>$null
+            $stashCount = $null | git stash list 2>$null | measure-object | select -expand Count
         } else {
             $status = @()
+            $stashCount = 0
```
dahlbyk: Can drop this - initialized above.

GitUtils.ps1
dahlbyk: If we do go the `EnableStashStatus` route, let's do that check outside the `EnableFileStatus` check. No need for the settings to be interdependent.

.gitignore
fritzh: Good spot, I will remove.

GitUtils.ps1
fritzh: I tried `--no-pager` but unfortunately it didn't work. Good line of thought though, I didn't think about that. Is there anything else that could be causing git to grab all input like that?

fritzh:
_Thoughts on an EnableStashStatus flag that defaults to $false?_
Excellent point - most of my co-workers use it but one or two others can't be bothered, so makes great sense. I will make that change shortly.

_No idea if people actually use them, but to align with the other path components maybe add BeforeStashText = ' (' and AfterStashText = ')' options? Should they have their own colors apart from the color of the stash count itself?_
Will do on `BeforeStashText` and `AfterStashText` - not sure about their own colours, as the other status indicators all seem to have the same colour?

_Can you rebase just your changes on master (without the gitprompt merge/revert)?_
Done :-)

fritzh:
I have now added the configuration options.

GitUtils.ps1
dahlbyk: Well I'm stumped... I can confirm the swallowed input issue, and the fix, but I can't reproduce for any other Git command (`reflog` seeming like it may be most similar). Very odd.

dahlbyk: > If we do go the `EnableStashStatus` route, let's do that check outside the `EnableFileStatus` check. No need for the settings to be interdependent.

Status setting looks good on the output side, but we're still doing the work even if we don't need it.

dahlbyk:
Almost there, just need another `EnableStashStatus` check.

> > Can you rebase just your changes on master (without the gitprompt merge/revert)?
> 
> Done :-)

I'm still seeing a few process/pageant-related commits here. Try a `git rebase --interactive master` and delete the lines for commits before "Added stash count to prompt.".

dahlbyk:
Superseded by #181 - thanks for the contribution!

PR #183, opened 2015-03-30
Refactor all git calls to capture STDERR properly

### LINKED ISSUE #182: Posh-git cmdlets misbehave if you have $ErrorActionPreference = "Stop"
One of the most annoying if you go to the repository folder with detached head then prompt function fails and shows ugly _C:\MyProjectPS>_

mnaoumov:
Fixes #182

### DIFF
--- a/GitPrompt.ps1
+++ b/GitPrompt.ps1
@@ -140,7 +140,7 @@ function Write-GitStatus($status) {
             $prefix = if ($s.EnableWindowTitle -is [string]) { $s.EnableWindowTitle } else { '' }
             $Host.UI.RawUI.WindowTitle = "$prefix$repoName [$($status.Branch)]"
         }
-    } elseif ( $Global:PreviousWindowTitle ) {
+    } elseif ((Test-Path -Path Variable:Global:PreviousWindowTitle)) {
         $Host.UI.RawUI.WindowTitle = $Global:PreviousWindowTitle
     }
 }
--- a/GitPrompt.ps1
+++ b/GitPrompt.ps1
@@ -133,7 +133,7 @@ function Write-GitStatus($status) {
         Write-Prompt $s.AfterText -BackgroundColor $s.AfterBackgroundColor -ForegroundColor $s.AfterForegroundColor
 
         if ($WindowTitleSupported -and $s.EnableWindowTitle) {
-            if( -not $Global:PreviousWindowTitle ) {
+            if (-not (Test-Path -Path Variable:Global:PreviousWindowTitle)) {
                 $Global:PreviousWindowTitle = $Host.UI.RawUI.WindowTitle
             }
             $repoName = Split-Path -Leaf (Split-Path $status.GitDir)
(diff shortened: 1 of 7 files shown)


GitTabExpansion.ps1
```diff
@@ -299,6 +299,7 @@ if (Test-Path Function:\TabExpansion) {
 }
 
 function TabExpansion($line, $lastWord) {
+    "line: $line, lastWord: $lastWord" > c:\dev\333.txt
```
dahlbyk: Debugging?

GitUtils.ps1
```diff
@@ -345,17 +345,17 @@ function Stop-SshAgent() {
...
-    git checkout -q $Upstream
-    $branches = (git branch --no-color --no-merged) | where { $_ -notmatch '^\* ' }
+    $head = exec { git rev-parse --abbrev-ref HEAD }
+    exec { git checkout -q $Upstream }
+    $branches = (exec { git branch --no-color --no-merged }) | where { $_ -notmatch '^\* ' }
     foreach ($line in $branches) {
         $branch = $line.SubString(2)
         if (!$Quiet) { Write-Host "Rebasing $branch onto $Upstream..." }
-        git rebase -q $Upstream $branch > $null 2> $null
+        safeexec { git rebase -q $Upstream $branch } > $null
```
dahlbyk: Why use `safeexec` here when the surrounding calls all use `exec`?

dahlbyk:
Sorry for the delay in looking at this - really nice work! I'm going to give this a try locally, but everything makes sense at a glance.

Can you add a note to `Exec.ps1` referencing https://github.com/mnaoumov/Invoke-NativeApplication for anyone looking for more detail as to what's going on there?

GitPrompt.ps1
```diff
@@ -133,7 +133,7 @@ function Write-GitStatus($status) {
         Write-Prompt $s.AfterText -BackgroundColor $s.AfterBackgroundColor -ForegroundColor $s.AfterForegroundColor
 
         if ($WindowTitleSupported -and $s.EnableWindowTitle) {
-            if( -not $Global:PreviousWindowTitle ) {
+            if (-not (Test-Path -Path Variable:Global:PreviousWindowTitle)) {
```
dahlbyk: Can you apply this strict revision to `elseif ( $Global:PreviousWindowTitle )` below? Getting this:

> The variable '$Global:PreviousWindowTitle' cannot be retrieved because it has not been set.

GitUtils.ps1
mnaoumov: Because the original call had **2> $null**, so we should swallow all errors, that's what **safeexec** does

GitTabExpansion.ps1
mnaoumov: Fixed

GitPrompt.ps1
mnaoumov: Fixed

GitUtils.ps1
dahlbyk: So it did - thanks for clarifying.

dahlbyk:
Cool, fixes look good. Going to run with this for today and see if anything seems out of sorts.

dahlbyk:
@rkeithhill @lzybkr what do you think about this PR?

rkeithhill:
Take a look at my comment on the related issue - #182.  With 0.6.1, I can't repro the issue this PR is meant to fix.  Can you?

I simply see the `<SHA1>...` in the detached HEAD state.  It also inserts safeexec in places that weren't ignoring stderr before and the TestInPrompt function seems a bit brittle, relying on index 2 up the PS callstack.  And I'm also not a big fan of using aliases in a script.  That is just one more level of indirection to lookup.  :-)

dahlbyk:
Yeah, I think we're probably better off with targeted error handling (or, as in #307, using Git options to suppress errors).

Still, belated thanks for the contribution @mnaoumov!

PR #186, opened 2015-04-14
Warn of PowerShell 2.0 deprecation

### LINKED ISSUE #163: Deprecate PowerShell 2.0 Support
Dropping support for PowerShell 2.0 means fewer workarounds like #162. Before we do so, it would be good to warn those using 2.0 about the change:

* [ ] Add warning mentioning that PowerShell 2.0 will no longer be supported.
* [ ] Provide link to [PowerShell 3.0 download](https://www.microsoft.com/en-us/download/details.aspx?id=34595).
* [ ] Provide link to this issue for those wishing to express dissent.
* [ ] Provide a setting to disable the warning?

theaquamarine:
Closes #163

### DIFF
--- a/posh-git.psm1
+++ b/posh-git.psm1
@@ -1,9 +1,12 @@
+param([switch]$NoVersionWarn = $false)
+
 if (Get-Module posh-git) { return }
 
-if ($PSVersionTable.PSVersion.Major -lt 3) {
+if ($PSVersionTable.PSVersion.Major -lt 3 -and !$NoVersionWarn) {
     Write-Warning ("posh-git support for PowerShell 2.0 is deprecated; you have version $($Host.Version).`n" +
     "To download version 3.0, please visit https://www.microsoft.com/en-us/download/details.aspx?id=34595`n" +
-    "For more information and to discuss this, please visit https://github.com/dahlbyk/posh-git/issues/163`n")
+    "For more information and to discuss this, please visit https://github.com/dahlbyk/posh-git/issues/163`n" +
+    "To suppress this warning, change your profile to include 'Import-Module posh-git -Args `$true'.")
 }
 
 Push-Location $psScriptRoot
--- a/posh-git.psm1
+++ b/posh-git.psm1
@@ -2,8 +2,10 @@ param([switch]$NoVersionWarn = $false)
 
 if (Get-Module posh-git) { return }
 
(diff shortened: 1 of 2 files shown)


posh-git.psm1
```diff
@@ -1,5 +1,14 @@
+param([switch]$NoVersionWarn = $false)
+
 if (Get-Module posh-git) { return }
 
+if ($PSVersionTable.PSVersion.Major -lt 3 -and !$NoVersionWarn) {
+    Write-Warning ("posh-git support for PowerShell 2.0 is deprecated; you have version $($Host.Version).`n" +
```
dahlbyk: Let's use `$($PSVersionTable.PSVersion)` instead of `$($Host.Version)`, since the latter will only match for the standard host (not in the NuGet console, for example).

dahlbyk:
So sorry for the delay in looking at this. Looks great other than my comment on `$Host`.

readme.md
```diff
@@ -30,7 +30,7 @@ Install-Module posh-git
 Installing (manual)
 -------------------
 
-0. Verify you have PowerShell 2.0 or better with $PSVersionTable.PSVersion
+0. Verify you have PowerShell 2.0 or better with `$PSVersionTable.PSVersion`. PowerShell 3.0 is preferred as 2.0 support is deprecated.
```
dahlbyk: Bah, I think I just merged a conflict (#171).

theaquamarine:
Version check fixed, conflict fixed.

dahlbyk:
Perfect, thanks!

dingmanyao:
Solve the problem of warning, can modified first line of the posh-git.psm1:
param([switch]$NoVersionWarn = $false) ==>param([switch]$NoVersionWarn = $true)
image
这个问题是因为PowerShell3.0需要软件支持，你的系统没有所以无法升级到3.0，但是使用2.0就有警告，那么两种办法：1.去掉警告，继续使用2.0，不要升级，用上面的办法就可以；2.打开https://www.microsoft.com/en-us/download/details.aspx?id=34595   下载需要的软件，再升级。

dahlbyk:
> Solve the problem of warning, can modified first line of the posh-git.psm1:

Rather than modify the module, I would suggest updating `profile.example.ps1` (or copy its contents into your own `$PROFILE`) to change the `Import-Module` line to:

```powershell
Import-Module .\posh-git -ArgumentList $true
```

PR #192, opened 2015-05-17
Added tags file to .gitignore

jeyoung:
There seems to be a hook in git for Windows 1.95 that generates a tags file by default. I added the tags file to .gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 VERSION
 obj
+tags


dahlbyk:
I can't reproduce this with Git 1.9.5... what's in the file?

jeyoung:
It is reproducible if you have ctags.exe installed in Windows. The file contains tags used for semantic navigation of code in vim.

Sent from my Windows Phone
________________________________

jeyoung:
Just retried this on a clean machine with git for Windows and ctags.exe, and the tags file is not generated. Traced it back to a vim plugin creating a hook as part of its set up. The pull request can be abandoned, if you wish so.

dahlbyk:
Seems like a good candidate for your global `.gitignore`.

PR #195, opened 2015-05-25
Support custom key file names with ssl connections

c-mu:
Start-SshAgent finds all keyfiles (standard and cusomt named) and passes their paths to Add-SshKey.

### DIFF
--- a/GitUtils.ps1
+++ b/GitUtils.ps1
@@ -313,7 +313,13 @@ function Start-SshAgent([switch]$Quiet) {
             }
         }
     }
-    Add-SshKey
+    $keyfiles = Get-ChildItem -Path "$home/.ssh/*" -Exclude *.*, known_hosts
+    $keyfilepaths = @()
+    foreach($f in $keyfiles)
+    {
+        $keyfilepaths = $keyfilepaths + $f.FullName -replace "c:", "/c" -replace "\\", "/" 
+    }
+    Add-SshKey($keyfilepaths)
 }
 
 function Get-SshPath($File = 'id_rsa')


theaquamarine:
Might make more sense in ```Add-SshKey```?

The pageant logic searches for all .ppk files in ~/.ssh if no files are provided to ```Add-SshKey```, this could do the same thing rather than just letting ssh-add load the files it looks for by default.

dahlbyk:
On the flip side, I can see value in aligning `Add-SshKey` with [`ssh-add`](http://linux.die.net/man/1/ssh-add)'s no-arg behavior.

What if there were an `-All` option?

* `Add-SshKey` - default `ssh-add`
* `Add-SshKey -All` - all keys in `~/.ssh`
* `Add-SshKey [-File] 'list','of','files'` - specific paths

From a single-responsibility standpoint, would it make sense to remove the `Add-SshKey` call from `Start-SshAgent` to be more explicit about what key(s) should be loaded?

GitUtils.ps1
```diff
@@ -313,7 +313,13 @@ function Start-SshAgent([switch]$Quiet) {
             }
         }
     }
-    Add-SshKey
+    $keyfiles = Get-ChildItem -Path "$home/.ssh/*" -Exclude *.*, known_hosts
```
dahlbyk: Maybe extract `Get-SshDirectory` from `Get-SshPath` so we're consistent in referencing that path? Just noticed we have slightly different logic in the `plink` branch of `Add-SshKey`.

GitUtils.ps1
```diff
@@ -313,7 +313,13 @@ function Start-SshAgent([switch]$Quiet) {
             }
         }
     }
-    Add-SshKey
+    $keyfiles = Get-ChildItem -Path "$home/.ssh/*" -Exclude *.*, known_hosts
+    $keyfilepaths = @()
+    foreach($f in $keyfiles)
+    {
+        $keyfilepaths = $keyfilepaths + $f.FullName -replace "c:", "/c" -replace "\\", "/"
```
dahlbyk: I haven't found this sort of path manipulation to be necessary with MinGW, and [that seems to be one of their goals](http://sourceforge.net/p/mingw/mailman/message/8686481/).

theaquamarine:
Removing the ```Add-SshKey``` call makes sense, putting it separately into the profile would make it more obviously user-configurable, which would work well with the ```-All``` option. Plus, it'd be more accurate naming, since the function is ```Start-SshAgent``` rather than ```Start-SshAgentAndAddKeys```. Seems like a good idea.

c-mu:
Just one more thought on the single-responsibility approach: I think leaving out Add-SshKey from Start-SshAgent would over do it. The SSH-Agent is rather useless without any key associated with it. I agree with you on the changes proposed to Add-SshKey, but I think that Start-SshAgent should also take a `[-File] 'list', 'of', 'Files'` parameter, potentially with an `-All` option. This way the API would ensure that the SSH Agent is started in a usable way by reminding the developer that keys are actually required to make the connection happen. What's your take on this?

dahlbyk:
I agree that it's rather useless without an add, but I'm not sure the happy path user is really impacted if we change the default profile to `Start-SshAgent; Add-SshKey`. Seems simpler than mirroring `Add-SshAgent` parameters onto `Start-SshAgent`?

c-mu:
I'll do both, then you can review and select the one you like better. 
Should I also add an option to use a different path to hold the keys? I was thinking something along the lines `Add-SshKey [-File] 'list', 'of', 'files' -All -Path 'Custom/KeyFile/Path'`, where `-Path` defaults to "$env:USERPROFILE/.ssh". Same would go for the extended version of `Start-SshAgent`.
I'll start with the other changes while waiting for your feedback.

dahlbyk:
I think I'm fine with the extra parameters on `Start-SshAgent`. It's not like these will really ever be used outside a profile. (Please correct me if this is an invalid assumption!)

Adding `-Path` seems reasonable. Or we could make sure absolute file paths can work, ideally from pipeline (e.g. `gci C:\my-ssh | Add-SshKey`)? Or both, I suppose? Thoughts?

c-mu:
The `-File` already takes any kind of path and subsequently passes it to the SSH agent being used. It's the `-All` that currently implicitly assumes that the key-files reside in `$home/.ssh/`. The `-Path` would allow to specify a different path being used for providing the key-files for both, `-File` and `-All`. So for `-File` one could then either pass in the FullName, or just the file name. In terms of precedence I'd suggest to consider `-Path` as the default, which can be overriden by providing a FullName to `-File`, e.g.:`
* `Add-SshKey 'I:/dont/like/folders/starting/with/fullstop/myKey', "$home/.ssh/id_rsa"` (already works)
* `Add-SshKey -All` (already works, assuming keys reside in $home/.ssh)
* `Add-SshKey 'I:/dont/like/folders/starting/with/fullstop/myKey', 'id_rsa'` (would work with `-Path` defaulting to $home/.ssh)
* `Add-SshKey -Path 'I:/dont/like/folders/starting/with/fullstop' -All` (would use all keys residing in I:/dont/like/folders/starting/with/fullstop)

Same of course works for pageant using .ppk. 

I don't know if it's ever going to be used, but it struck me as an inconsistency in the API of one single commandlet. If you think that I should implement it, I'd propose to also make this available via `Start-SshAgent` by adding a `-KeyPath` parameter. Or maybe call it `-SshDirectory`?

c-mu:
And yes, I need to verify that piping path strings works as well ;)

dahlbyk:
> In terms of precedence I'd suggest to consider `-Path` as the default, which can be overriden by providing a FullName to `-File`...

You're spot on. Make it so!

`-Path` seems sufficient. Probably not much effort to support `Add-SshKey -All -Path 'C:/ssh1','C:/ssh2'`, too?

Issue #72, opened 2013-01-09
Update default color scheme


koistya:
Dark red color is barely visible on the dark blue background of the PowerShell window :(

`C:\Projects\SiteSDK [master +0 ~1 -0]>`

dahlbyk:
Pondering this today... I wonder if it would make sense for `$GitPromptSettings` to parse the `color.branch.<slot>` settings when the module loads to respect those if set, falling back on the [`git status` defaults](https://github.com/git/git/blob/e46fe3df01435bf523d2ab4f2755556c0e4e6f78/wt-status.c#L23-L33) if not?

dahlbyk:
I'm finally coming around... proposed workaround: https://github.com/dahlbyk/posh-git/pull/197#issuecomment-107987397

dahlbyk:
Fixed by #197.

PR #197, opened 2015-05-30
Use brighter red and green colors.  The dark red and green are difficult to read.

haroldalmon:
Use brighter colors.  The default DarkRed and Darkgreen are difficult to read.

### DIFF
--- a/GitPrompt.ps1
+++ b/GitPrompt.ps1
@@ -26,20 +26,20 @@ $global:GitPromptSettings = New-Object PSObject -Property @{
 
     BeforeIndexText           = ""
     BeforeIndexForegroundColor= [ConsoleColor]::DarkGreen
-	BeforeIndexForegroundBrightColor= [ConsoleColor]::Green
+    BeforeIndexForegroundBrightColor= [ConsoleColor]::Green
     BeforeIndexBackgroundColor= $Host.UI.RawUI.BackgroundColor
 
     IndexForegroundColor      = [ConsoleColor]::DarkGreen
-	IndexForegroundBrightColor= [ConsoleColor]::Green
+    IndexForegroundBrightColor= [ConsoleColor]::Green
     IndexBackgroundColor      = $Host.UI.RawUI.BackgroundColor
 
     WorkingForegroundColor    = [ConsoleColor]::DarkRed
-	WorkingForegroundBrightColor = [ConsoleColor]::Red
+    WorkingForegroundBrightColor = [ConsoleColor]::Red
     WorkingBackgroundColor    = $Host.UI.RawUI.BackgroundColor
 
     UntrackedText             = ' !'
     UntrackedForegroundColor  = [ConsoleColor]::DarkRed
-	UntrackedForegroundBrightColor  = [ConsoleColor]::Red
+    UntrackedForegroundBrightColor  = [ConsoleColor]::Red
     UntrackedBackgroundColor  = $Host.UI.RawUI.BackgroundColor
 
     ShowStatusWhenZero        = $true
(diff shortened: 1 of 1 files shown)


dahlbyk:
So I've been stubbornly pushing back on this (see #72) for a long time, but I think I've finally convinced myself that this needs to be fixed. However, I'd still prefer to align with the `git status` default if the console isn't using a ridiculous blue background. So...what if these settings had a bit of smarts to detect if we're in a default PowerShell host (with a "blue" `DarkMagenta` background)?

```powershell
BeforeIndexForegroundColor= if ($Host.UI.RawUI.BackgroundColor -eq [ConsoleColor]::DarkMagenta) { [ConsoleColor]::Green } else { [ConsoleColor]
::DarkGreen }
```

Is that a reasonable compromise?

haroldalmon:
Yes, but I believe you need to set the IndexForegroundColor also. I believe the BeforeIndexForegroundColor is the color of a default empty string. The resulting color is very good.

Also, I recommend you set the WorkingForegroundColor and WorkingForegroundColor to Red instead of DarkRed. I rate DarkRed as poor. Red as fair. I tried this but did not like the result. The Red is still too dark. I think the Git default should at least be Red. If people change the Git defaults to Green and Red, at least the screen is readable.

For my own PC, I gave up on the standard colors and used cyan and yellow instead of DarkRed. It work be nice if Git used Yellow for untracked and maybe unmodified files. Maybe I'll ask the Git team to consider changing the default DarkRed to Yellow.

Anyway, go with the change you suggested and consider doing the same for the two DarkRed I mentioned above.

dahlbyk:
I was just using that as an example - the same pattern would be used on all the values you changed here.

As for changing the Git defaults, that's already configurable - see https://github.com/dahlbyk/posh-git/issues/72#issuecomment-20050873.

haroldalmon:
I think you should implement your idea. Users are going to like the improvement.

dahlbyk:
Care to take a crack at it and push another commit to this PR?

haroldalmon:
Sure, will code and test on the weekend.

haroldalmon:
I had to change your sample code and decided to add some other properties so that the original properties were not dead code.  The original colors are used if the screen background is not DarkMagenta I tested this change and the code seems to work.  

I pushed the changes to my GitHub account.  Do I need to do anything else to get the code to you?  

Do you want any more changes?

GitPrompt.ps1
```diff
@@ -167,6 +171,15 @@ function Write-GitStatus($status) {
 if(!(Test-Path Variable:Global:VcsPromptStatuses)) {
     $Global:VcsPromptStatuses = @()
 }
+$s = $global:GitPromptSettings
+
+# Override some of the normal colors if the background color is set to the default DarkMagenta.
+if ($Host.UI.RawUI.BackgroundColor -eq [ConsoleColor]::DarkMagenta) { $s.BeforeIndexForegroundColor = $s.BeforeIndexForegroundBrightColor }
```
dahlbyk: This approach works for me. Can you combine these into a single `if` block?

dahlbyk:
> I pushed the changes to my GitHub account. Do I need to do anything else to get the code to you?

Nope, your update worked perfectly. Note the new commits that showed up before your comment (and on the Commits tab).

I think a single `if` would be easier to read, otherwise this seems good to :ship:.

GitPrompt.ps1
haroldalmon: I collapsed the 4 ifs into one block and pushed the change to GitHub.  I tested the change.

dahlbyk:
:sparkles: Thanks!

haroldalmon:
Thanks for walking me through the process. I saw Matthew J McCullough talk about pull requests in a youtube video a few months ago and though I would try one.

dahlbyk:
You bet. If you're looking for other contribution opportunities, check out http://up-for-grabs.net/.

PR #208, opened 2015-08-22
Use GitStatusCache when it's installed.

cmarcusreid:
Cache status information in external [git-status-cache](https://github.com/cmarcusreid/git-status-cache) process and retrieve using [git-status-cache-posh-client](https://github.com/cmarcusreid/git-status-cache-posh-client) scripts. This pull request uses git-status-cache-posh-client's Get-GitStatusFromCache function when the module is installed. This function returns status information in a format that's easily consumable within posh-git.

The external cache is built on top of libgit2 and dramatically improves performance on large repositories. Cache hits do not become more expensive with larger repositories. Performance of both cache hits and cache misses with the cache is significantly faster than current status retrieval.

Inspired by:
- https://github.com/dahlbyk/posh-git/issues/78
- https://github.com/dahlbyk/posh-git/pull/52
- https://github.com/dahlbyk/posh-git/pull/104
- https://github.com/dahlbyk/posh-git/pull/145

My primary motivator was making posh-git usable on some very large repositories I interact with every day; this solution has worked great and has been baking for a while. Recognizing that there are previous attempts at this type of improvement I've attempted to incorporate feedback from the earlier conversations.

I'm happy to address any additional comments or feedback!

---
## Performance

Cost for serving a cache hit in the cache process is generally between 0.1-0.3 ms, but this metric doesn't include the overhead involved in a full request.

The following measurements were taken on git repositories containing the specified file count. Each file was a text file containing a single sentence of text. Each case was run 5 times and the numbers reported below are averages. Each individual measurement was taken with a high resolution timer at 1 ms precision.
### Request from git-status-cache-posh-client

The measurements below capture:
- Serializing a request to JSON in PowerShell.
- Sending the JSON request over the named pipe.
- Compute the current git status (cache miss) or retrieving it from the cache (cache hit) int the cache process.
- Sending the JSON response over the named pipe back to the PowerShell client.
- Deserializing the JSON response into a PSCustomObject
##### No files modified

|  | cache miss | cache hit |
| --- | --- | --- |
| 10 file repository | 3.0 ms | 1.0 ms |
| 100 file respository | 3.0 ms | 1.0 ms |
| 1,000 file respository | 5.2 ms | 1.0 ms |
| 10,000 file respository | 22.4 ms | 1.0 ms |
| 100,000 file repository | 176.6 ms | 1.0 ms |
##### 10 files modified

|  | cache miss | cache hit |
| --- | --- | --- |
| 10 file repository | 2.8 ms | 1.0 ms |
| 100 file respository | 2.8 ms | 1.0 ms |
| 1,000 file respository | 4.4 ms | 1.0 ms |
| 10,000 file respository | 20.6 ms | 1.0 ms |
| 100,000 file repository | 176.6 ms | 1.0 ms |
### Using git-status-cache-posh-client to back posh-git

The measurements below capture all the steps from the git-status-cache-posh-client section as well as the cost for posh-git to render the prompt. This extra step has a fixed cost of around 7-10 ms.

Costs were gathered using timestamps from posh-git's debug output. Timings for posh-git without git-status-cache and git-status-cache-posh-client are included for in the "posh-git without cache" column for reference.
##### No files modified

|  | cache miss | cache hit | posh-git without cache |
| --- | --- | --- | --- |
| 10 file repository | 11.8 ms | 8.6 ms | 57.0 ms |
| 100 file respository | 11.8 ms | 8.6 ms | 62.2 ms |
| 1,000 file respository | 14.2 ms | 8.6 ms | 68.2 ms |
| 10,000 file respository | 30.8 ms | 8.0 ms | 133.8 ms |
| 100,000 file repository | 181.6 ms | 9.6 ms | 754.4 ms |
##### 10 files modified

|  | cache miss | cache hit | posh-git without cache |
| --- | --- | --- | --- |
| 10 file repository | 12.2 ms | 9.4 ms | 63.2 ms |
| 100 file respository | 12.2 ms | 9.4 ms | 65.0 ms |
| 1,000 file respository | 14.4 ms | 9.4 ms | 72.8 ms |
| 10,000 file respository | 31.0 ms | 9.4 ms | 134.0 ms |
| 100,000 file repository | 179.6 ms | 9.0 ms | 763.6 ms |


GitUtils.ps1
```diff
@@ -117,45 +117,73 @@ function Get-GitStatus($gitDir = (Get-GitDirectory)) {
         $filesUnmerged = @()
 
         if($settings.EnableFileStatus -and !$(InDisabledRepository)) {
-            dbg 'Getting status' $sw
-            $status = git -c color.status=false status --short --branch 2>$null
-        } else {
```
dahlbyk: Looks like this `else` block got lost in the shuffle?

GitUtils.ps1
```diff
@@ -117,45 +117,73 @@ function Get-GitStatus($gitDir = (Get-GitDirectory)) {
...
+                $filesUnmerged = $cacheResponse.Conflicted
+
+                $branch = $cacheResponse.Branch
+                $upstream = $cacheResponse.Upstream
+                $aheadBy = $cacheResponse.AheadBy
+                $behindBy = $cacheResponse.BehindBy
+
+                if ($cacheResponse.State) { $branch += "|" + $cacheResponse.State }
+            }
+            else {
```
dahlbyk: Go ahead and add this to the previous line.

dahlbyk:
> My primary motivator was making posh-git usable on some very large repositories I interact with every day; this solution has worked great and has been baking for a while. Recognizing that there are previous attempts at this type of improvement I've attempted to incorporate feedback from the earlier conversations.

:100: This does seem to address all my concerns with the previous implementations.

* [x] Cache is optional.
* [x] No loss of information.
* [x] Effectively zero changes to the mechanism for writing status.

To test this out:

1. Clone [git-status-cache-posh-client](https://github.com/cmarcusreid/git-status-cache-posh-client) somewhere.
2. Create a `bin/` directory with `GitStatusCache.exe` from [git-status-cache Latest](https://github.com/cmarcusreid/git-status-cache/releases/latest)
3. Add `Import-Module <path-to>\git-status-cache-posh-client\GitStatusCachePoshClient` to `$PROFILE`

So far the caching seems to be working exactly as I would expect!

dahlbyk:
I did just introduce a merge conflict on you... :crying_cat_face:

GitUtils.ps1
cmarcusreid: Yep, but removed intentionally. If I'm reading the original code correctly the empty status set by the else block serves to make the loop no-op (and has no additional references below). In this version the loop will only be hit when settings.EnableFileStatus is true (checked before checking whether to get it from the cache or from "git status), so the else block is now dead code. Please correct me if I'm misunderstanding.

cmarcusreid:
If you run install.ps1 in git-status-cache-posh-client, it should do the download on your behalf as part of setup as well as add the import to your profile. I like the idea of the warning or auto-download in the case where you enable it through another path and will follow-up on https://github.com/cmarcusreid/git-status-cache-posh-client/issues/4 (may be a few days before I get a chance).

GitUtils.ps1
dahlbyk: This is what I was thinking we needed, but you're absolutely right that it's dead code.

```powershell
        if($settings.EnableFileStatus -and !$(InDisabledRepository)) {
            if ($settings.EnableFileStatusFromCache -eq $null) {
                $settings.EnableFileStatusFromCache = (Get-Module GitStatusCachePoshClient) -ne $null
            }

            if ($settings.EnableFileStatusFromCache) {
                dbg 'Getting status from cache' $sw
                ...
            } else {
                dbg 'Getting status' $sw
                $status = ...
            }
        } else {
            $status = @()
            # Never used            
        }
```

dahlbyk:
> If you run install.ps1 in git-status-cache-posh-client, it should do the download on your behalf as part of setup as well as add the import to your profile.

Ironic that I would overlook an `install.ps1` modeled after the one we have here! You should mention that in the README. :grinning:

cmarcusreid:
README clarified. Not sure how I forgot that. :)

dahlbyk:
> README clarified. Not sure how I forgot that. :)

:+1: 

-----------------------------

I'm hoping to get a bit of [visibility on this](https://twitter.com/dahlbyk/status/636034700453109761) to kick the tires before merging, but I'm not _too_ worried about it since the caching is optional.

ajryan:
I just tested this and it rocks. Waaay faster.

andrewpmartinez:
So much nicer. posh-git was starting to go incredibly slow.

MarkusAmshove:
Hi @cmarcusreid ,

this really works well!

Can you consider adding the StashStatus from #215 into your Pipelines?

I tried to file a PR at your repositories, but I'm no C++ guy and can't get the project building :-)

Kantis:
Fails on windows 10, had to add -UseBasicParsing to all wgets in GitStatusCachePoshClient.ps1 to make install.ps1 to run without errors.

cmarcusreid:
Hey @MarkusAmshove!

Stash support was added to the cache for in https://github.com/cmarcusreid/git-status-cache/pull/12. Relevant integration for this PR is in 008836e. If you have both, posh-git should show it when the setting is enabled.

If you started using the cache before this change, you may need to update your cache bits (grab the latest git-status-cache-posh-client and call Update-GitStatusCache to download.)

You can check if the cache is returning stash information by call Get-GitStatusFromCache:

	D:\posh-git [useGitStatusCache ≡ +1 ~0 -0 ~]> git stash save "My sample stash."
	Saved working directory and index state On useGitStatusCache: My sample stash.
	HEAD is now at 008836e Read stash count from cache.
	D:\posh-git [useGitStatusCache ≡]> $global:GitPromptSettings.EnableStashStatus = $true
	D:\posh-git [useGitStatusCache ≡ (1)]> Get-GitStatusFromCache


	Version           : 1
	Path              : D:\posh-git
	RepoPath          : D:/posh-git/.git/
	WorkingDir        : D:/posh-git/
	State             :
	Branch            : useGitStatusCache
	Upstream          : origin/useGitStatusCache
	AheadBy           : 0
	BehindBy          : 0
	IndexAdded        : {}
	IndexModified     : {}
	IndexDeleted      : {}
	IndexTypeChange   : {}
	IndexRenamed      : {}
	WorkingAdded      : {}
	WorkingModified   : {}
	WorkingDeleted    : {}
	WorkingTypeChange : {}
	WorkingRenamed    : {}
	WorkingUnreadable : {}
	Ignored           : {}
	Conflicted        : {}
	Stashes           : {@{Name=stash@{0}; Sha1Id=771975d52c2507e9c3a0bb6bc151d66af03f2ea; Message=On useGitStatusCache: My sample stash.}}

cmarcusreid:
@Kantis I don't repro on Windows 10, so I wonder what's different about our environments. Could you clarify the following to help me get this fixed?

1. What exactly was the failure?
2. Was this on a call to the install script, retrieving information from the cache, or both?

Kantis:
@cmarcusreid 
1. Dont have it anymore, sadly. Sorry. I can't reproduce it any longer. Assuming it might be working without the -UseBasicParsing now after I installed the vcredist.
2. When callign the install script

cmarcusreid:
@Kantis Thanks. Are you by chance on a machine without Internet Explorer? I found [this](https://technet.microsoft.com/library/3e3dac17-3373-4d22-a54a-9d56a4a556c3(v=wps.630).aspx):

-UseBasicParsing
Uses the response object for HTML content without Document Object Model (DOM) parsing.
This parameter is required when Internet Explorer is not installed on the computers, such as on a Server Core installation of a Windows Server operating system.

----

Either way, I've logged issue https://github.com/cmarcusreid/git-status-cache-posh-client/issues/9 on git-status-cache-posh-client and will get this fixed.

Kantis:
@cmarcusreid great stuff :+1: Just got it running and this is amazing. :star:

eamodio:
Works great here too! The speed improvement is incredible!

cmarcusreid:
@dahlbyk: I just wanted to leave a note to check the current status on this PR and to make sure you're not waiting on anything from my end. Thanks!

dahlbyk:
Honestly I'm just swamped. The positive feedback bodes well, will try to get a final review for you soon. Sorry!

eamodio:
@dahlbyk any further movement on this? I've been testing it for a while and haven't run into any issues

variab1e:
How can I ensure the cache is being utilized? I have the client installed and the `Get-GitStatusCacheStatistics` works, but only increments when I directly call `Get-GitStatusFromCache`. If I do a ls within the directory it still takes ages and the stats don't change? I am using the forked posh-git.

cmarcusreid:
@variab1e: If the call count isn't incrementing outside of your manual calls, then posh-git isn't calling it. Double check that you're on the "useGitStatusCache" branch and reload PowerShell. The PR is not included in the master branch.

Once you have the PR you should have be able to explicitly disable or enable via $global:GitPromptSettings.EnableFileStatusFromCache. On the first request posh-git will automatically set this to $true if it finds the cache. See 2d3cf044b43aec5c753a2f7ae87b25009aec5f0e for details of how this is set.

Please let me know if this doesn't resolve!

MarkusAmshove:
Any updates on this? :)

cmarcusreid:
@MarkusAmshove Nothing from my end as I'm not aware of any outstanding feedback. I'm still using it daily. :)

Ultimately up to @dahlbyk on whether this makes sense to merge or whether it's more appropriate to continue to live in the fork.

dahlbyk:
> Ultimately up to @dahlbyk on whether this makes sense to merge or whether it's more appropriate to continue to live in the fork.

I do plan to merge, I just haven't had a chance to properly review. :soon: I hope.

jlevitt:
This works great! Thanks for putting it together. I was previously waiting ~3 seconds after every command. I would have had to disable posh git if not for this.

dahlbyk:
Hey @cmarcusreid, I'd like to include this in our [next release](https://github.com/dahlbyk/posh-git/milestone/1). Unfortunately, #319 and #326 introduced merge conflicts, and the latter seems like it might require a new bit in the cache response.

If you're not able to resolve the conflicts, can you [let us edit this PR](https://github.com/blog/2247-improving-collaboration-with-forks)?

dahlbyk:
> If you're not able to resolve the conflicts, can you let us edit this PR?

I took a crack at resolving the conflicts (helps a lot to separately commit the [indentation change](https://github.com/dahlbyk/posh-git/commit/b90f7dfdc8e75ea522706a837c1435899f4755ca)): https://github.com/dahlbyk/posh-git/compare/git-status-cache.

cmarcusreid:
Thanks for the ping. Taking a look at the conflicts now.

cmarcusreid:
Since my last merge of master the Arrays changed to Lists which broke the cache response building code. I've merged in your conflict resolution and updated the cache response building code to add to the Lists rather than overwrite.

dahlbyk:
Amusing that we both pushed essentially equivalent commits a few minutes apart. In the spirit of #319, I wonder which version will perform better for very large changes? @lzybkr can you tell at a glance, or should we profile it?

lzybkr:
Pipe to `foreach` is definitely slower than `AddRange`, but if possible, you should avoid the cast to `[string[]]` because that might copy the collection.

Instead, `Get-GitStatusFromCache` should ensure those properties implement `IEnumerable<string>` to make the cast unnecessary.

cmarcusreid:
The response of Get-GitStatusFromCache is built by Convert-FromJson. Unfortunately Convert-FromJson converts arrays to Object[] even though they're strings in this case. I don't think we have a way to eliminate the cast (without doing equivalent work ourselves), but I'm open to suggestions if someone sees a better solution. I've merged 3e13ee8 based on the AddRange note.

dahlbyk:
> if possible, you should avoid the cast to [string[]] because that might copy the collection.

We could use `[Linq.Enumerable]::Cast` or `[Linq.Enumerable]::OfType`, though [calling static generic methods](http://stackoverflow.com/a/4244013/54249) seems to require a workaround.

cmarcusreid:
I would expect the LINQ version to have comparable cost to the earlier foreach as it also incurs an enumeration. Probably the next step if we want to  drill into this is to profile each option and see if there's an appreciable difference. Do you feel this is a worthwhile investigation?

lzybkr:
I expect the LINQ version to always win. I improved the overhead of `ForEach-Object` some in PowerShell 5.1, but it's still much more expensive than a method call.

I suppose it's possible if you have a huge number of objects (approaching physical memory limits), then the pipeline might win, but for typical workloads sensitive to performance, you should prefer .Net methods over the pipeline where possible.

cmarcusreid:
Got it. Right now we're using the string[] cast. Any intuition on that versus the LINQ cast?

lzybkr:
They have differing semantics - LINQ cast will only do a C# cast whereas PowerShell [string[]] cast will convert each element of the array using PowerShell rules, e.g. calling ToString if the object is something other than a string.

That said, a LINQ cast avoids a copy of the array, so in theory it should be better, but it can't be called without reflection (PowerShell does not provide a way to specify a generic argument only used in the return type).

I measured the two options and LINQ is about 3x faster even with the reflection (tested with a 5000 strings in an object[]) using:

```powershell
# Create 5000 strings
$objArray = @("a") * 5000
$collection = [System.Collections.Generic.List[string]]::new()
$m = [System.Linq.Enumerable].GetMethod('Cast').MakeGenericMethod([string])
$collection.AddRange($m.Invoke($null, (,$objArray)))
```

cmarcusreid:
Excellent, thanks! I'll make this adjustment when I have a chance (probably sometime over the weekend).

dahlbyk:
Working on this now, will push to this branch.

dahlbyk:
Not thrilled with b07bce5, but it seems to work? PowerShell auto-enumerating stuff is tricky to work around. 😦

dahlbyk:
> That is a reasonable concern, but the command invocation overhead is unfortunately not small, I would need to measure to know which is worse.

Per comments on https://github.com/dahlbyk/posh-git/commit/b07bce5ba557f7cdc6a14247171aff6f0837ed9e, I've inlined the reflected method call instead of wrapping it in a PowerShell function.

-----------------------------------

I know a few folks have mentioned they have been running with this locally for a while. Update to this branch's latest and confirm your projects still work as expected?

GitUtils.ps1
```diff
@@ -146,60 +148,87 @@ function Get-GitStatus($gitDir = (Get-GitDirectory)) {
         $stashCount = 0
 
         if($settings.EnableFileStatus -and !$(InDisabledRepository)) {
-            dbg 'Getting status' $sw
-            $status = Invoke-Utf8ConsoleCommand { git -c color.status=false status --short --branch 2>$null }
-            if($settings.EnableStashStatus) {
-                dbg 'Getting stash count' $sw
-                $stashCount = $null | git stash list 2>$null | measure-object | Select-Object -expand Count
+            if ($settings.EnableFileStatusFromCache -eq $null) {
+                $settings.EnableFileStatusFromCache = (Get-Module GitStatusCachePoshClient) -ne $null
```
rkeithhill: Are we sure we want to do this?  It makes it harder to "turn off" this feature if something is misbehaving (or you want to test with it off).  Right now, you'd have to unload (remove-module GitStatusCachePoshClient) and then be careful not to execute a command that would auto-load it again.

cmarcusreid:
It looks like rename operations currently break the prompt. Other operations look fine on my machine.

Manual execution of the cast throws an InvalidOperationException. Unable to cast object of type System.Management.Automation.PSObject to type System.String.

rkeithhill:
@cmarcusreid Which rename operations?  Is this the rename of the prompt function that the Chocolatey install adds to your profile?  If so, that will be removed before 0.7.0 is released.  Importing the module now puts in place the default posh-git prompt.

cmarcusreid:
Rename operations on files in the index or working directories. :)

I believe it's just the new cast bit in this PR; I'll take a look sometime this weekend if no one beats me to  it.

GitUtils.ps1
```diff
@@ -159,13 +159,19 @@ function Get-GitStatus($gitDir = (Get-GitDirectory)) {
 
                 $indexAdded.AddRange($castStringSeq.Invoke($null, (,@($cacheResponse.IndexAdded))))
                 $indexModified.AddRange($castStringSeq.Invoke($null, (,@($cacheResponse.IndexModified))))
-                $indexModified.AddRange($castStringSeq.Invoke($null, (,@($cacheResponse.IndexRenamed | Select-Object -ExpandProperty Old))))
```
cmarcusreid: Cast was choking on the System.Management.Automation.PSObject returned by Select-Object.

GitUtils.ps1
```diff
@@ -159,13 +159,19 @@ function Get-GitStatus($gitDir = (Get-GitDirectory)) {
 
                 $indexAdded.AddRange($castStringSeq.Invoke($null, (,@($cacheResponse.IndexAdded))))
                 $indexModified.AddRange($castStringSeq.Invoke($null, (,@($cacheResponse.IndexModified))))
-                $indexModified.AddRange($castStringSeq.Invoke($null, (,@($cacheResponse.IndexRenamed | Select-Object -ExpandProperty Old))))
+                $indexRenamedOld = $cacheResponse.IndexRenamed.Old
```
cmarcusreid: I just learned that you could do this today; what is this trick called (for search purposes)?

It returns either null (zero paths), a string (one path), or an object[] (many paths).

GitUtils.ps1
cmarcusreid: It auto-selects based on the module's presence the first time through (when set to the default null value), but if you manually set it to false that'll stick on subsequent calls since it's no longer null.

rkeithhill: Got it.

GitUtils.ps1
rkeithhill: [Member enumeration](https://blogs.msdn.microsoft.com/powershell/2012/06/13/new-v3-language-features/) but it is only available in PS v3 and higher.   To make that work on PS v2, use:
```powershell
$indexRenamedOld = $cacheResponse.IndexRenamed | Foreach-Object { $_.Old }
```

cmarcusreid: Drat. I'm reverting back to the initial foreach and add version for the renamed files as I don't see a way to use the AddRange version without doing more enumerations (extracting the property then casting from objects to strings). (Recall the earlier version with Select-Object didn't work due to Select-Object's return type.)

Please feel free to make additional tweaks if there's a faster approach.

dahlbyk: I'm content with this.

dahlbyk:
🎉 

Appreciate everyone's patience with this getting this merged. Thanks for the contribution, @cmarcusreid!

PR #211, opened 2015-08-26
Improve pageant process handling

GrahamTheCoder:
Don't wait indefinitely for pageant (given that it's a service-like process that adds a tray icon and waits for input), and don't error when there are already 2 pageant processes.

**Hanging bug**
This is the old behaviour I observed (that I believe f28bb54 removes): 
Start state:
* I am on Windows 10 with pageant 0.63 running. GitUtils runs as part of my powershell profile.
* I start powershell.exe and run `$env:GIT_SSH = "plink"`
* I kill any pageant.exe that is running

Repro:
* I run  `Start-Process powershell.exe` which creates a new powershell window,  starts pageant, then hangs forever (waiting for it to return).
* I close the hanging powershell window, pageant stays running
* I run  `Start-Process powershell.exe` in my original powershell window which starts fine since pageant is already running

**Multiple-pageant bug**
When using GitExtensions on my machine I end up with two pageant processes (one is created by ShimGen), 99021d4 fixes the powershell error thrown in Start-SshAgent due to having an array returned. Stop-SshAgent also uses this, so should work with multiple pageant processes, but you need to call Stop-SshAgent once for each running SshAgent process which could be better, but is an improvement over doing nothing and erroring.

### DIFF
--- a/GitUtils.ps1
+++ b/GitUtils.ps1
@@ -303,7 +303,7 @@ function Start-SshAgent([switch]$Quiet) {
         $pageant = Get-Command pageant -TotalCount 1 -Erroraction SilentlyContinue
         $pageant = if ($pageant) {$pageant} else {Guess-Pageant}
         if (!$pageant) { Write-Warning "Could not find Pageant."; return }
-        & $pageant
+         Start-Process -NoNewWindow $pageant
     } else {
         $sshAgent = Get-Command ssh-agent -TotalCount 1 -ErrorAction SilentlyContinue
         $sshAgent = if ($sshAgent) {$sshAgent} else {Guess-Ssh('ssh-agent')}
--- a/GitUtils.ps1
+++ b/GitUtils.ps1
@@ -244,7 +244,7 @@ function Set-TempEnv($key, $value) {
 # is a running agent.
 function Get-SshAgent() {
     if ($env:GIT_SSH -imatch 'plink') {
-        $pageantPid = Get-Process | Where-Object { $_.Name -eq 'pageant' } | Select -ExpandProperty Id
+        $pageantPid = Get-Process | Where-Object { $_.Name -eq 'pageant' } | Select -ExpandProperty Id -First 1
         if ($null -ne $pageantPid) { return $pageantPid }
     } else {
         $agentPid = $Env:SSH_AGENT_PID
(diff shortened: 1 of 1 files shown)


theaquamarine:
Do we actually end up waiting for pageant? It always returns immediately for me, even if it's asking for a key to be unlocked. Some quick testing timing the options for launching pageant looks like there's not much difference.

GitUtils.ps1
```diff
@@ -303,7 +303,7 @@ function Start-SshAgent([switch]$Quiet) {
         $pageant = Get-Command pageant -TotalCount 1 -Erroraction SilentlyContinue
         $pageant = if ($pageant) {$pageant} else {Guess-Pageant}
         if (!$pageant) { Write-Warning "Could not find Pageant."; return }
-        & $pageant
+         Start-Process -NoNewWindow $pageant
```
dahlbyk: Whitespace czar has come to visit - extra space here.

dahlbyk:
Is the waiting indefinitely an artifact of the instance already running?

theaquamarine:
I don't think makes a difference, since an existing process would have caused ```Start-SshAgent``` to exit by then. Attempting to launch pageant while it's already running manually (without specifying a key to add) causes it to immediately return as normal, but with a "Pageant is already running" alert.

GrahamTheCoder:
This is the behaviour I observe on Windows 10 with pageant 0.63 starting with no pageant.exe process is running:
* I start powershell.exe and run `$env:GIT_SSH = "plink"`
* I run  `Start-Process powershell.exe` which creates a new powershell window,  starts pageant, then hangs forever waiting for it to return.
* I close the hanging powershell window, pageant stays running
* I run  `Start-Process powershell.exe` in my original powershell window
* powershell.exe starts fine since pageant is already running

dahlbyk:
I'm not a Pageant user, so I'll defer to whatever you come up with. I don't see a down-side to switching from `&` to `Start-Process`.

GrahamTheCoder:
The only downside I can think of is that it wouldn't work in Powershell 2.0 and below. The readme says 2.0 is deprecated, does that mean this is fine? Powershell 3.0 did come out 3 years ago...

theaquamarine:
No downsides and fixes issues, let's get this merged.

dahlbyk:
:100:

PR #202, opened 2015-06-23
Adding an additional git prompt symbol for a branch's tracking state

paulmarsy:
@dahlbyk pull request for issues:
#172
#190

As per the discussion in #190 
- The symbol showing the state of the working tree has been added back in
- The remote tracking state is always shown as it is now decoupled from the working tree symbol

I also added a space between the branch name and symbol for clarity.

Should the `DelimText` character be used to segment the different parts of the prompt?

In sync with no local changes
synced no changes
In sync with unstaged local changes
unstaged synced
In sync with staged local changes
staged synced
Ahead of the remote branch
ahead
Behind the remote branch
behind
Both ahead & behind the remote branch
ahead and behind

The three unicode characters used (0x2261, 0x2191 & 0x2193) render correctly in:
- ConEmu with Consolas font
- Windows PowerShell console with the default Lucida Console font (Windows 10)
- Windows PowerShell console with the default 'Raster Fonts' (Windows 8.1 & Windows Server 2012 R2)
- Windows PowerShell ISE with the default Lucida Console font (Windows 10 & 8.1) both script pane and console pane
- Command Prompt / cmd.exe (Windows 10, 8.1 & Windows Server 2012 R2)
- 'Git Bash' - mintty
- 'Git Bash' - msysgit

RE: @bazzilic & https://github.com/dahlbyk/posh-git/issues/190#issuecomment-114446204 I'm unsure where to find and test those fonts.

PR #213, opened 2015-08-26
Differentiate Untracked/Identical Branch in Prompt

### LINKED ISSUE #212: Distinguish no upstream from up-to-date
Currently if you have checked out a local branch with no upstream branch set, the prompt shows the (in my opinion, misleading) "triple equals" character. It would be nice if that character were omitted when there is no upstream set, so the user doesn't get the wrong impression that a remote branch is set up which the local branch is up-to-date to.

FWIW, this is how the [__git_ps1](https://github.com/git/git/blob/master/contrib/completion/git-prompt.sh) bash script works

dahlbyk:
Fixes #212 with follow-up from #202:

* Swaps back to previous branch colors: Ahead = Green; Untracked/Identical = Cyan
* Restores previous behavior of initializing with `BranchForegroundColor`/`BranchBackgroundColor` and then overriding as appropriate.
* Hides branch status symbol if it's not set (drops an extra space)
* Adds `$GitPromptSettings.BranchUntrackedSymbol` if a symbol is desired there (default = nothing)

image

/cc @mattdeckard @paulmarsy

### DIFF
--- a/GitPrompt.ps1
+++ b/GitPrompt.ps1
@@ -30,6 +30,7 @@ $global:GitPromptSettings = New-Object PSObject -Property @{
     LocalStagedStatusForegroundColor            = [ConsoleColor]::Cyan
     LocalStagedStatusBackgroundColor            = $Host.UI.RawUI.BackgroundColor
 
+    BranchUntrackedSymbol                       = $null
     BranchForegroundColor                       = [ConsoleColor]::Cyan
     BranchBackgroundColor                       = $Host.UI.RawUI.BackgroundColor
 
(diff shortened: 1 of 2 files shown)


mattdeckard:
Awesome. Thanks for fixing this so quickly!

dahlbyk:
Been using this all day, I think we're good.

paulmarsy:
@dahlbyk Sorry for overlooking these use cases, let me know if there are any outstanding issues or if more get raised that I can help out with.

Would there be a tangible benefit in having places where external entities depend on the module behaving or being in a specific way listed for future regression testing? Is it something that isn't likely to be a frequent enough problem to warrant the effort maintaining it or would it end up being such a large piece of documentation that the interface to the module should be refactored to be more explicit and decoupled from the internal code which should be able to be changed without affecting any consumers.

Documentation wise two options come to mind:

  - A wiki page where all the external dependencies can be listed in the same place and so can easily be found, however it could become outdated if not maintained.
  - A PowerShell comment tag next to the code which stating the consumers of the interface  and outlining what the the expectation is, this has the benefit of being close to the interface in question but could bloat the code unnecessarily and some of the expectations may not fit with a specific piece of code.

  ```
<# INTERFACE:
     # Required by: Git for Windows ( https://desktop.github.com/ )
     # Expectation: The BranchForegroundColor setting is available to be set and read to be used in feature X of GfW
     #>
     BranchForegroundColor                       = [ConsoleColor]::Cyan
  ```

Code wise maybe extract the global variables and functions into their own area, not to change them in any way but just have them moved to a specific and decoupled file from the rest of the module.

dahlbyk:
Being more explicit about public contracts is definitely a good idea. There a few neglected PRs along those lines that could probably be brought along.

That said, GitHub Desktop doesn't explicitly depend on that setting, I was just loading two different versions of posh-git (local dev plus GitHub's) and the latter was sad because `$GitPromptSettings` was already defined with incompatible settings. So really the tip there is to try and make changes there additive.

PR #215, opened 2015-09-01
Redid the Stash-Status from PR #181 from fritzh

MarkusAmshove:
Hey,

I redid the pullrequest of fritzh for the git stash status in the prompt, as I really like to have it in posh-git.

His PR wasn't requested because he also had changes to the ssh-agent thingy.

stashprompt

### DIFF
--- a/GitUtils.ps1
+++ b/GitUtils.ps1
@@ -188,7 +188,7 @@ function Get-GitStatus($gitDir = (Get-GitDirectory)) {
             HasWorking      = [bool]$working
             Working         = $working
             HasUntracked    = [bool]$filesAdded
-            stashCount      = $stashCount
+            StashCount      = $stashCount
         }
 
         dbg 'Finished' $sw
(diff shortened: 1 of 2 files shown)


GitUtils.ps1
```diff
@@ -186,6 +188,7 @@ function Get-GitStatus($gitDir = (Get-GitDirectory)) {
             HasWorking      = [bool]$working
             Working         = $working
             HasUntracked    = [bool]$filesAdded
+            stashCount      = $stashCount
```
dahlbyk: Nitpick: Can you flip the property to PascalCase?

MarkusAmshove:
Stupid mistake, I'm sorry :-)

Should be fixed now.

dahlbyk:
No need to apologize, that's why we have code review. PowerShell is case-insensitive, so it's really just for our benefit.

:8ball:

PR #220, opened 2015-09-18
Disabled stash count calculation when EnableStashStatus is $false. 

TheSned:
The stash count calculation was taking several seconds. I updated Get-GitStatus to skip the calculation if the EnableStashStatus setting is $false.


dahlbyk:
I swear we were checking this at one point... Odd.

Thanks for the fix!

PR #225, opened 2015-10-12
created fast way to update all repos in $pwd

LogoiLab:
I was tired of having to open github for desktop and clicking sync for every single one of my repositories so i made this little script to be used within a posh-git shell

### DIFF
--- a/Siphon.ps1
+++ b/Siphon.ps1
@@ -1,0 +1,35 @@
+#requires -Version 1
+function Siphon-Git
+{
+    param
+    (
+        [String]
+        $Root
+    )
+    if($Root -eq $null)
+    {
+        Write-Error -Message 'Cannot handle null paths! Aborting...'
+        Break
+    }
+    else
+    {
+        if(Test-Path $Root)
+        {
+            $Root = Convert-Path $Root
+            Set-Location -Path $Root
+            $Children = Get-ChildItem
+            foreach($Child in $Children)
+            {
+                Write-Host "Currently pulling `"$Child`"..."
+                Set-Location -Path $Child
+                git pull
+                Set-Location -Path '..\'
+            }
+        }
+        else
+        {
+            Write-Error -Message "Path `"$Root`" not found! Aborting..."
+            Break
+        }
+    }
+}


theaquamarine:
Mass pulling of repos seems kinda dangerous, or at least a way to give your mergetool a good workout. Fetching or maybe `--ff-only` would probably be a better choice, but even then...

* If you're going to mass-update everything, covering submodules with `--recurse-submodules` would probably be worthwhile.

* `git pull` only updates the current branch, and in this situation updating all local branches seems like a better option. Similar to `git-up` https://github.com/aanand/git-up https://github.com/msiemens/PyGitUp

* `Syphon` is not an approved powershell verb. https://msdn.microsoft.com/en-GB/Library/ms714428(VS.85).aspx

* At the moment, the script leaves the user in `$Root` when it completes. `Push-Location` and `Pop-Location` would be better.

Siphon.ps1
```diff
@@ -0,0 +1,35 @@
+#requires -Version 1
+function Siphon-Git
+{
+    param
+    (
+        [String]
+        $Root
```
theaquamarine: Probably worth using `[parameter(Mandatory=$true)]`, or adding a default?

Siphon.ps1
```diff
@@ -0,0 +1,35 @@
...
+    )
+    if($Root -eq $null)
```
theaquamarine: Better to use `if(-not($Root))` - otherwise running simply `Siphon-Git` makes it to line 16 when Test-Path throws an invalid data exception.

Siphon.ps1
```diff
@@ -0,0 +1,35 @@
...
+    {
+        Write-Error -Message 'Cannot handle null paths! Aborting...'
```
theaquamarine: `Throw` would be a better option, since it's a critical error.

LogoiLab:
just a little utility. just thought you might be able to do something with it is all.

Siphon.ps1
```diff
@@ -0,0 +1,35 @@
...
+        Break
+    }
+    else
+    {
+        if(Test-Path $Root)
+        {
+            $Root = Convert-Path $Root
+            Set-Location -Path $Root
+            $Children = Get-ChildItem
```
rkeithhill: `Get-ChildItem` as used here will returns files as well as directories.

rkeithhill:
I think this sort of utility is good one for your profile.  You can also make it available as a Gist for others to use but it doesn't seem to fit with posh-git's focus on getting status info and tab completion.

PR #233, opened 2015-11-16
Display git tag name in the prompt

JackDMF:
This pull request tries to handle two situations (should they be separated?):
1. I use git as a work horse for my local development, but the code is committed to SVN in the end, thus I also use posh-svn. And I use tags in git to keep track of the points where I committed to SVN or updated from SVN. I figured it might be nice to see these tags on the posh-git command line.
2. If the current branch is ahead of the master or behind it would be nice to see just how far it is. So I added the numbers to the prompt.

Tag is displayed along the branch name similar to this:

```
[master|release-1.0]
```

The tag is colored in dark gray so that it does not stand out too much.
Might help in certain situations.

The behind/ahead by number is displayed as follows:
`[master ↓12]` behind by is always printer after the arrow
`[master 1↑]` ahead by is always printer before the arrow
`[master 1↕12]` This should be consistent (ahead-by arrow behind-by)

Please review my changes and please tell me, if this is a valuable contribution or not :)

### DIFF
--- a/GitPrompt.ps1
+++ b/GitPrompt.ps1
@@ -50,9 +50,10 @@ $global:GitPromptSettings = New-Object PSObject -Property @{
     BranchBehindAndAheadStatusForegroundColor   = [ConsoleColor]::Yellow
     BranchBehindAndAheadStatusBackgroundColor   = $Host.UI.RawUI.BackgroundColor
     
-    TagForegroundColor                        = [ConsoleColor]::DarkGray
-    TagForegroundBrightColor                  = [ConsoleColor]::Gray
-    TagBackgroundColor                        = $Host.UI.RawUI.BackgroundColor
+    EnableTagDisplay                            = $true
+    TagForegroundColor                          = [ConsoleColor]::DarkGray
+    TagForegroundBrightColor                    = [ConsoleColor]::Gray
+    TagBackgroundColor                          = $Host.UI.RawUI.BackgroundColor
 
     BeforeIndexText                             = ""
     BeforeIndexForegroundColor                  = [ConsoleColor]::DarkGreen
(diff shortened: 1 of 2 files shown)


JackDMF:
Please review my changes and please tell me, if this is a valuable contribution or not :)

rkeithhill:
I really like the ahead/behind count as an option (perhaps even enabled by default?).  

The tag display should probably be in a separate PR.  I don't see as much of a common need for displaying tags associated with HEAD.  In most (perhaps all) of the source I work with, you wouldn't see tags very often.

drawfour:
Pull request #256 was already merged and did the Ahead/Behind count, enabled by default but as an option.  So this pull request should probably just be for the tag display.

rkeithhill:
@drawfour Thanks.  There are so many options in GitPromptSettings I must have missed this one.  BTW why not enable this feature by default?  Is there a perf cost?  I don't think since we should have the counts.  I really like the ahead/behind by counts in VS and VS Code.

RE the tag display.  Agreed.  Let's close this and see if the OP wants to resubmit tag display as a separate PR. However, the sort of change could probably be done on an individual basis pretty easily.  Perhaps we should provide an example of how to take the output from `Get-GitStatus`, get additional info like tag info and create the prompt text.  To make this easier, perhaps we could split up the individual bits of `Write-GitStatus` into separate functions like `Write-BranchStatus`, `Write-GitIndexStatus`, and `Write-GitWorkingDirStatus`. Usually folks only want to tweak one of these.  In this case, the OP could create his own version of `Write-BranchStatus` to display tag info.  Then in his prompt function he could call Write-GitIndexStatus and Write-GitWorkingDirStatus to create the rest of the status summary text.  This would make overriding an individual section easier to do.

You could make the functions even finer grained as well.  Perhaps break down `Write-BranchStatus` into `Write-BranchName` and `Write-BranchSyncStatus`?

drawfour:
@rkeithhill  It is enabled by default.  It has three settings: "Full" - shows both Ahead/Behind with counts and separate arrows, "Compact" which is what this pull request was originally showing with a value, double arrow, and another value (or if only ahead or behind, just a count with proper arrow), and "Minimal" which is the original way of just showing a single arrow/double arrow.  I left it enabled at Full by default.

rkeithhill:
Wait a tic. Ah that PR just get accepted yesterday?  OK that is why I'm not seeing it yet. I'm loading a newer versioned snapshot from a few days ago.  Thanks for the tip and for that PR!  This is a nice enhancement.

drawfour:
Yup.  I've been using that in my private version for about 10 months now, and I can't live without it. :)  Glad others like it too.

JackDMF:
Hey! Glad you like that change!
Don't know whether the tag display is of use for anyone - I don't use it anymore myself. So: should I close the PR?

PR #237, opened 2015-12-07
Don't use $pwd to get current location

theaquamarine:
`$pwd` is rewritable, so can be wrong. Use the same method PowerShell's (v3 and above) built-in prompt gets the location.

PowerShell 2 used `Get-Location` instead. I'm not sure why they changed, but using their current method seems to make more sense. This method works in PowerShell 2, as well.


rkeithhill:
In the process of updating profile.example.ps1, I've addressed this.

PR #240, opened 2015-12-23
readme: how to disable prompt status completely

vivainio:


rkeithhill:
I've made this change in PR #321

PR #251, opened 2016-01-31
GitTabExpansion.ps1: complete remote unique branches for 'git checkout'

mah:
See 'git help checkout', specifying a branch that doesn't exist locally
but exists in exactly one remote will set up a tracking branch.


dahlbyk:
Way overdue thanks for this!

PR #256, opened 2016-02-19
Show AheadyBy/BehindBy counts based on configuration

drawfour:
Add prompt setting BranchBehindAndAheadDisplay, which operates as follows:

```
If only behind or ahead:
    If set to "Full" or "Compact", display count along with appropriate arrow, otherwise just show the arrow

If both behind and ahead:
    If set to "Full", then display count along with arrow indicating direction of behind/ahead
    If set to "Compact", display the behind count immediately next to double-arrow, then ahead count
    If set to "Minimal" (or not recognized), display the old behavior of just the double-arrow.
```


paulmarsy:
Would there be interest in a "ColourCoded" display option, where the behind by arrow's colour changes based upon how many commits it is behind by?

Say magenta if it's behind by 1, dark magenta 2-4, red 5-10, dark red 10+?

That would convey an indication of how behind the branch is without making the prompt any bigger  if the exact count isn't of interest but more the general severity how how much it is behind by.

drawfour:
Interesting idea.  Each arrow (for Full mode) can be color coded as well, but Compact/Minimal mode only have one symbol.  That can be changed according to the overall delta (i.e. sum of Ahead/Behind is the deciding factor). Or we could have a mode between "Compact" and "Minimal" that will show individual arrows for Ahead/Behind, with a color coded option - just no numbers.

bazzilic:
It is better not to rely on colors to convey information as people are likely to have custom color schemes.

GitPrompt.ps1
```diff
@@ -82,6 +82,9 @@ $global:GitPromptSettings = New-Object PSObject -Property @{
 
     AutoRefreshIndex                            = $true
 
+    # Valid values are "Full", "Compact", and "Minimal"
```
dahlbyk: Thoughts on [defining an enum](http://social.technet.microsoft.com/wiki/contents/articles/26436.how-to-create-and-use-enums-in-powershell.aspx#DefiningAnEnum) to represent the valid states?

GitPrompt.ps1
```diff
@@ -82,6 +82,9 @@ $global:GitPromptSettings = New-Object PSObject -Property @{
 
     AutoRefreshIndex                            = $true
 
+    # Valid values are "Full", "Compact", and "Minimal"
+    BranchBehindAndAheadDisplay                 = "Full"
```
dahlbyk: Not sure about defaulting to `Full` - we're already taking up more space than we used to. Then again, leaving the counts disabled by default means most people will never see them. Anyone else have an opinion?

GitPrompt.ps1
```diff
@@ -137,41 +140,55 @@ function Write-GitStatus($status) {
...
             # This condition should not be possible but defaulting the variables to be safe
-            $branchStatusSymbol          = "?"
+            $branchStatusText            = "?"
         }
 
         Write-Prompt (Format-BranchName($status.Branch)) -BackgroundColor $branchStatusBackgroundColor -ForegroundColor $branchStatusForegroundColor
-
-        if ($branchStatusSymbol) {
-            Write-Prompt  (" {0}" -f $branchStatusSymbol) -BackgroundColor $branchStatusBackgroundColor -ForegroundColor $branchStatusForegroundColor
+
```
dahlbyk: Trailing whitespace! :trollface:

GitPrompt.ps1
drawfour: That's exactly why I left it default to Full and then added the example showing how to make it Compact.  I find that the staged files section takes way more space than the Full view, unless you're dealing with seriously out of sync branches.

GitPrompt.ps1
drawfour: I thought about it, but seemed like overkill to me, since nothing else uses enums yet, and the syntax for enums seems less Powershell-y than some weird hack.  If enums were already used, I would have defined them that way.  But I can add.

rkeithhill: You don't need to use an enum but once we drop support on master for PS v2, you can use:
```powershell
[ValidateSet('Compact','Full','Minimal')]
$BranchBehindAndAheadDisplay = "Full"
```
PS v2 support can't be dropped soon enough.  :-)

drawfour: Unfortunately, that doesn't work because it's a set of properties being initialized, not a variable.  We could move it outside the GitPromptSettings declaration as a special case, but that's kind of ugly.

rkeithhill: Doh.  Yeah, sometimes the narrow GitHub diff window bites me.  :-)

dahlbyk: For 1.0 we could consider converting `$GitPromptSettings` to a giant config function of doom, as @jaykul proposed [long ago](https://github.com/dahlbyk/posh-git/pull/111).

dahlbyk:
Works for me - belated thanks!

drawfour:
Happy to contribute! I use this tool every day.

GitPrompt.ps1
rkeithhill: That's an interesting approach.  It does have the benefit of providing tab-completion/validation on each parameter.

drawfour: That change that was proposed has a side effect of having to pass the same parameters that you've already changed to the config method, since the default value for the parameter is a hard-coded defined value.  You could still save the settings in a variable/property bag for storing, and perhaps have the default value for the parameter be the value already in the property bag, so that once changed, it will be used for subsequent sets.

Or you can do the Add-Type -TypeDefinition method to create enums you need and a class that has a bunch of public properties.  They won't have tab-completion, but they will have parameter validation.

dahlbyk: Opened https://github.com/dahlbyk/posh-git/issues/340 for discussion, targeting 1.0.

PR #266, opened 2016-03-11
Do not override user's Window Title, add to it instead.

AdilHindistan:
Current script messes user's Window Title. This does not work for people like me who has a customized Window Title to see whether I am on an elevated prompt, or which user account I am using at the moment which differs depending on the environment I am on. This change simply maintains that information while updating Window Title with Posh-Git repo information.


AdilHindistan:
Ignore this - I will update my fork first, and then send a pull request

Issue #268, opened 2016-03-14
Get-SshPath returning empty path for id_rsa


motowilliams:
[Get-SshPath](https://github.com/dahlbyk/posh-git/blob/master/GitUtils.ps1#L328)

`join-path` doesn't seem to like concatenating the paths for `$home` & `'.ssh\$File'` as it errors on the wrapped `Resolve-Path` with can't find the path `...\.ssh\$File`.  Dropping down to .NET fixes this `Resolve-Path ([io.path]::combine($home, '.ssh', $File))`

Powershell Version

```
> $PSVersionTable.PSVersion

Major  Minor  Build  Revision
-----  -----  -----  --------
5      0      10586  63
```

Windows Version

```
> [environment]::OSVersion.Version

Major  Minor  Build  Revision
-----  -----  -----  --------
10     0      10586  0
```

Issue #153, opened 2014-08-25
posh-git.psm1 sets $Env:HOME, which screws with Emacs


mrcslws:
Emacs reads your `.emacs` from `%HOME%\.emacs`. Documentation: https://www.gnu.org/software/emacs/manual/html_node/emacs/Windows-HOME.html

If %HOME% isn't set, then it uses `%APPDATA%\.emacs`.

`posh-git.psm1` specifically sets `$Env:HOME`.

So using posh-git causes Emacs to suddenly start loading the `.emacs` from a different location, which is really confusing. It took a while before I tracked down why Emacs was misbehaving, and longer before I tracked the issue down to posh-git.

(Otherwise posh-git has been very helpful!)

dahlbyk:
Proposal:

1. If `git` resolves to `cmd\git.exe`, we're good.
1. If `git` resolves as `git.cmd`, set `%HOME%` and warn with suggestion to upgrade Git.
2. If `git` resolves as `bin\git.exe`, set `%HOME% and warn with suggestion to adjust `PATH` and/or aliases to check `cmd\` first.

PetRose:
I have a somewhat related problem, during the Chef Learning exercise (chef-dk) trying to set up Github Windows, on a fresh Win 8.1 (64bit), without Chocolatey (isn´t suggested during the install guidelines).
I get a "WARNING: git command could not be found. Please create an alias or add it to your PATH."
when I spin off a Git Shell from GIT. It seems theres several issues not being taken care of, during Github Windows install:
1) Powershell User profile, not being amended with proper Git settings:
 %HOMEDRIVE%%HOMEPATH%\Documents\WindowsPowerShell\Microsoft.PowerShell_profile.ps1
such as adding Git executables (or Command) to the %PATH% environment:
    %LOCALAPPDATA%\GitHub\shell.ps1
and possibly the Module path too. I can read the 'shell.ps1' saying a few of the commands inside this would not be needed to run manually, its taken care of during install. Obviously not for me.
2) The git executables (and commands) seems scattered across %LOCALAPPDATA% with weird foldernames (I a newbie, sure). The command 'dir C:\git.exe  /s' on root drive, gives me these (my substition of LOCALAPPDATA variable inserted)
  Directory of 
%LOCALAPPDATA%\GitHub\PortableGit_c2ba306e536fdf878271f7fe636a147ff37326ad\bin
19-12-2014  05:57         1.503.744 git.exe
  Directory of
[...]

PR #269, opened 2016-03-14
Fixes Get-SshPath from returning an emtpy value

motowilliams:
Fixes issue #268

### DIFF
--- a/GitUtils.ps1
+++ b/GitUtils.ps1
@@ -328,7 +328,7 @@ function Start-SshAgent([switch]$Quiet) {
 function Get-SshPath($File = 'id_rsa')
 {
     $home = Resolve-Path (Invoke-NullCoalescing $Env:HOME ~)
-    Resolve-Path (Join-Path $home ".ssh\$File") -ErrorAction SilentlyContinue 2> $null
+    Resolve-Path ([io.path]::combine($home, '.ssh', $File))
 }
 
 # Add a key to the SSH agent


dahlbyk:
Interesting...are you getting an error other than:

```
C:\Dev\OSS\posh-git [master ≡]> Resolve-Path (Join-Path $home ".ssh\$File")
Resolve-Path : Cannot find path 'C:\Users\Keith\.ssh\id_rsa' because it does not exist.
At line:1 char:1
+ Resolve-Path (Join-Path $home ".ssh\$File")
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (C:\Users\Keith\.ssh\id_rsa:String) [Resolve-Path], ItemNotFoundException
    + FullyQualifiedErrorId : PathNotFound,Microsoft.PowerShell.Commands.ResolvePathCommand
```
While testing this myself, I stumbled on a related error:

```
Cannot overwrite variable HOME because it is read-only or constant.
At line:1 char:1
+ $home = Resolve-Path (Invoke-NullCoalescing $Env:HOME ~)
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : WriteError: (HOME:String) [], SessionStateUnauthorizedAccessException
    + FullyQualifiedErrorId : VariableNotWritable
```

We should stop trying to set `$HOME`, but I wonder if that's part of your problem?

theaquamarine:
I was just wondering how that `$home` bit was working, since it's a read-only variable. It's also an automatic variable, so wouldn't it be a better option in line 330? I'm not sure about `$env:Home`, but `~` is usually a bad plan- it's provider specific and most providers don't implement it, whereas $home always refers to the user's home folder.

Isn't the returning an empty path a function of the ErrorAction and stream redirection rather than `Join-Path` vs `[io.path]::combine`? Returning an empty path seems like a reasonable thing to do if it can't find the file specified.

I like the not hard coding backslash as path separator.

motowilliams:
Yes I was also getting the writing to home error but I was also getting the error with that Join-Path.

dahlbyk:
> I was also getting the error with that Join-Path.

What's the error message? Just trying to understand the problem - the `Join-Path` is pretty straightforward, not sure why its behavior would have changed in newer PowerShell/Windows.

rkeithhill:
Um, yeah $Home is a built-in, readonly PowerShell variable. There should be no need to use $env:Home or `~` and if you never use `~` you never need to resolve the path (to expand `~`).  Fixed by PR #325

dahlbyk:
> Um, yeah `$Home` is a built-in, readonly PowerShell variable. There should be no need to use `$env:Home` or `~` and if you never use `~` you never need to resolve the path (to expand `~`).

See https://github.com/dahlbyk/posh-git/issues/153 for some background on use of `%HOME%`.

rkeithhill:
The `%HOME%` environment variable is different from the PowerShell `$Home` variable although both should specify the same path.   Or is the issue not how you come up with the user's home dir and something else?

dahlbyk:
It may have changed in newer versions, but I think the point of using `%HOME%` was to align with Git for Windows and the MSYS utilities that depend on the same. Basically, if `%HOME%` doesn't match `$Home` there could be unexpected behavior.

PR #281, opened 2016-05-04
Update readme status explanation with settings properties

### LINKED ISSUE #131: Exclamation mark meaning
It is said in the readme that exclamation mark denotes conflicted files. However, it also appears when there is an untracked file in working dir. Is it by design?

dahlbyk:
Closes #131, following on from #202 and #221.

@paulmarsy @abbiekressner thoughts?

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -62,26 +62,31 @@ By default, the status summary has the following format:
 
     [{HEAD-name} S +A ~B -C !D | +E ~F -G !H W]
 
+* ` [` (`BeforeText`)
 * `{HEAD-name}` is the current branch, or the SHA of a detached HEAD
  * Cyan means the branch matches its remote
  * Green means the branch is ahead of its remote (green light to push)
  * Red means the branch is behind its remote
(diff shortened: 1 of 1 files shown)


abbiekressner:
Looks great, @dahlbyk! Thanks!

paulmarsy:
Sorry for the slow reply... documentation update looks good, which made me look at the history of the GitPrompt and I like the additions and changes 😄 

My only comment is the naming of the settings.. I originally went with 'symbol' in the name to help guide future development towards a single character and so smaller footprint for the prompt rather than text which while accurate about what it can contain doesn't convey the notion of keeping the prompt unobtrusive.

I don't mind either as it's common sense to avoid bloating the prompt, but the settings do have a mixture of both 'text' and 'symbol' terms across all the settings containing 1-2 characters of punctuation.  

At this point in time it's likely harmful to align them to one or the other as people will have written code to look for those setting names, but thought I would give the reasoning why I used 'symbol' so the naming would be considered for any new settings.

PR #283, opened 2016-05-23
Export Write-VcsStatus to help module autoloading

Jackbennett:
Hey, When we use your module in the profile on Cmder we try to import it as late as we can because it slows down the console startup. 

We can just call `Write-VcsStatus` when within a git repo but you don't export that command so we can't leave powershell to auto import it. This fixed that.

### DIFF
--- a/posh-git.psd1
+++ b/posh-git.psd1
@@ -25,6 +25,7 @@ PowerShellVersion = '2.0'
 FunctionsToExport = @('Invoke-NullCoalescing',
         'Write-GitStatus',
         'Write-Prompt',
+	'Write-VcsStatus',
         'Get-GitStatus',
         'Enable-GitColors',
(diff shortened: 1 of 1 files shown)


posh-git.psd1
```diff
@@ -25,6 +25,7 @@ PowerShellVersion = '2.0'
 FunctionsToExport = @('Invoke-NullCoalescing',
         'Write-GitStatus',
         'Write-Prompt',
+	'Write-VcsStatus',
```
dahlbyk: Can you flip tab to spaces for consistency? :trollface:

Jackbennett:
For future reference, could I have for pushed this? Wasn't sure since I'd already made the PR.

theaquamarine:
Yes, force pushing or updating your branch should show up here.

PR #284, opened 2016-05-24
Update the exports to include the repo status call for prompts.

Jackbennett:
Better commit than #283. Now with spaces. Should have known notepad wouldn't manage the edit. Mirror the changes in psm1 since I had the repo but I thought you only needed to put them in the psd1 if you had it.

Improves module Autoloading support in PS3.0+


theaquamarine:
lgtm, and thanks for putting the export in both files!

dahlbyk:
:sparkles: :bow:

Jackbennett:
Thank you all.

PR #286, opened 2016-05-25
Add 'push' to the list of commands to handle tab tag completion

joshgo:
Hi Keith, First time using posh-git. I'm not a powershell guy, but I was able to follow the code and make the minor modification. Hopefully, that's all that is needed. I added this because tab completion for tags exist in Git Bash when using "git push".

### DIFF
--- a/GitTabExpansion.ps1
+++ b/GitTabExpansion.ps1
@@ -234,10 +234,11 @@ function GitTabExpansion($lastBlock) {
             gitRemoteBranches $matches['remote'] $matches['ref'] $matches['branch']
         }
 
-        # Handles git push remote <branch>
-        # Handles git pull remote <branch>
-        "^(?:push|pull).* (?:\S+) (?<branch>[^\s\:]*)$" {
-            gitBranches $matches['branch']
+        # Handles git push remote <ref>
+        # Handles git pull remote <ref>
+        "^(?:push|pull).* (?:\S+) (?<ref>[^\s\:]*)$" {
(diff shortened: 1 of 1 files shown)


dahlbyk:
We handle `push` a little differently because `<origin> is required before the branch/tag. Try [here](https://github.com/joshgo/posh-git/blob/e415a03bd92d26a23e11f2b9a898ec8707c26bf3/GitTabExpansion.ps1#L233-L235) instead?

joshgo:
> ` is required before the branch/tag
Single quote/accent ? Apologies, I'm not sure what you are saying.

I looked at line 233-235 as per your suggestion, that doesn't get invoked when I type:

`git push origin {tab-key}`

[Line 239-241](https://github.com/joshgo/posh-git/blob/e415a03bd92d26a23e11f2b9a898ec8707c26bf3/GitTabExpansion.ps1#L239-L241) is what gets invoked.

If I add a call to `gitTags`, it works. Without changing the regex variable name, the call would be: `gitTags matches['branch']`. I don't like that it says 'branch', it's a little misleading...but works.

dahlbyk:
> Line 239-241 is what gets invoked.

Oh, right you are. 233-235 is specifically for `git push foo bar:<tab>`.

> I don't like that it says 'branch', it's a little misleading...but works.

Agreed. Let's change that to `ref`.

joshgo:
Changes are done. Thanks for feedback.

dahlbyk:
:sparkles: :bow:

PR #287, opened 2016-05-29
Adds help topic for module.

rkeithhill:
If you ship this then users of posh-git will be able to execute "man about_posh-git" and get the info that is currently in the readme.md file.  It is recommended that PowerShell modules provide an about help topic for users to get started with using a module.  It is also a very handy way of looking up the "legend" for the status prompt.

In case someone has their console width set to 80 chars, we limit each line's length to fewer than 80 chars.  Also, PowerShell help files are just text which get dumped to the console.  There is currently no support for markdown rendering.

### DIFF
--- a/en-US/about_posh-git.help.txt
+++ b/en-US/about_posh-git.help.txt
@@ -1,0 +1,120 @@
+﻿TOPIC
+    Posh-Git
+
+SHORT DESCRIPTION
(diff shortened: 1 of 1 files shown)


dahlbyk:
It would be neat to generate this from the readme somehow, but we'll live with double maintenance for now. Thanks for the contribution!

Issue #258, opened 2016-02-22
Powershell hangs on startup


jpierson:
Just recently I installed [Win32-OpenSSH](https://github.com/PowerShell/Win32-OpenSSH) using the [new chocolatey package](https://chocolatey.org/packages/win32-openssh/) and I noticed that after rebooting I'm no longer able to start up PowerShell due to it hanging indefinitely. Not sure if this issue is an issue more with posh-git or Win32-OpenSSH but I thought I'd open some discussion.

Here is a starting place.
https://github.com/PowerShell/Win32-OpenSSH/issues/42

jpierson:
Possibly related to https://github.com/dahlbyk/posh-git/issues/103

dahlbyk:
I assume commenting out the posh-git import in your `$PROFILE` (or starting PowerShell with `-NoProfile`) fixes the issue. What do you get from `Get-Command ssh-* | select Path`? Are you able to `ssh-add` in isolation (from PowerShell or `cmd.exe`)?

jpierson:
Actually I more surgically commented out the line `Start-SshAgent -Quiet` from [profile.example.ps1](https://github.com/dahlbyk/posh-git/blob/master/profile.example.ps1#L25) which then did allow me to get into PowerShell. Note that the fact that PowerShell ISE still worked was also a clue to where the problem was occurring.

`Get-Command ssh-* | select Path`

>
Path
----
C:\Program Files\OpenSSH-Win64\ssh-add.exe
C:\Program Files\OpenSSH-Win64\ssh-agent.exe
C:\Program Files\OpenSSH-Win64\ssh-keygen.exe
C:\Users\jeff\ShittyApplications\OpenSSH-Win32\ssh-keygen.exe
C:\Program Files\OpenSSH-Win64\ssh-keyscan.exe
C:\Program Files\OpenSSH-Win64\ssh-keysign.exe
C:\Program Files\OpenSSH-Win64\ssh-pkcs11-helper.exe

I was able to run `ssh-add` without a problem as well once I commented out the line mentioned above.

dahlbyk:
So we have a problem in [`Start-SshAgent`](https://github.com/dahlbyk/posh-git/blob/f71b6366513fdb40f2126bf1d303ebd749cac2d2/GitUtils.ps1#L297-L326). Do you see anything if you run `Start-SshAgent` (no `-Quiet`) PowerShell loads? Are you able to run `ssh-agent` by itself?fffk

[...]

Issue #292, opened 2016-06-28
Powershell stops initializing ssh-agent


nigurr:
When I try to open new windows, I am getting this message and it doesn't show up the console prompt.
This apparently started happening after installation of `Win32 ssh`
https://github.com/PowerShell/Win32-OpenSSH

```
Windows PowerShell
Copyright (C) 2015 Microsoft Corporation. All rights reserved.

agent_start pid:30344, dbg:1, child:0, pipe:0
```

Can you please let me know the issue? Powershell opens up if I comment out `poshgit` call in `$profile`

Abdullah2993:
I also have the same issue. Any thoughts?

nigurr:
Temp Fix: https://github.com/dahlbyk/posh-git/pull/295

I hard coded GIT ssh path, change it according to your system. Looking for a way to set globally.

dahlbyk:
See https://github.com/dahlbyk/posh-git/issues/258 for more information.

dahlbyk:
Closing as duplicate of #258.

dahlbyk:
This should be resolved by #586 in an upcoming v0.8 release.

PR #295, opened 2016-06-29
Fixes issues with openssh-win32 agent.

nitin88:
When Openssh-win32 agent is installed, Get-Command ssh-agent ssh-add returns openssh related commands which are not compat with Poshgit. Use ssh stuff provided with git only

Fixes https://github.com/dahlbyk/posh-git/issues/258 https://github.com/dahlbyk/posh-git/issues/292

### DIFF
--- a/GitUtils.ps1
+++ b/GitUtils.ps1
@@ -260,8 +260,8 @@ function Get-SshAgent() {
             if ($null -ne $sshAgentProcess) {
                 return $agentPid
             } else {
-                setenv 'SSH_AGENT_PID', $null
-                setenv 'SSH_AUTH_SOCK', $null
+                setenv 'SSH_AGENT_PID' $null
+                setenv 'SSH_AUTH_SOCK' $null
             }
         }
     }
@@ -373,8 +373,8 @@ function Stop-SshAgent() {
             Stop-Process $agentPid
         }
 
-        setenv 'SSH_AGENT_PID', $null
-        setenv 'SSH_AUTH_SOCK', $null
+        setenv 'SSH_AGENT_PID' $null
+        setenv 'SSH_AUTH_SOCK' $null
     }
 }
 
--- a/GitUtils.ps1
+++ b/GitUtils.ps1
@@ -222,21 +222,21 @@ function Get-AliasPattern($exe) {
    "($($aliases -join '|'))"
 }
 
-function setenv($key, $value) {
-    [void][Environment]::SetEnvironmentVariable($key, $value, [EnvironmentVariableTarget]::Process)
-    Set-TempEnv $key $value
+function Set-Env($Key, $Value) {
+    Set-Item Env:$Key -Value $Value -ErrorAction SilentlyContinue
+    Set-TempEnv -Key $Key -Value $Value
 }
 
 function Get-TempEnv($key) {
     $path = Get-TempEnvPath($key)
     if (Test-Path $path) {
         $value =  Get-Content $path
-        [void][Environment]::SetEnvironmentVariable($key, $value, [EnvironmentVariableTarget]::Process)
+        Set-Item Env:$key -Value $value -ErrorAction SilentlyContinue
     }
 }
 
 function Set-TempEnv($key, $value) {
-    $path = Get-TempEnvPath($key)
+    $path = Join-Path ($Env:TEMP) ".ssh\$key.env"
     if ($value -eq $null) {
         if (Test-Path $path) {
             Remove-Item $path
@@ -245,11 +245,6 @@ function Set-TempEnv($key, $value) {
         New-Item $path -Force -ItemType File > $null
         $value | Out-File -FilePath $path -Encoding ascii -Force
     }
-}
-
-function Get-TempEnvPath($key){
-    $path = Join-Path ([System.IO.Path]::GetTempPath()) ".ssh\$key.env"
-    return $path
 }
 
 # Retrieve the current SSH agent PID (or zero). Can be used to determine if there
@@ -261,14 +256,13 @@ function Get-SshAgent() {
     } else {
         $agentPid = $Env:SSH_AGENT_PID
         if ($agentPid) {
-            $sshAgentProcess = Get-Process | Where-Object { $_.Id -eq $agentPid -and $_.Name -eq 'ssh-agent' }
-            if ($null -ne $sshAgentProcess) {
+            $pid = Get-Process -Id $agentPid -ErrorAction SilentlyContinue
+            if ($pid) {
                 return $agentPid
-            } else {
-                setenv 'SSH_AGENT_PID' $null
-                setenv 'SSH_AUTH_SOCK' $null
-            }
+            } 
         }
+        Set-Env -Key 'SSH_AGENT_PID' -Value $null
+        Set-Env -Key 'SSH_AUTH_SOCK' -Value $null
     }
 
     return 0
@@ -287,15 +281,34 @@ function Find-Pageant() {
 
 # Attempt to guess $program's location. For ssh-agent/ssh-add.
 function Find-Ssh($program = 'ssh-agent') {
+    $sshLocation = Get-Command $program -TotalCount 1 -ErrorAction SilentlyContinue
+    if($sshLocation){
+        return $sshLocation
+    }
+
     Write-Verbose "$program not in path. Trying to guess location."
     $gitItem = Get-Command git -Erroraction SilentlyContinue | Get-Item
     if ($gitItem -eq $null) { Write-Warning 'git not in path'; return }
 
     $sshLocation = join-path $gitItem.directory.parent.fullname bin/$program
-    if (get-command $sshLocation -Erroraction SilentlyContinue) { return $sshLocation }
+    if (get-command $sshLocation -Erroraction SilentlyContinue) { 
+        return $sshLocation 
+    }
 
     $sshLocation = join-path $gitItem.directory.parent.fullname usr/bin/$program
-    if (get-command $sshLocation -Erroraction SilentlyContinue) { return $sshLocation }
+    if (get-command $sshLocation -Erroraction SilentlyContinue) { 
+        return $sshLocation 
+    }
+}
+
+function Find-Service($service = 'ssh-agent') {
+     # Find if wind32-openssh available and use it
+    return Get-Service -Name $service -ErrorAction SilentlyContinue
+}
+
+function Find-ServicePID($service = 'ssh-agent') {
+    # Find if wind32-openssh available and use it
+    return (Get-Service $service -ErrorAction SilentlyContinue | Get-Process).Id
 }
 
 # Loosely based on bash script from http://help.github.com/ssh-key-passphrases/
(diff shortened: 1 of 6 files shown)


GitUtils.ps1
```diff
@@ -282,15 +281,30 @@ function Find-Pageant() {
 
 # Attempt to guess $program's location. For ssh-agent/ssh-add.
 function Find-Ssh($program = 'ssh-agent') {
+    $sshLocation = 'C:\Program Files\Git\usr\bin\'
```
nitin88: This has to be given as preference from user perspective. not sure where to keep this and how to import that var here. temp hardcoded

dahlbyk: Is this actually necessary? I would expect the existing code in `Find-Ssh` to be able to resolve the correct location relative to the discovered `git` command.

GitUtils.ps1
```diff
@@ -256,8 +256,7 @@ function Get-SshAgent() {
     } else {
         $agentPid = $Env:SSH_AGENT_PID
         if ($agentPid) {
-            $sshAgentProcess = Get-Process | Where-Object { $_.Id -eq $agentPid -and $_.Name -eq 'ssh-agent' }
-            if ($null -ne $sshAgentProcess) {
+            if (((Get-Process -Id $agentPid -FileVersionInfo -ErrorAction SilentlyContinue).Filename).Contains("Git")) {
```
dahlbyk: If `$agentPid` is valid but the process has been killed, this will blow up when you try to access `Filename`.

nigurr: (Get-Process -Id 11111111 -FileVersionInfo -ErrorAction SilentlyContinue).Filename

doesn't throw any error. you will get empty data

nitin88:
Reworked on this. Win32-OpenSsh install ssh as service. leverage that functionality. Posh-git no longer requires Git Ssh agent.

nitin88:
@dahlbyk please review the same.

GitUtils.ps1
```diff
@@ -307,6 +301,16 @@ function Find-Ssh($program = 'ssh-agent') {
     }
 }
 
+function Find-Service($service = 'ssh-agent') {
+     # Find if wind32-openssh available and use it
+    return Get-Service -Name $service -ErrorAction SilentlyContinue
+}
+
+function Find-SericePID($service = 'ssh-agent') {
```
dahlbyk: Typo: Serice

dahlbyk:
I like this approach better than the explicit check for Git in the `ssh-agent` path. I'm not a Win32-OpenSSH user (yet, I suppose), but I'm hoping we can get broader confirmation that this works as expected.

Thanks for taking this on!

GitUtils.ps1
```diff
@@ -282,15 +281,34 @@ function Find-Pageant() {
...
+}
+
+function Find-Service($service = 'ssh-agent') {
+     # Find if wind32-openssh available and use it
+    return Get-Service -Name $service -ErrorAction SilentlyContinue
+}
+
+function Find-SericePID($service = 'ssh-agent') {
+    # Find if wind32-openssh available and use it
+    return (get-wmiobject win32_service | where { $_.name -eq $service}).processID
```
theaquamarine: `(Get-Service $Service | Get-Process).Id` will be much faster than this.

GitUtils.ps1
```diff
@@ -312,13 +330,19 @@ function Start-SshAgent([switch]$Quiet) {
         if (!$pageant) { Write-Warning "Could not find Pageant."; return }
         Start-Process -NoNewWindow $pageant
     } else {
-        $sshAgent = Get-Command ssh-agent -TotalCount 1 -ErrorAction SilentlyContinue
-        $sshAgent = if ($sshAgent) {$sshAgent} else {Find-Ssh('ssh-agent')}
-        if (!$sshAgent) { Write-Warning 'Could not find ssh-agent'; return }
+        $sshService = Find-Service('ssh-agent')
+        if($sshService) {
+            Start-Service ssh-agent -ErrorAction SilentlyContinue 
+            $pid = Find-SericePID('ssh-agent')
```
theaquamarine: You've already got the service as `$sshService` so could avoid calling `Find-ServicePID` with `($sshService | Get-Process).Id`

GitUtils.ps1
```diff
@@ -349,10 +372,21 @@ function Add-SshKey() {
...
+            if($agentPath){
+                $sshPath = Split-Path ($agentPath) -Parent
+            }
+            if($sshPath){
+                $sshAdd = Get-Command (Join-Path $sshPath 'ssh-add')
+            }
+        }
+        if(!$sshAdd){
+            $sshAdd = Find-Ssh('ssh-add')
+        }
```
theaquamarine: This feels overcomplicated- if we've got a running ssh-agent, `Get-Command ssh-add` or `Find-Ssh('ssh-add')` should return the matching ssh-add so just calling `ssh-add` should work. If not, maybe the logic to find the correct binary for the agent should be in Find-Ssh?

GitUtils.ps1
```diff
@@ -363,6 +397,15 @@ function Add-SshKey() {
     }
 }
 
+function Get-SshAdd()
```
theaquamarine: Doesn't appear to be used?

DarwinJS:
I'm late to the game here, but here is how I figure out if a [1] running instance of a process, is [2] from the win32-openssh folder: 

    $TargetFolder = "C:\Program Files\OpenSSH-Win64"
    $SSHServiceInstanceExistsAndIsOurs = ([bool]((Get-WmiObject win32_service | ?{$_.Name -ilike 'sshd'} | select -expand PathName) -ilike "*$TargetFolder*"))

This would let you understand which package the *running* service is coming from.

Also - I think the SSH in git is much more debugged than Win32-OpenSSH, especially for git-ish stuff.  So I would stick with it since you depend on git being installed anyway.

Maybe if another ssh is loaded you warn the user that poshgit may not work with it?

GitUtils.ps1
nigurr: Cleaned up :)

GitUtils.ps1
nigurr: If you look carefully, $agentPid is set [whether it's openssh or git ssh], so the new block of code always gets executed and returns right ssh-add. the old code is unnecessary Find-Ssh. I just kept it for the sake of not breaking anyone in the worst case 😄

GitUtils.ps1
nigurr: Thanks for the tip. adopted this

GitUtils.ps1
nigurr: corrected

nigurr:
@DarwinJS 
1. I don't want to hard code any path. the path might change in the next build.
2. ssh is independent tool. the one which comes in "git" is open ssh one. Microsoft forked it and baking into Windows.
3. Upcoming releases of Windows, *will* have "openssh-win32" in it's core.

So Its good to support and if you ever look at memory/cpu utilization openssh-win32 is like 80% better than git one

nitin88:
@DarwinJS 
1. I don't want to hard code any path. the path might change in the next build.
2. ssh is independent tool. the one which comes in "git" is open ssh one. Microsoft forked it and baking into Windows.
3. Upcoming releases of Windows, *will* have "openssh-win32" in it's core.

So Its good to support and if you ever look at memory/cpu utilization openssh-win32 is like 80% better than git one

GitUtils.ps1
nitin88: If you look carefully, the new logic always gets the ssh-add correctly [with openssh or without]. Just to not to break anyone in the worst case scenario, kept the old logic, we should cleanup in next release or so if nothing is breaking 👍

DarwinJS:
Good info.  Thanks.

GitUtils.ps1
```diff
@@ -222,20 +222,20 @@ function Get-AliasPattern($exe) {
...
+    Set-Item Env:$Key -Value $Value -ErrorAction SilentlyContinue
+    Set-TempEnv -Key $Key -Value $Value
 }
 
 function Get-TempEnv($key) {
     $path = Join-Path ($Env:TEMP) ".ssh\$key.env"
     if (Test-Path $path) {
         $value =  Get-Content $path
-        [void][Environment]::SetEnvironmentVariable($key, $value, [EnvironmentVariableTarget]::Process)
+        Set-Item Env:$key -Value $value -ErrorAction SilentlyContinue
```
dahlbyk: Why switch from `[Environment]::SetEnvironmentVariable(...)` to `Set-Item`? I don't remember why I used `SetEnvironmentVariable()` in the first place, but I'm curious what the difference is from a OpenSsh-Win32 perspective?

dahlbyk:
One last question, but I think this is looking pretty good. Can you pull in dahlbyk:master and resolve the conflicts?

nitin88:
downmerged from master

dahlbyk:
I think we're good here.

@theaquamarine I believe this resolves the environment variable issue in #302?

rkeithhill:
Another thing to consider is how this should work on PowerShell v6 on Linux and macOS.

dahlbyk:
@nitin88 sorry to have let this gone stale - can you either resolve the conflict or [let us edit this PR](https://github.com/blog/2247-improving-collaboration-with-forks)?

> Another thing to consider is how this should work on PowerShell v6 on Linux and macOS.

Does `Get-Service` exist on Linux/macOS? If it doesn't, we should be able to add guards rather easily ahead of [1.0 support for Core](https://github.com/dahlbyk/posh-git/issues/328).

From a cross-platform standpoint, is there a useful distinction between these?

```powershell
Set-Item Env:$Key -Value $Value -ErrorAction SilentlyContinue
[void][Environment]::SetEnvironmentVariable($Key, $Value)
```

nitin88:
@dahlbyk both are same. One is pure powershell implementation while the other one is C# implementation.

I am not sure whether 'Get-Service' exists in *nix. Ideally the code is consistent between Windows/*nix except calling layer underneath has changed.

Note: Now you can contribute

rkeithhill:
As of 6.0.0-alpha.14, `Get-Service` is not available on Linux.

brunoyb:
Hi! With Windows 10 April 2018 Update Win32-OpenSSH was installed automatically for me, so the ssh-agent stopped working properly (I have Start-SshAgent in my profile.ps1). As a workaround I simply removed the Win32-OpenSSH tools from PATH but it would be nice if it could work normally, and I believe this is what this PR is doing! Any ETA on when we can go forward with this? Is there something we can do?

Thanks!

dahlbyk:
I suppose this impacting more people via Windows Update is a good reason to wrap up the ssh-agent situation. The goal for v1 is to move everything ssh-ish into a separate module, but it would be good if everything worked reasonably in v0.x.

There are a handful of open ssh PRs that are significantly out of date, if anyone cares to consolidate/update them targeting the `v0` branch.

rkeithhill:
FWIW I'm using the new OpenSSH implementation in Windows and as a result I do not need any of the SSH support in posh-git.  I set up the new ssh-agent to start automatically and start it.  Then, when you add your private keys with the OpenSSH version of ssh-add, that key will be remembered across reboots and because ssh-agent is started as a service, you don't need to start it in your PowerShell profile anymore.

In order to get Git to use the new OpenSSH, you will want to set this system or user env var:
```
GIT_SSH = C:\Windows\system32\OpenSSH\ssh.exe  
```
Double check that path.  I'm actually using the latest version of OpenSSH for Windows installed via `choco install OpenSSH`.  If you use Chocolatey to install the latest version, then use the path it installs to i.e. `C:\Program Files\OpenSSH-Win64\ssh.exe`.

GitUtils.ps1
```diff
@@ -287,15 +281,34 @@ function Find-Pageant() {
...
     $sshLocation = join-path $gitItem.directory.parent.fullname usr/bin/$program
-    if (get-command $sshLocation -Erroraction SilentlyContinue) { return $sshLocation }
+    if (get-command $sshLocation -Erroraction SilentlyContinue) { 
+        return $sshLocation 
+    }
+}
+
+function Find-Service($service = 'ssh-agent') {
+     # Find if wind32-openssh available and use it
+    return Get-Service -Name $service -ErrorAction SilentlyContinue
```
rkeithhill: All references to Get-Service and Start-Service need to be behind a "Windows-only" firewall.

dahlbyk:
Closing in deference to #586. Sorry we didn't review this in a more timely manner!

PR #297, opened 2016-08-05
Fixed syntax error on setenv calls.

hjoelr:
While trying to understand how posh-git handles environment variables, I found a bug when removing the environment variables. The function calls were separating parameters out with commas when they should be separated by spaces.

### DIFF
--- a/GitUtils.ps1
+++ b/GitUtils.ps1
@@ -260,8 +260,8 @@ function Get-SshAgent() {
             if ($null -ne $sshAgentProcess) {
                 return $agentPid
             } else {
-                setenv 'SSH_AGENT_PID', $null
-                setenv 'SSH_AUTH_SOCK', $null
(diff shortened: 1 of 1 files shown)


dahlbyk:
Good catch!

hjoelr:
@dahlbyk No problem. Glad to give back.

I have some questions about the implementation of the SSH_AGENT_PID and SSH_AUTH_SOCK environment variables in posh-git and I'm trying to figure out where best to post those. Should I create an issue on them?

dahlbyk:
> I have some questions about the implementation of the SSH_AGENT_PID and SSH_AUTH_SOCK environment variables in posh-git and I'm trying to figure out where best to post those. Should I create an issue on them?

Yeah, opening an issue should get you the most :eyes: on the question. Feel free to `blame` that part of the code - I don't remember who implemented that bit.

PR #300, opened 2016-08-20
Updated PowerShellGet info + editing

juneb:
Changed passive to active voice. 
Updated PowerShellGet info. 
Clarified execution policy.

I wanted to clarify using an alias instead of editing $env:Path, but I didn't understand it either. If you can explain it to me, I'll update this further. Thanks!

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -1,44 +1,56 @@
 posh-git
 ========
 
-A set of PowerShell scripts which provide Git/PowerShell integration
-
+A set of PowerShell scripts that integrate Git and PowerShell.
 ### Prompt for Git repositories
-   The prompt within Git repositories can show the current branch and the state of files (additions, modifications, deletions) within.
+   The prompt in Git repositories show the current branch and the status of files (additions, modifications, deletions).
 
 ### Tab completion
    Provides tab completion for common commands when using git.
    E.g. `git ch<tab>` --> `git checkout`
 
-Usage
+Notes
 -----
+Posh-git adds variables to your session to let you customize it, including `$GitPromptSettings`, `$GitTabSettings`, and `$TortoiseGitSettings`. For an example of integrating the tab completion and/or git prompt into your profile, see `profile.example.ps1`. 
 
-See `profile.example.ps1` as to how you can integrate the tab completion and/or git prompt into your own profile.
-Prompt formatting, among other things, can be customized using `$GitPromptSettings`, `$GitTabSettings` and `$TortoiseGitSettings`.
+Note on performance: Displaying file status in the git prompt for a very large repo can be prohibitively slow. Rather than turn off file status entirely, you can disable it on a repo-by-repo basis by adding individual repository paths to $GitPromptSettings.RepositoriesInWhichToDisableFileStatus.
 
-Note on performance: displaying file status in the git prompt for a very large repo can be prohibitively slow. Rather than turn off file status entirely, you can disable it on a repo-by-repo basis by adding individual repository paths to $GitPromptSettings.RepositoriesInWhichToDisableFileStatus.
 
-Installing via PsGet
---------------------
+Installing Posh-git with PowerShellGet
+-----------------------------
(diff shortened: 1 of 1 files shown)


theaquamarine:
Thanks! The alias stuff is just to make an alias for `git` to the git executable, to avoid adding it to the path.

readme.md
```diff
@@ -1,44 +1,56 @@
 posh-git
 ========
 
-A set of PowerShell scripts which provide Git/PowerShell integration
-
+A set of PowerShell scripts that integrate Git and PowerShell.
 ### Prompt for Git repositories
-   The prompt within Git repositories can show the current branch and the state of files (additions, modifications, deletions) within.
+   The prompt in Git repositories show the current branch and the status of files (additions, modifications, deletions).
```
theaquamarine: "shows" rather than "show"?

juneb:
Sorry about the typo. Want me to fix that or will you? 

About the alias (sorry - I'm very literal :) ), where do you put the alias? 

Also, I'm using GitHub for Windows + posh-git so I don't have to deal with authentication. As a result, my Git.exe is in:

$home\appdata\local\github\PortableGit_624c8416ee51e205b3f892d1d904e06e6f3c57c8\mingw32\bin

Are there special instructions for installing and updating this configuration?

theaquamarine:
I guess if you end up adding another commit, you might as well do it, but I'll fix the typo if not.

I think the alias would be added to your profile.

The Git Shell GitHub for Windows provides adds that folder to the path (but only in Git Shell) and includes posh-git, so as long as you're happy using Git Shell, installation's handled and I assume updating GitHub for Windows would update posh-git.

For posh-git and GH4W outside of Git Shell, I'm not sure since I don't know much about GH4W, particularly what it does for authentication. It seems a bit complicated and my inclination would be separate installs of posh-git and git, but:

- If you want to install posh-git separately and use GH4W's git, it'd be a case of adding git's location to $env:Path (and updating the path every time GH4W updates git). It wouldn't really be much different from manually installing posh-git and git, from posh-git's perspective.

- If you want to use GH4W's posh-git and git, you'd have to run posh-git from GH4W's install folder and add git to the path, or run Git Shell's environment setup script, `%localappdata%\GitHub\shell.ps1`. Updating it would mean manually replacing the files, or relying on GH4W updates.

readme.md
```diff
@@ -1,44 +1,56 @@
...
+Previous version said that PowerShell 2.0 is deprecated. Although people have requested that it be deprecated, I don't think it has been, so I removed that line.
+-->
+
+0. Verify that you have PowerShell 2.0 or greater (`$PSVersionTable.PSVersion`). PowerShell 3.0 is preferred. 
+
+1. Use `Get-ExecutionPolicy` to verify that the local execution policy allows you to run scripts. `RemoteSigned` or `Unrestricted` are sufficient. 
+To change the execution policy so that scripts can run, start PowerShell with the 'Run as Administrator' option and then use `Set-ExecutionPolicy`. For example: `Set-ExecutionPolicy RemoteSigned -Scope CurrentUser -Confirm`.
 
-1. Verify execution of scripts is allowed with `Get-ExecutionPolicy` (should be `RemoteSigned` or `Unrestricted`). If scripts are not enabled, run PowerShell as Administrator and call `Set-ExecutionPolicy RemoteSigned -Scope CurrentUser -Confirm`.
+2. Verify that `git` can run in the PowerShell console. If you get a  "command not found" error, you need to add a git alias or add `%ProgramFiles(x86)%\Git\cmd` (or `%ProgramFiles%\Git\cmd` if you're still on 32-bit) to your `PATH` environment variable.
```
rkeithhill: Actually Git installs into `%ProgramFiles%` on Windows.  In fact, if Git has installed itself into that location then **it** normally adds a PATH entry during installation.  I'd say if the command is not found, we should recommend folks go to http://git-scm.com and install the latest version of Git.

rkeithhill:
@juneb There are now conflicts with this PR.  Can you resolve the conflicts?  If you don't want to mess with that, I can make your changes onto the latest README.md.  Just let me know.

juneb:
Sure. I'll work on it tonight.

rkeithhill:
June, I've migrated some of your changes into a revamped README.md that I just merged in. I'm going to close this PR but please feel free to look over the updated README.md and submit a new PR with fixes based on the new content.

PR #303, opened 2016-08-25
Save current working branch to variable

Alduhoo:
This will save the current branch to a variable, useful for scripting purposes.
This will also enable commands like `git push -u origin $GitBranch`

### DIFF
--- a/GitPrompt.ps1
+++ b/GitPrompt.ps1
@@ -160,6 +160,7 @@ function Write-GitStatus($status) {
             $branchStatusSymbol          = "?"
         }
 
+        $Global:GitBranch = $status.Branch
         Write-Prompt (Format-BranchName($status.Branch)) -BackgroundColor $branchStatusBackgroundColor -ForegroundColor $branchStatusForegroundColor
 
         if ($branchStatusSymbol) {


JayBazuzi:
Isn't this something scripts should figure out themselves?

Alduhoo:
Probably, but for interactive sessions I still think this is useful. It saves me time from having to visually select the branch name in the prompt and then copying to whatever command I want to run.

dahlbyk:
> This will also enable commands like git push -u origin $GitBranch

Tip: this is equivalent to `git push -u origin HEAD`.

----

We already [save `$Global:GitStatus`](https://github.com/dahlbyk/posh-git/blob/c39da7811fe37d5ec1d9b5b1ab4f6f40abc8bf40/GitPrompt.ps1#L269-L270), so I'm not really inclined to further pollute the global namespace by default. You can either customize your [`prompt` function](https://github.com/dahlbyk/posh-git/blob/c39da7811fe37d5ec1d9b5b1ab4f6f40abc8bf40/profile.example.ps1#L12-L21) with `$Global:GitBranch = $GitStatus.Branch` or just use `$GitStatus` directly.

PR #304, opened 2016-09-15
Use ANSI color codes if SupportsVirtualTerminal

### LINKED ISSUE #282: posh-git prompt makes PowerShell transcripts less readable
With posh-git:

```
>

C:\devcd log-ssh
 [
master
 ≡
]
>

C:\dev\log-sshls


    Directory: C:\dev\log-ssh


-a----        4/28/2016   6:21 PM          35815 LICENSE
-a----        4/28/2016   6:25 PM           4477 Log-SSH.pl
-a----        4/28/2016   6:21 PM             74 README.md

```

Without posh-git:

```
PS C:\dev> cd log-ssh
PS C:\dev\log-ssh> ls


    Directory: C:\dev\log-ssh


Mode                LastWriteTime     Length Name
----                -------------     ------ ----
-a----       2016-04-28     18:21      35815 LICENSE
-a----       2016-04-28     18:25       4477 Log-SSH.pl
-a----       2016-04-28     18:21         74 README.md

```

dahlbyk:
I have been nerd-sniped: https://github.com/Jaykul/PowerLine/issues/1.

The notion of `Write-Prompt` either returning a `string` (maybe empty) or actually doing a `Write-Host` seems _quite_ wrong, but I'm not sure how else to handle it without breaking users of older hosts. Any ideas, @rkeithhill?

Opened as WIP because I'm seeing some weird behavior in the prompt after running Git commands that use the pager. Haven't had a chance to dig into what might be going on there, but I'm open to ideas.
powershell_2016-09-14_23-39-18

Closes #282

---

Also, for SEO purposes (because I was surprised that I had to write them myself), herein can be found PowerShell functions to convert `ConsoleColor` to [ANSI color escape codes/sequences](https://en.wikipedia.org/wiki/ANSI_escape_code#Colors).

### DIFF
--- a/GitPrompt.ps1
+++ b/GitPrompt.ps1
@@ -273,7 +273,10 @@ if ($Host.UI.RawUI.BackgroundColor -eq [ConsoleColor]::DarkMagenta) {
     $s.WorkingForegroundColor               = $s.WorkingForegroundBrightColor 
 }
 
-function Global:Write-VcsStatus { $Global:VcsPromptStatuses | foreach { & $_ } }
+function Global:Write-VcsStatus {
+  Set-ConsoleMode -ANSI
+  $Global:VcsPromptStatuses | foreach { & $_ }
+}
 
 # Add scriptblock that will execute for Write-VcsStatus
 $PoshGitVcsPrompt = {
--- a/GitPrompt.ps1
+++ b/GitPrompt.ps1
@@ -106,11 +106,18 @@ if (Get-Module NuGet) {
 }
 
 function Write-Prompt($Object, $ForegroundColor, $BackgroundColor = -1) {
+    if ($Host.UI.SupportsVirtualTerminal) {
+      $e = [char]27 + "["
+      $f = Get-ForegroundVirtualTerminalSequence $ForegroundColor
+      $b = Get-BackgroundVirtualTerminalSequence $BackgroundColor
+      return "${f}${b}${Object}${e}0m"
+    }
     if ($BackgroundColor -lt 0) {
         Write-Host $Object -NoNewLine -ForegroundColor $ForegroundColor
     } else {
         Write-Host $Object -NoNewLine -ForegroundColor $ForegroundColor -BackgroundColor $BackgroundColor
     }
+    return ""
 }
 
 function Format-BranchName($branchName){
@@ -126,8 +133,9 @@ function Format-BranchName($branchName){
 
 function Write-GitStatus($status) {
     $s = $global:GitPromptSettings
+    $p = ''
     if ($status -and $s) {
-        Write-Prompt $s.BeforeText -BackgroundColor $s.BeforeBackgroundColor -ForegroundColor $s.BeforeForegroundColor
+        $p += Write-Prompt $s.BeforeText -BackgroundColor $s.BeforeBackgroundColor -ForegroundColor $s.BeforeForegroundColor
 
         $branchStatusSymbol          = $null
         $branchStatusBackgroundColor = $s.BranchBackgroundColor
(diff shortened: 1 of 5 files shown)


dahlbyk:
> Opened as WIP because I'm seeing some weird behavior in the prompt after running Git commands that use the pager. Haven't had a chance to dig into what might be going on there, but I'm open to ideas.

It's pretty clearly an issue with `less.exe` or PowerShell's virtual terminal support, but I'm way out of my depth on troubleshooting this (cc @shiftkey). A child `powershell` process affects its parent session, but a new child shell is unaffected. :unamused:
powershell_2016-09-15_23-22-30

I have confirmed that the `LESS` and `TERM` environment variables are not set. Various fiddling hasn't found any magic combinations that fix it.

The first hit on "powershell virtual terminal" brings up https://github.com/PowerShell/PowerShell/issues/1177, which references `less.exe`. Will drop a note over there, too.

dahlbyk:
Hrm... wondering if `Write-ColoredHost` from https://blog.fastconnect.fr/?p=568 would be a better approach? (via @aaronmcdavid in https://github.com/dahlbyk/posh-git/issues/282#issuecomment-218876824)

@mcoolive, I don't see a license on your blog - can we have permission to use that code? (BTW, your blog gives a certificate error: `NET::ERR_CERT_COMMON_NAME_INVALID`.)

Thinking implementation would look something like...

1. Replace `Write-Prompt` with `Get-ColoredString`
2. Extract `Get-GitStatusPrompt` from `Write-GitStatus`, keeping the latter for backward compatibility using `Write-ColoredHost`
3. Export `Get-GitStatusPrompt` and `Get-VcsStatusPrompt`

lzybkr:
You are hitting the issue I reported (well, it's my bug) with escape sequences being emitted after running `git`.

The workaround for now is to set the console mode. It's a simple P/Invoke, I have a sample (that does more than you need) here: https://gist.github.com/lzybkr/f2059cb2ee8d0c13c65ab933b75e998c

In my personal `prompt`, I simply call `Set-ConsoleMode -ANSI` using the above gist, I think it's reasonable to do the equivalent in a function like `Get-VcsStatusPrompt`.

GitPrompt.ps1
```diff
@@ -105,12 +105,66 @@ if (Get-Module NuGet) {
     $WindowTitleSupported = $false
 }
 
+# Color codes from https://msdn.microsoft.com/en-us/library/windows/desktop/mt638032(v=vs.85).aspx
+function Get-ForegroundVirtualTerminalSequence($Color) {
+    $e = [char]27 + "["
+    switch ($Color) {
+      ([ConsoleColor]::Black)       { "${e}30m" }
```
lzybkr: You should use `break`, otherwise every case is needlessly tested.
You should also consider ordering the cases so that the more commonly used colors appear first - PowerShell does not generate efficient switch tables so each case is tested before the preceding cases.

lzybkr: Actually, using an array would be more efficient - indexing into the array will be much faster.

Jaykul:
@eweilnau is working on a [Write-AnsiHost](https://github.com/eweilnau/PowerLine/commit/f09111bd16de2bfc82b27c065ce0daed08da78a8) if you want to go that route -- trying to implement a _little_ more than just colors (i.e. movement).

GitPrompt.ps1
dahlbyk: Check my latest? (I moved these functions to a different file, too.)

dahlbyk:
> In my personal `prompt`, I simply call `Set-ConsoleMode -ANSI` using the above gist, I think it's reasonable to do the equivalent in a function like `Get-VcsStatusPrompt`.

:+1:

What would be the preferred way to guard against trying to compile or use that type on non-Windows?

----

> @eweilnau is working on a Write-AnsiHost if you want to go that route (the idea being to accept escaped text and output using write-host for compatibilty) -- he's trying to implement a little more than just colors (i.e. movement).

:cool: not sure if it would make sense to take a dependency on PowerLine or copy the relevant bits here.

GitPrompt.ps1
```diff
@@ -262,7 +273,10 @@ if ($Host.UI.RawUI.BackgroundColor -eq [ConsoleColor]::DarkMagenta) {
     $s.WorkingForegroundColor               = $s.WorkingForegroundBrightColor 
 }
 
-function Global:Write-VcsStatus { $Global:VcsPromptStatuses | foreach { & $_ } }
+function Global:Write-VcsStatus {
+  Set-ConsoleMode -ANSI
```
dahlbyk: It works! :tada:

Jaykul:
We could definitely consider collaborating on a "escape sequences to write-host" translation stuff as a small external LegacyHost module, and PowerLine and Posh-Git could both take dependencies on that, if you like?

By the way, I forgot to mention this: I don't _just_ test for SupportsVirtualTerminal, I also explicitly support ConEmu -- because currently, their ANSI support is far more thorough than Windows', and it works way back on old versions of Windows 😁 

```posh
$Host.UI.SupportsVirtualTerminal -or ($Env:ConEmuANSI -eq "ON")
```

dahlbyk:
> We could definitely consider collaborating on a "escape sequences to write-host" translation stuff as a small external LegacyHost module, and PowerLine and Posh-Git could both take dependencies on that, if you like?

:+1:

> By the way, I forgot to mention this: I don't just test for SupportsVirtualTerminal, I also explicitly support ConEmu -- because currently, their ANSI support is far more thorough than Windows', and it works way back on old versions of Windows :grin:
> 
> $Host.UI.SupportsVirtualTerminal -or ($Env:ConEmuANSI -eq "ON")

Good tip. I've moved the check to the initialization of `$GitPromptSettings.AnsiConsole`, which should allow users of other ANSI-capable consoles to make the switch.

Jaykul:
Oh, I like that, I'll add a $UseAnsi switch parameter in my set-powerlineprompt instead of testing myself.

rkeithhill:
@dahlbyk 
> What would be the preferred way to guard against trying to compile or use that type on non-Windows?

This is what I use for testing for `isAdminProcess`:
```powershell
if (($PSVersionTable.PSVersion.Major -le 5) -or $IsWindows) {
```
Cross-platform PowerShell starts at v6.  At v6 and after, you can test for the Windows platform via the built-in variable `$IsWindows`.

dahlbyk:
> This is what I use for testing for `isAdminProcess`:
> ```powershell
> if (($PSVersionTable.PSVersion.Major -le 5) -or $IsWindows) {
> ```
> Cross-platform PowerShell starts at v6.  At v6 and after, you can test for the Windows platform via the built-in variable `$IsWindows`.

Would it be unreasonable to just do something like this early on, to use throughout?
```powershell
$IsWindows = $IsWindows -or ($PSVersionTable.PSVersion.Major -le 5)
```

rkeithhill:
`$IsWindows` is readonly on PowerShell v6 so let's come up with a slightly different name.  Also do the `$PSVersionTable.PSVersion.Major -le 5` first because if it is $true it will short-circuit the second test `$IsWindows`.  This variable doesn't exist on v5 or lower.

Jaykul:
You should be able to just test: `if($global:IsWindows -ne $False) { "Yep, it's Windows" }` unless you do that strict-mode thing

dahlbyk:
> unless you do that strict-mode thing

Speaking of which, we've had a few PRs come in about fixing strict mode errors but I don't really have an opinion. Does the PS community have a convention on modules' handling of strict mode?

dahlbyk:
Reopening targeting `develop` for 1.0.

Jaykul:
> we've had a few PRs come in about fixing strict mode errors but I don't really have an opinion. Does the PS community have a convention on modules' handling of strict mode?

Let me just say: 

1. Shell languages are dynamically scoped. All of them.
2. Nobody should set `StrictMode` in the global scope.

I can't speak for the community, but I don't see much value in strict mode -- because it largely forces me to give up dynamic scoping. I mean, it's great that it can catch the fact that `$SomeVariable` isn't set before it's read -- maybe that _was_ a mistake.  But in this case (and many others) that's the whole point.

### Who thinks this is more maintainable?:

```posh
if((Get-Variable IsWindows -Scope Global -ValueOnly -ErrorAction Ignore) -ne $False) { "Yep, it's Windows" } 
```

rkeithhill:
From some (very limited) testing I did it seems that `Set-StrictMode` in global scope doesn't impact script in a module.  RE someVariable isn't set - PSSA helps with that to some extent (when it isn't generating false positives).

dahlbyk:
> Who thinks this is more maintainable?

:trollface: 

I tend to agree - if we're going to use a dynamic language, we might as well take advantage of it being dynamic.

Jaykul:
> From some (very limited) testing I did it seems that Set-StrictMode in global scope doesn't impact script in a module. RE someVariable isn't set - PSSA helps with that to some extent (when it isn't generating false positives).

I believe @rkeithhill is correct, although I don't quite know why -- StrictMode is supposed to affect child scopes, and modules are (technically?) in the global scope. But I'll take it 😉 

Anyway, bottom line is: don't worry about it.

PR #307, opened 2016-10-08
Removed error thrown by symbolic-ref and describe

JanJoris:
git describe can be replaced by tag --points-at <commitish>. Other than symbolic-ref, describe does not have a --quiet switch, thus always adds to $error in case of a detached head with no tag pointing to that commit. Using tag --point-at HEAD will return a tag if one is found but not throw an error when there is none.

It is a bug which was reported through a user of [oh-my-posh](https://github.com/JanJoris/oh-my-posh/issues/27).


dahlbyk:
:100: thanks!

PR #312, opened 2016-11-22
Fix error when testing for admin on PSCore Linux.

rkeithhill:
This is somewhat experimental but the short-circuiting *if* statement should cause all existing users of Windows PowerShell (not core) to go through the old code path.  Only PS Core users on LInux/OSX will go through the "id -u" path.  I have tested this on Ubuntu but I don't have a Mac so can't test there.

Sorry about the whitespace changes.  I have VSCode configured to trim trailing whitespace.

### DIFF
--- a/GitPrompt.ps1
+++ b/GitPrompt.ps1
@@ -95,8 +95,16 @@ $global:GitPromptSettings = New-Object PSObject -Property @{
     TruncatedBranchSuffix                       = '...'
 }
 
-$currentUser = [Security.Principal.WindowsPrincipal]([Security.Principal.WindowsIdentity]::GetCurrent())
-$isAdminProcess = $currentUser.IsInRole([Security.Principal.WindowsBuiltInRole]::Administrator)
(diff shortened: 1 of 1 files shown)


dahlbyk:
> Sorry about the whitespace changes. I have VSCode configured to trim trailing whitespace.

No worries. Feel free to use `add -p` to put whitespace-only changes in separate commits.

rkeithhill:
@dahlbyk Thanks. Is that `git add -p`?  I haven't used that before, cool.  BTW I'm adding a VSCode settings file for the repo so anybody using VSCode will have trailing whitespace trimmed in the files they edit.

dahlbyk:
> Is that `git add -p`? I haven't used that before, cool.

Yep, many commands (`add`, `reset`, `checkout`, `stash` at least) accept `--patch`/`-p` to let you stage/unstage/discard/stash individual lines (or you can use the `e` option to apply a change that never existed on disk). I use it all the time.

> BTW I'm adding a VSCode settings file for the repo so anybody using VSCode will have trailing whitespace trimmed in the files they edit.

:point_right: #330

PR #319, opened 2016-12-27
Speed up Get-GitStatus

lzybkr:
These changes shouldn't change the core logic of Get-GitStatus,
they are simply avoiding some performance problems in PowerShell.

Briefly, the changes include:

* Avoid array addition by using List[string]
* Use the looping capability of the switch statement
* Use break/continue to avoid testing switch conditions
* Use a custom implementation of Select -Unique
* Avoid calling dbg if it would do nothing

I tested these changes in a repo 5000 files staged:
Before: 3.33s
After:  0.15s

### DIFF
--- a/GitUtils.ps1
+++ b/GitUtils.ps1
@@ -94,6 +94,18 @@ function Get-GitBranch($gitDir = $(Get-GitDirectory), [Diagnostics.Stopwatch]$sw
     }
 }
 
+function GetUniquePaths([System.Collections.Generic.IEnumerable[string][]] $pathCollections) {
+    $hash = New-Object System.Collections.Specialized.OrderedDictionary
+    foreach ($pathCollection in $pathCollections)
+    {
+        foreach ($path in $pathCollection)
+        {
+            $hash[$path] = 1
+        }
+    }
+    $hash.Keys
+}
+
 function Get-GitStatus($gitDir = (Get-GitDirectory)) {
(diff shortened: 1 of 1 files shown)


GitUtils.ps1
```diff
@@ -129,56 +141,69 @@ function Get-GitStatus($gitDir = (Get-GitDirectory)) {
...
-        $workingPaths = $filesAdded + $filesModified + $filesDeleted + $filesUnmerged
-        $index = New-Object PSObject @(,@($indexPaths | ?{ $_ } | Select -Unique)) |
-            Add-Member -PassThru NoteProperty Added    $indexAdded |
-            Add-Member -PassThru NoteProperty Modified $indexModified |
-            Add-Member -PassThru NoteProperty Deleted  $indexDeleted |
-            Add-Member -PassThru NoteProperty Unmerged $indexUnmerged
-        $working = New-Object PSObject @(,@($workingPaths | ?{ $_ } | Select -Unique)) |
+        #
+        # This collection is used twice, so create the array just once
+        $filesAdded = $filesAdded.ToArray()
```
dahlbyk: Delayed review... I'm curious why `$filesAdded` is special here? I don't see anything about its usage below that would be problematic.

lzybkr: I didn't want to change anything, so I create a single array for the two places this is used. All the others are used just once, so inline was cleaner.

PR #325, opened 2016-12-29
Simplify/fix join-path.

### LINKED ISSUE #268: Get-SshPath returning empty path for id_rsa

rkeithhill:
Fix #268 and add some initial Pester tests.

### DIFF
--- a/test/Ssh.Tests.ps1
+++ b/test/Ssh.Tests.ps1
@@ -1,0 +1,12 @@
+. $PSScriptRoot\Shared.ps1
+
+Describe 'SSH Function Tests' {
+    Context 'Get-SshPath Tests' {
+        It 'Returns the correct default path' {
+            Get-SshPath | Should BeExactly $Home\.ssh\id_rsa
+        }
+        It 'Returns the correct path given a filename' {
+            Get-SshPath mykey | Should BeExactly $Home\.ssh\mykey
+        }
+    }
+}
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -1,0 +1,66 @@
+// Available variables which can be used inside of strings.
+// ${workspaceRoot}: the root folder of the team
+// ${file}: the current opened file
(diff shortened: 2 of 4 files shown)


test/Ssh.Tests.ps1
```diff
@@ -0,0 +1,12 @@
+. $PSScriptRoot\Shared.ps1
+
+Describe 'SSH Function Tests' {
+    Context 'Get-SshPath Tests' {
+        It 'Returns the correct default path' {
+            Get-SshPath | Should BeExactly $Home\.ssh\id_rsa
+        }
+        It 'Returns the correct path given a filename' {
+            Get-SshPath mykey | Should BeExactly $Home\.ssh\mykey
```
dahlbyk: Note that these path expectations will likely fail on not-Windows.

rkeithhill: It shouldn't but I will test it.  PowerShell itself handles both `\` and `/` as path sep chars.  Issues arise though when a path is passed down to a native application.

rkeithhill: OK so PowerShell internally handles both just fine but when you "compare exactly", the expected has to be correct.  Submitting a PR to fix this.

PR #326, opened 2016-12-29
Upstream Gone

dahlbyk:
Just stumbled on this edge case. To reproduce (assuming `origin/gone` doesn't exist):

```
> git branch gone
> git config branch.gone.remote origin
> git config branch.gone.merge refs/heads/gone
> git checkout gone
Switched to branch 'gone'
Your branch is based on 'origin/gone', but the upstream is gone.
  (use "git branch --unset-upstream" to fixup)
```

I had originally thought to use ✘ for the symbol, but font support seems limited. I settled on either × or ∙ . Thoughts?

powershell_2016-12-29_10-26-38

----------

Also, semicolons are :cool:

### DIFF
--- a/GitPrompt.ps1
+++ b/GitPrompt.ps1
@@ -39,6 +39,10 @@ $global:GitPromptSettings = New-Object PSObject -Property @{
     BranchForegroundColor                       = [ConsoleColor]::Cyan
     BranchBackgroundColor                       = $Host.UI.RawUI.BackgroundColor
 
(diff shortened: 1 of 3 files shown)


rkeithhill:
IMO some form of `x` would be more intuitive than ` ∙`.

dahlbyk:
> IMO some form of `x` would be more intuitive than `∙`.

Sticking with `x`, then.

PR #329, opened 2016-12-29
Avoid unsupported type constraint

dahlbyk:
Fixes PowerShell 2.0 regression from #319.

cc @lzybkr


lzybkr:
LGTM

rkeithhill:
So I don't get a crash on v2 (unsupported type constraint) but the numbers are missing:
image
Same for files in the index if I add the file to the index.  I'll look into this.  Hopefully it is something easy to fix.

rkeithhill:
OK, on v2 it looks like the `Index` and `Working` properties are coming through as type string - value `-NoEnumerate`.  I'll see if I can fix it.

PR #331, opened 2016-12-30
Fix Get-SshPath Pester tests on Linux.

rkeithhill:
Tested on Ubuntu 16.04.

### DIFF
--- a/test/Ssh.Tests.ps1
+++ b/test/Ssh.Tests.ps1
@@ -2,11 +2,14 @@
 
 Describe 'SSH Function Tests' {
     Context 'Get-SshPath Tests' {
+        BeforeAll {
+            $sepChar = [System.IO.Path]::DirectorySeparatorChar
+        }
         It 'Returns the correct default path' {
-            Get-SshPath | Should BeExactly $Home\.ssh\id_rsa
+            Get-SshPath | Should BeExactly "${Home}${sepChar}.ssh${sepChar}id_rsa"
         }
         It 'Returns the correct path given a filename' {
-            Get-SshPath mykey | Should BeExactly $Home\.ssh\mykey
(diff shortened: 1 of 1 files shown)


rkeithhill:
Tested on Linux:
image

test/Ssh.Tests.ps1
```diff
@@ -2,11 +2,14 @@
...
+        BeforeAll {
+            $sepChar = [System.IO.Path]::DirectorySeparatorChar
+        }
         It 'Returns the correct default path' {
-            Get-SshPath | Should BeExactly $Home\.ssh\id_rsa
+            Get-SshPath | Should BeExactly "${Home}${sepChar}.ssh${sepChar}id_rsa"
         }
         It 'Returns the correct path given a filename' {
-            Get-SshPath mykey | Should BeExactly $Home\.ssh\mykey
+            Get-SshPath mykey | Should BeExactly "${Home}${sepChar}.ssh${sepChar}mykey"
```
dahlbyk: This works, but is kind of gross. I wonder if Pester could learn `Should BePath` that would translate `/` in the expected value to `[System.IO.Path]::DirectorySeparatorChar`?

rkeithhill: That would be a good suggestion for the [Pester project](https://github.com/pester/Pester).  In the meantime we could create a function like `MakeNativePath` or some such name that would replace `\` with '/' e.g.:
```powershell
function MakeNativePath([string]$Path) { 
    $Path -replace '\\|/', [System.IO.Path]::DirectorySeparatorChar 
}
Get-SshPath mykey | Should BeExactly (MakeNativePath "${Home}\.ssh\mykey") 
```
What do you think - as in interim?

PR #333, opened 2016-12-30
Fix PR 319 to work on PS v2/v3.

rkeithhill:
Verified that v4 has the -NoEnumerate parameter on Write-Output.  Azure VMs come in handy.  BTW I removed the Select -uniq from the old version of the code since the GetUniquePaths function handles this.

cc @lzybkr

### DIFF
--- a/GitUtils.ps1
+++ b/GitUtils.ps1
@@ -207,17 +207,33 @@ function Get-GitStatus($gitDir = (Get-GitDirectory)) {
 
         $indexPaths = @(GetUniquePaths $indexAdded,$indexModified,$indexDeleted,$indexUnmerged)
         $workingPaths = @(GetUniquePaths $filesAdded,$filesModified,$filesDeleted,$filesUnmerged)
-        $index = Write-Output -NoEnumerate $indexPaths |
-            Add-Member -PassThru NoteProperty Added    $indexAdded.ToArray() |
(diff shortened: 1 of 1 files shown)


GitUtils.ps1
```diff
@@ -207,17 +207,33 @@ function Get-GitStatus($gitDir = (Get-GitDirectory)) {
...
+                Add-Member -PassThru NoteProperty Deleted  $indexDeleted |
+                Add-Member -PassThru NoteProperty Unmerged $indexUnmerged
+            $working = New-Object PSObject @(,@($workingPaths | Where-Object { $_ })) |
+                Add-Member -PassThru NoteProperty Added    $filesAdded |
+                Add-Member -PassThru NoteProperty Modified $filesModified |
+                Add-Member -PassThru NoteProperty Deleted  $filesDeleted |
+                Add-Member -PassThru NoteProperty Unmerged $filesUnmerged
+        }
+        else {
+            $index = Write-Output -NoEnumerate $indexPaths |
```
lzybkr: `Write-Output -NoEnumerate $indexPaths` is basically the same as `,$indexPaths` which should work on all versions, I've just been using the former because I think it's clearer.

dahlbyk: Reverted in https://github.com/dahlbyk/posh-git/pull/337

PR #337, opened 2016-12-30
Semi-revert "Fix PR 319 to work on PS v2/v3."

dahlbyk:
Follow-up from https://github.com/dahlbyk/posh-git/pull/333/files#r94205904. Is this what you meant, @lzybkr? Seems to work as expected.


rkeithhill:
OK yeah, that's better. :-)

PR #343, opened 2016-12-31
Support `git push <remote> +<tab>`

### LINKED ISSUE #173: Feat: Support additional short-hand modifiers in branch auto-complete

dahlbyk:
Finishes #174 (by @dan-turner) to close #173. Turns out it's a bit more complicated than I had thought due to tab expansion needing to return the full "word" being expanded (including the `+`). I think this will do it, though?


dan-turner:
Better late than never :)

dahlbyk:
> Better late than never :)

Only two years! Sigh.

Belated thanks for the bug report and start to a fix!

Issue #328, opened 2016-12-29
Roadmap to 1.0


dahlbyk:
First off, huge thanks to @rkeithhill for his recent help in pushing posh-git forward! He suggested adopting SemVer and using 1.0 for the [forewarned](https://github.com/dahlbyk/posh-git/pull/186) dropping of support for PowerShell 2.0. Before we do that, I think it makes sense to cut a "final" (modulo minor bug fixes) pre-1.0 release for those still using PowerShell 2.0.

The 1.0 milestone also give us an opportunity to revisit the many ways that posh-git is currently distributed:

* `git clone <url> && .\install.ps1`
* Download ZIP & `.\install.ps1`
* [Chocolatey](https://chocolatey.org/packages/poshgit) (which always installs or upgrades to `master`)
* [PSGet](http://psget.net/directory/posh-git/)
* [PowerShell Gallery](https://www.powershellgallery.com/packages/posh-git)

So... tentative roadmap:

## posh-git 0.7

* [ ] Fix Chocolatey package to install specific version.
* [ ] Close out existing PRs that are compatible with 2.0
  * [ ] In particular, finally merge https://github.com/dahlbyk/posh-git/pull/208, assuming it works with PowerShell 2.0 /cc @cmarcusreid
* [ ] Close out trivial issues that I've just been ignoring for months or years 😒 

## posh-git 1.0

* [ ] Drop 2.0 support and warnings
* [ ] Review exported commands - surface area seems too big (e.g. [`Invoke-NullCoalescing`](https://github.com/dahlbyk/posh-git/issues/93))
* [ ] PowerShell Core support
* [ ] [ANSI support?](https://github.com/dahlbyk/posh-git/issues/304)

dahlbyk:
cc @ferventcoder and @mwrock for Chocolatey input. Do you know of any other Chocolatey packages for PowerShell modules that are now available from the gallery? Would it make sense to try to restructure the package to just make sure PowerShell Gallery and the posh-git module are installed?

dahlbyk:
cc @shiftkey for GitHub Desktop implications. Do you have any telemetry for PowerShell versions used in the wild?

ferventcoder:
@dahlbyk I would consider PS-Get separate and a release to the PowerShell Gallery means the module goes there. 

If you also decide to continue releasing to Chocolatey, I would treat it differently and not just have it use PS-Get. Some folks may want to install without any additional dependencies.

[...]

PR #349, opened 2017-01-03
One attempt at providing a prompt function during module import.

### LINKED ISSUE #217: Move simple prompt to the module.
I frequently use PowerShell for many tasks and have many modules that I do not load automatically in my profile. When loading posh-git I have to manually execute the simple prompt code. Moving the simple prompt to the module make the it more user friendly.

rkeithhill:
Fix #217.  This would only export the prompt function if the prompt function is still set to the default.

I'm not suggesting we accept this just yet, just wanted to put something up folks could play with.

### DIFF
--- a/GitPrompt.ps1
+++ b/GitPrompt.ps1
@@ -306,6 +306,30 @@ $PoshGitVcsPrompt = {
     Write-GitStatus $GitStatus
 }
 
+# A simple prompt that displays Git status summary info when inside of a Git repo.
+function prompt {
+    $origLastExitCode = $LASTEXITCODE
+
+    # A UNC path has no drive so it's better to use the ProviderPath e.g. "\\server\share".
+    # However for any path with a drive defined, it's better to use the Path property.
+    # In this case, ProviderPath is "\LocalMachine\My"" whereas Path is "Cert:\LocalMachine\My".
+    # The latter is more desirable.
+    $pathInfo = $ExecutionContext.SessionState.Path.CurrentLocation
+    $curPath = if ($pathInfo.Drive) { $pathInfo.Path } else { $pathInfo.ProviderPath }
+    if ($curPath -and $curPath.ToLower().StartsWith($Home.ToLower()))
+    {
+        $curPath = "~" + $curPath.SubString($Home.Length)
+    }
+
+    Write-Host $curPath -NoNewline
+
+    # Write the Git status summary information to the host.
+    Write-VcsStatus
+
+    $global:LASTEXITCODE = $origLastExitCode
+    "> "
+}
+
 # Install handler for removal/unload of the module
 $Global:VcsPromptStatuses += $PoshGitVcsPrompt
 $ExecutionContext.SessionState.Module.OnRemove = {
--- a/GitPrompt.ps1
+++ b/GitPrompt.ps1
@@ -96,6 +96,9 @@ $global:GitPromptSettings = New-Object PSObject -Property @{
 
     EnableWindowTitle                           = 'posh~git ~ '
 
+    PromptSuffix                                = '> '
+    PromptDebugSuffix                           = ' [DBG]>> '
+
     Debug                                       = $false
 
     BranchNameLimit                             = 0
--- a/GitPrompt.ps1
+++ b/GitPrompt.ps1
@@ -306,32 +306,4 @@ $PoshGitVcsPrompt = {
     Write-GitStatus $GitStatus
 }
 
-# A simple prompt that displays Git status summary info when inside of a Git repo.
-function prompt {
-    $origLastExitCode = $LASTEXITCODE
-
-    # A UNC path has no drive so it's better to use the ProviderPath e.g. "\\server\share".
-    # However for any path with a drive defined, it's better to use the Path property.
-    # In this case, ProviderPath is "\LocalMachine\My"" whereas Path is "Cert:\LocalMachine\My".
-    # The latter is more desirable.
-    $pathInfo = $ExecutionContext.SessionState.Path.CurrentLocation
-    $curPath = if ($pathInfo.Drive) { $pathInfo.Path } else { $pathInfo.ProviderPath }
-    if ($curPath -and $curPath.ToLower().StartsWith($Home.ToLower()))
-    {
-        $curPath = "~" + $curPath.SubString($Home.Length)
-    }
-
-    Write-Host $curPath -NoNewline
-
-    # Write the Git status summary information to the host.
-    Write-VcsStatus
-
-    $global:LASTEXITCODE = $origLastExitCode
-    "> "
-}
-
-# Install handler for removal/unload of the module
 $Global:VcsPromptStatuses += $PoshGitVcsPrompt
-$ExecutionContext.SessionState.Module.OnRemove = {
-    $Global:VcsPromptStatuses = $Global:VcsPromptStatuses | Where-Object { $_ -ne $PoshGitVcsPrompt }
-}
--- a/posh-git.psm1
+++ b/posh-git.psm1
@@ -1,4 +1,4 @@
-param([switch]$NoVersionWarn = $false)
+param([switch]$NoVersionWarn)
 
 if (Get-Module posh-git) { return }
 
@@ -11,15 +11,16 @@ if ($psv.Major -lt 3 -and !$NoVersionWarn) {
     "To suppress this warning, change your profile to include 'Import-Module posh-git -Args `$true'.")
 }
 
-Push-Location $psScriptRoot
-.\CheckVersion.ps1 > $null
+& $PSScriptRoot\CheckVersion.ps1 > $null
 
-. .\Utils.ps1
-. .\GitUtils.ps1
-. .\GitPrompt.ps1
-. .\GitTabExpansion.ps1
-. .\TortoiseGit.ps1
-Pop-Location
+# Get the current prompt function *before* we define a prompt function by dot sourcing GitPrompt.ps1.
+$currentPromptDef = if ($funcInfo = Get-Command prompt -ErrorAction SilentlyContinue) { $funcInfo.Definition }
+
+. $PSScriptRoot\Utils.ps1
+. $PSScriptRoot\GitUtils.ps1
+. $PSScriptRoot\GitPrompt.ps1
+. $PSScriptRoot\GitTabExpansion.ps1
+. $PSScriptRoot\TortoiseGit.ps1
 
 if (!$Env:HOME) { $Env:HOME = "$Env:HOMEDRIVE$Env:HOMEPATH" }
 if (!$Env:HOME) { $Env:HOME = "$Env:USERPROFILE" }
@@ -27,10 +28,9 @@ if (!$Env:HOME) { $Env:HOME = "$Env:USERPROFILE" }
 Get-TempEnv 'SSH_AGENT_PID'
 Get-TempEnv 'SSH_AUTH_SOCK'
 
-Export-ModuleMember `
-    -Alias @(
-        '??') `
-    -Function @(
+$exportModuleParams = @{
+    Alias = @('??') # TODO: Remove in 1.0.0
+    Function = @(
         'Invoke-NullCoalescing',
         'Write-GitStatus',
         'Write-Prompt',
@@ -46,6 +46,21 @@ Export-ModuleMember `
         'Add-SshKey',
         'Get-SshPath',
         'Update-AllBranches',
-        'tgit')
+        'tgit'
+    )
+}
 
+# Get the default prompt definition.
+if ($psv.Major -eq 2) {
+    $defaultPromptDef = "`$(if (test-path variable:/PSDebugContext) { '[DBG]: ' } else { '' }) + 'PS ' + `$(Get-Location) + `$(if (`$nestedpromptlevel -ge 1) { '>>' }) + '> '"
+}
+else {
+    $defaultPromptDef = [Runspace]::DefaultRunspace.InitialSessionState.Commands['prompt'].Definition
+}
 
+# If there is no prompt function or the prompt function is the default, export the posh-git prompt function.
+if (!$currentPromptDef -or ($currentPromptDef -eq $defaultPromptDef)) {
+    $exportModuleParams.Function += 'prompt'
+}
+
+Export-ModuleMember @exportModuleParams
--- a/posh-git.psm1
+++ b/posh-git.psm1
@@ -13,9 +13,6 @@ if ($psv.Major -lt 3 -and !$NoVersionWarn) {
 
 & $PSScriptRoot\CheckVersion.ps1 > $null
 
-# Get the current prompt function *before* we define a prompt function by dot sourcing GitPrompt.ps1.
-$currentPromptDef = if ($funcInfo = Get-Command prompt -ErrorAction SilentlyContinue) { $funcInfo.Definition }
-
 . $PSScriptRoot\Utils.ps1
 . $PSScriptRoot\GitUtils.ps1
 . $PSScriptRoot\GitPrompt.ps1
@@ -38,6 +35,7 @@ else {
 
 # If there is no prompt function or the prompt function is the default, export the posh-git prompt function.
 $promptReplaced = $false
+$currentPromptDef = if ($funcInfo = Get-Command prompt -ErrorAction SilentlyContinue) { $funcInfo.Definition }
 if (!$currentPromptDef -or ($currentPromptDef -eq $defaultPromptDef)) {
     Set-Item Function:\prompt -Value {
         $origLastExitCode = $LASTEXITCODE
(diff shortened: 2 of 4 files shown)


posh-git.psm1
```diff
@@ -11,26 +11,79 @@ if ($psv.Major -lt 3 -and !$NoVersionWarn) {
     "To suppress this warning, change your profile to include 'Import-Module posh-git -Args `$true'.")
 }
 
-Push-Location $psScriptRoot
```
dahlbyk: Are we just getting lucky that nobody using PS v2 has complained about this breaking anything?

GitPrompt.ps1
```diff
@@ -306,8 +306,4 @@ $PoshGitVcsPrompt = {
     Write-GitStatus $GitStatus
 }
 
-# Install handler for removal/unload of the module
 $Global:VcsPromptStatuses += $PoshGitVcsPrompt
-$ExecutionContext.SessionState.Module.OnRemove = {
-    $Global:VcsPromptStatuses = $Global:VcsPromptStatuses | Where-Object { $_ -ne $PoshGitVcsPrompt }
```
dahlbyk: Why remove?

posh-git.psm1
rkeithhill: There was a pop-location later but if something crashed out before the pop-location, it would have left the current location to the module's root dir.

GitPrompt.ps1
rkeithhill: BTW I need to update that logic a bit to check if someone might have installed a different prompt function after we installed the posh-git default one.

posh-git.psm1
rkeithhill: Oh yeah, $PSScriptRoot works only inside module (PSM1) files for PS v2 but not in PS1 files IIRC. That was fixed in v3 I believe.

GitPrompt.ps1
rkeithhill: And as a "module level" operation, posh-git.psm1 seemed like a more fitting place for this handler.

posh-git.psm1
dahlbyk: Ah, that makes more sense 👍

GitPrompt.ps1
```diff
@@ -306,8 +306,4 @@ $PoshGitVcsPrompt = {
     Write-GitStatus $GitStatus
 }
 
-# Install handler for removal/unload of the module
 $Global:VcsPromptStatuses += $PoshGitVcsPrompt
```
dahlbyk: > And as a "module level" operation, posh-git.psm1 seemed like a more fitting place for this handler.

I wonder if this should move alongside the removal operation?

rkeithhill:
I've done some testing on PS v2 and PS v5 both with modified prompt functions and without. It is looking pretty good and I'm feeling better about posh-git installing this when the user hasn't modified their prompt function.

BTW when the user has their own prompt function that uses posh-git commands, we don't remove it. This has the consequence of causing posh-git to be immediately reloaded when their prompt is evaluated after `remove-module posh-git`.  There are several workarounds including configuring a dumb prompt to allow posh-git to be removed e.g.:
```powershell
function prompt { "$(pwd)> " }
```
or for v3 or higher
```powershell
Set-Item Function:\prompt ([Runspace]::DefaultRunspace.InitialSessionState.Commands['prompt'].Definition)
```
They can also close all PowerShell sessions and start powershell with the `-noprofile` parameter.  I believe this may be important in certain install situations (perhaps with Chocolatey).  For PowerShellGet, each different version of posh-git will be installed into its own dir.  So there shouldn't be any conflicts with updating.

posh-git.psm1
```diff
@@ -11,26 +11,79 @@ if ($psv.Major -lt 3 -and !$NoVersionWarn) {
...
+    Set-Item Function:\prompt -Value {
+        $origLastExitCode = $LASTEXITCODE
+
+        # A UNC path has no drive so it's better to use the ProviderPath e.g. "\\server\share".
+        # However for any path with a drive defined, it's better to use the Path property.
+        # In this case, ProviderPath is "\LocalMachine\My"" whereas Path is "Cert:\LocalMachine\My".
+        # The latter is more desirable.
+        $pathInfo = $ExecutionContext.SessionState.Path.CurrentLocation
+        $curPath = if ($pathInfo.Drive) { $pathInfo.Path } else { $pathInfo.ProviderPath }
+        if ($curPath -and $curPath.ToLower().StartsWith($Home.ToLower()))
```
dahlbyk: I'm a fan of the `$Home` check, nice addition!

1. Favor the `StringComparison` overload over comparing `ToLower`ed strings
2. According to [MSDN](https://msdn.microsoft.com/en-us/library/dd465121.aspx):
   > The string behavior of the file system, registry keys and values, and environment variables is best represented by `StringComparison.OrdinalIgnoreCase`.

Therefore: `$curPath.StartsWith($Home, [StringComparison]::OrdinalIgnoreCase)`

Commence countdown to case-insensitivity bug on Linux: 5...4...3...

posh-git.psm1
```diff
@@ -11,26 +11,79 @@ if ($psv.Major -lt 3 -and !$NoVersionWarn) {
...
+else {
+    $defaultPromptDef = [Runspace]::DefaultRunspace.InitialSessionState.Commands['prompt'].Definition
+}
+
+# If there is no prompt function or the prompt function is the default, export the posh-git prompt function.
+$promptReplaced = $false
+$currentPromptDef = if ($funcInfo = Get-Command prompt -ErrorAction SilentlyContinue) { $funcInfo.Definition }
+if (!$currentPromptDef -or ($currentPromptDef -eq $defaultPromptDef)) {
+    Set-Item Function:\prompt -Value {
+        $origLastExitCode = $LASTEXITCODE
```
dahlbyk: PS amateur question: any reason to use `$global:LASTEXITCODE` here, as below?

posh-git.psm1
```diff
@@ -11,26 +11,79 @@ if ($psv.Major -lt 3 -and !$NoVersionWarn) {
...
+
+        # Write the Git status summary information.
+        Write-VcsStatus
+
+        # When in debug mode, let user know
+        if ((Test-Path Variable:\PSDebugContext) -or [runspace]::DefaultRunspace.Debugger.InBreakpoint) {
+            $promptSuffix = "`n[DBG]: PS>> "
+        }
+        else {
+            $promptSuffix = "`nPS> "
```
dahlbyk: I'd prefer not to change the default posh-git prompt since it's what non-customizing users (like me 😀) are now used to. But we could add `PromptPrefix` and `PromptSuffix` (and `PromptDebug*ix`?) to `$GitPromptSettings` to make this particular arrangement easier to achieve without modifying `prompt`.

posh-git.psm1
```diff
@@ -11,26 +11,79 @@ if ($psv.Major -lt 3 -and !$NoVersionWarn) {
...
+if ($psv.Major -eq 2) {
+    $defaultPromptDef = "`$(if (test-path variable:/PSDebugContext) { '[DBG]: ' } else { '' }) + 'PS ' + `$(Get-Location) + `$(if (`$nestedpromptlevel -ge 1) { '>>' }) + '> '"
+}
+else {
+    $defaultPromptDef = [Runspace]::DefaultRunspace.InitialSessionState.Commands['prompt'].Definition
+}
+
+# If there is no prompt function or the prompt function is the default, export the posh-git prompt function.
+$promptReplaced = $false
+$currentPromptDef = if ($funcInfo = Get-Command prompt -ErrorAction SilentlyContinue) { $funcInfo.Definition }
```
dahlbyk: Is there a reason to prefer your construction over this one?

```powershell
$currentPromptDef = Get-Command prompt -ErrorAction SilentlyContinue) | Select-Object -ExpandProperty Definition
```

posh-git.psm1
```diff
@@ -11,26 +11,79 @@ if ($psv.Major -lt 3 -and !$NoVersionWarn) {
...
+        }
+        else {
+            $promptSuffix = "`nPS> "
+        }
+
+        $global:LASTEXITCODE = $origLastExitCode
+        $promptSuffix
+    }
+
+    $promptReplaced = $true
```
dahlbyk: Could we attach a property to our `prompt` instead of using a separate flag? I don't suppose there's any way to set its `Source = "posh-git"` like all the other functions we export?

dahlbyk:
> BTW when the user has their own prompt function that uses posh-git commands, we don't remove it. This has the consequence of causing posh-git to be immediately reloaded when their prompt is evaluated after `remove-module posh-git`.

Interesting... So my first thought is to have `OnRemove` check if the `prompt` it's about to restore includes `Write-VcsStatus`, and if so create a global `Write-VcsStatus` no-op on its way out (perhaps with some sort of warning that it did so). Or always create `Write-VcsStatus` just in case. Or let the reinstall loop happen - how _dare_ they uninstall posh-git, anyway?

Then again, technically `Write-VcsStatus` is meant to be the shim through which posh-hg and friends are also exposed, so that function _really_ shouldn't belong to posh-git at all. So we could make posh-git depend on a `posh-vcs` module (maintained here or moved to [posh-vcs](https://github.com/dahlbyk/posh-vcs)) which provides `Write-VcsStatus`, `Un`/`Register-VcsStatusPrompt`, and the default `prompt` infrastructure. Removing posh-git would then simply `Unregister-VcsStatusPrompt`, effectively punting the reinstall cycle to posh-vcs. I have no idea if there remains any demand for VCS support in PowerShell other than Git. (https://github.com/dahlbyk/posh-tf never took off :trollface:)

Or we could hack it (`Write-VcsStatus` no-op or equivalent) until 1.0 at which point we retire the `Write-VcsStatus` cruft in favor of a posh-git `prompt` that calls `Write-GitStatus`.

posh-git.psm1
rkeithhill: We can check for Linux and use Ordinal.

posh-git.psm1
rkeithhill: OK by me.  I was just floating this is an idea.  If we added a setting, what would it be called?  `CommandEntryOnNewline`?

rkeithhill:
I hadn't quite grokked what the Vcs* functions were for yet.  Thanks for the info.

> Or we could hack it (Write-VcsStatus no-op or equivalent) until 1.0

But in my prompt function I use `Write-GitStatus (Get-GitStatus)` so this could get a bit messy.  I don't think uninstall would be all that common.  In fact, `remove-module posh-git` is likely to be rare as well.  And this would only effect folks who have a customized prompt function.  So I guess I'm not sure we need to worry about this for 0.7.0 other than to be aware that it could crop up from time-to-time.  Plus there are simple workarounds.

GitPrompt.ps1
rkeithhill: If we do that, should we also move the decl/assignment of `$PoshGitVcsPrompt` just above this line as well?

dahlbyk:
So how about we just log a warning `OnRemove` if we are restoring a non-default `prompt`, regardless of its contents, e.g. "If prompt still references posh-git functions, it may cause the module to re-import.")? That should be enough info to explain a reimport without overcomplicating things.

rkeithhill:
Sounds good to me. I'm working on an update to this PR to address the StringComparison issue and a way to select a two line prompt.  Any thoughts on the name of that setting?  I'm currently using `CommandEntryOnNewline` but maybe it should be `TwoLinePrompt`?

dahlbyk:
> Any thoughts on the name of that setting? I'm currently using `CommandEntryOnNewline` but maybe it should be `TwoLinePrompt`?

I think we can get away with:

```powershell
# Write the Git status summary information.
Write-VcsStatus

$debugMode = (Test-Path Variable:\PSDebugContext) -or [runspace]::DefaultRunspace.Debugger.InBreakpoint

$global:LASTEXITCODE = $origLastExitCode
if ($debugMode) {
    Write-Host (if ($GitPromptSettings.PromptDebugSuffix) { $GitPromptSettings.PromptDebugSuffix } else { ' [DBG]>> ' })
} else {
    Write-Host (if ($GitPromptSettings.PromptSuffix) { $GitPromptSettings.PromptSuffix } else { '> ' })
}
```

Your two-line prompt config becomes:

```powershell
$GitPromptSettings.PromptDebugSuffix = '`n[DBG]: PS>> '
$GitPromptSettings.PromptSuffix = '`nPS> '
# plus Powerline config if applicable, which posh-git doesn't need to know
```

And anything more complex could be handled by a custom `prompt`.

rkeithhill:
What if we defaulted these two settings e.g.:
```powershell
$GitPromptSettings.PromptDebugSuffix = ' [DBG]: >> '
$GitPromptSettings.PromptSuffix = '> '
```
That would also help folks know how to use these two settings.

posh-git.psm1
```diff
@@ -33,83 +33,71 @@ else {
...
         # However for any path with a drive defined, it's better to use the Path property.
         # In this case, ProviderPath is "\LocalMachine\My"" whereas Path is "Cert:\LocalMachine\My".
         # The latter is more desirable.
         $pathInfo = $ExecutionContext.SessionState.Path.CurrentLocation
-        $curPath = if ($pathInfo.Drive) { $pathInfo.Path } else { $pathInfo.ProviderPath }
-        if ($curPath -and $curPath.ToLower().StartsWith($Home.ToLower()))
+        $currentPath = if ($pathInfo.Drive) { $pathInfo.Path } else { $pathInfo.ProviderPath }
+
+        # File system paths are case-sensitive on Linux and case-insensitive on Windows and macOS
+        if (($PSVersionTable.PSVersion.Major -ge 6) -and $IsLinux) {
```
dahlbyk: Micro-optimization, but we could calculate `$stringComparison` once, assuming we can get the scriptblock to close over the variable.

posh-git.psm1
```diff
@@ -33,83 +33,71 @@ else {
...
-
     $promptReplaced = $true
 }
 
 # Install handler for removal/unload of the module
 $ExecutionContext.SessionState.Module.OnRemove = {
-    $Global:VcsPromptStatuses = $Global:VcsPromptStatuses | Where-Object { $_ -ne $PoshGitVcsPrompt }
+    $global:VcsPromptStatuses = $global:VcsPromptStatuses | Where-Object { $_ -ne $PoshGitVcsPrompt }
 
     if ($promptReplaced) {
```
dahlbyk: Wondering if this flag has been made obsolete by the `$promptDef -eq $poshGitPromptScriptBlock` check? I suppose it's an optimization, but I wouldn't be opposed to reducing complexity at the expense of a marginally slower uninstall.

rkeithhill:
> Should we also drop the prompt definition from profile.example.ps1?

I think so.  Perhaps we can rename that as well? I tend to following a pattern in my profile scirpts of `<module-name>_config.ps1`.  It's a practice I've ripped off from Bash.

> but we could calculate $stringComparison once

The test takes between 2 & 4 mSecs and StringComparison is an enum (no heap impact).  It's probably not worth bothering.

> Wondering if this flag has been made obsolete

There's a small chance someone could have created a custom prompt function by copying that code.  In which case, we would remove it.  I could either way.

dahlbyk:
> I think so. Perhaps we can rename that as well?

Well anyone using `install.ps1` has a `$PROFILE` reference to `profile.example.ps1`, so we should rename/deprecate that file with prejudice. I'm inclined to defer fiddling there until after 0.7 drops when we settle on the install/import strategy for 1.0 (#328). Even then, it's probably more user-friendly for 1.0 to ship a `profile.example.ps1` that simply warns that the file is obsolete and that profiles should be updated to not reference it since it will be going away with 2.0.

> The test takes between 2 & 4 mSecs and StringComparison is an enum (no heap impact). It's probably not worth bothering.

:+1:

> > Wondering if this flag has been made obsolete
> 
> There's a small chance someone could have created a custom prompt function by copying that code. In which case, we would remove it. I could go either way.

I'm all for simplifying the code we have to maintain. If someone customizes their prompt poorly, that's on them - they can see all our source here.

rkeithhill:
> I'm inclined to defer fiddling there until after 0.7 drops when we settle on the install/import strategy for 1.0 (#328). 

👍

rkeithhill:
Also, if someone sets their `$GitPromptSettings.PromptSuffix` to something like `$null` or an empty string, PowerShell will not like that and use `PS>` instead e.g.:
```
~\GitHub\rkeithhill\posh-git [rkeithhill/auto-export-prompt ≡]> $GitPromptSettings.PromptSuffix = ''
~\GitHub\rkeithhill\posh-git [rkeithhill/auto-export-prompt ≡]PS>
```
I can put in a check for `$null` or empty string and then force the $promptSuffix to `' '`.  But should we try to save the user from themselves?

dahlbyk:
> Should we still remove the prompt function? I'm guessing yes but wanted to double check.

Yes.

> I can put in a check for `$null` or empty string and then force the `$promptSuffix` to `' '`. But should we try to save the user from themselves?

That's what motivated the extra complexity in my version - if you don't give a good override, you get the default experience. Still a good idea to set the settings properties to match the default.

```powershell
if ($debugMode) {
    Write-Host (if ($GitPromptSettings.PromptDebugSuffix) { $GitPromptSettings.PromptDebugSuffix } else { ' [DBG]>> ' })
} else {
    Write-Host (if ($GitPromptSettings.PromptSuffix) { $GitPromptSettings.PromptSuffix } else { '> ' })
}
```

dahlbyk:
I'm feeling pretty good about this! Is there any more testing you want to do on v2 or v5 (is v3 a concern)?

rkeithhill:
OK tested on v3. Works fine.  I think this is ready to be merged.

PR #322, opened 2016-12-29
Update profile example prompt function.

rkeithhill:

PR #351, opened 2017-01-03
Fix that module can be found by not imported?

### LINKED ISSUE #350: Installing from chocolatey on Powershell Version 2.0 is not making the module importable
When I run powershell 2.0 as administrator and install via chocolatey, it cannot import the module. It looks like chocolatey never actually install the module into an accessible location.

dahlbyk:
Fix #350 and https://github.com/dahlbyk/posh-git/pull/322#issuecomment-269663743, I think?

I can't explain why `Get-Module posh-git` would find a module but `Import-Module posh-git` wouldn't find the same. I'm wondering if the `Get-Module` result should be filtered through `Select-Object -First 1` and maybe sorted by version or something?

### DIFF
--- a/profile.example.ps1
+++ b/profile.example.ps1
@@ -1,6 +1,6 @@
 # Import the posh-git module, first via installed posh-git module.
 # If the module isn't installed, then attempt to load it from the cloned posh-git Git repo.
-$poshGitModule = Get-Module posh-git -ListAvailable
+$poshGitModule = Get-Module posh-git -ListAvailable | Sort-Object Version -Descending | Select-Object -First 1
 if ($poshGitModule) {
     $poshGitModule | Import-Module
 }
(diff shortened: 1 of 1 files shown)


rkeithhill:
>  I'm wondering if the Get-Module result should be filtered through Select-Object -First 1 and maybe sorted by version or something?

But if Get-Module finds anything, it should load the latest version unless one of the Version parameters are specified. In this case, they are not.

Anyway, your fix should eliminate the possibility of error but I would Sort by version -desc and Select the first module.   

We should also move the prompt function definition *before* we import the module so that the import will detect a custom prompt function is in use so it won't set a new prompt that get's clobbered later by the prompt function appearing after the import.  It's not that this won't work, it is a slight perf improvement.

dahlbyk:
> We should also move the prompt function definition _before_ we import the module so that the import will detect a custom prompt function is in use so it won't set a new prompt that get's clobbered later by the prompt function appearing after the import. It's not that this won't work, it is a slight perf improvement.

I'm inclined to drop the `prompt` from `profile.example.ps1` altogether as part of #349.

> I would Sort by version -desc and Select the first module.

Done. Good?

PR #356, opened 2017-01-04
Add issue template for new issues.

rkeithhill:
I "think" all we need is this specially named file but I'm not 100% sure of that.


dahlbyk:
We can try this... I've not had great success with my other issue template, but we'll see.

PR #358, opened 2017-01-06
Chocolatey Simplification + Uninstall

### LINKED ISSUE #239: Uninstall through Chocolatey
https://chocolatey.org/packages/poshgit#comment-2409077345

> So, I ran "cuninst poshgit" to uninstall, but it didn't remove anything. Posh-git is still running in my shell. For anyone wanting to remove it, remove it from your $profile file and delete the "c:\tools\poshgit" folder. If un-installation could be implemented in the chocolatey package, that would be great!

### LINKED ISSUE #355: Error in $profile after poshgit installed?
I'm seeing the following message when my $profile gets executed:

```
Rename-Item : Cannot rename because item at 'Function:\Prompt' does not exist.
At C:\Users\myuser\Documents\WindowsPowerShell\Microsoft.PowerShell_profile.ps1:66 char:1
+ Rename-Item Function:\Prompt PoshGitPrompt -Force
```

here's what the relevant line looks like:

```
. 'C:\tools\poshgit\dahlbyk-posh-git-464601f\profile.example.ps1'
Rename-Item Function:\Prompt PoshGitPrompt -Force
function Prompt() {if(Test-Path Function:\PrePoshGitPrompt){++$global:poshScope; New-Item function:\script:Write-host -value "param([object] `$object, `$backgroundColor, `$foregroundColor, [switch] `$nonewline) " -Force | Out-Null;$private:p = PrePoshGitPrompt; if(--$global:poshScope -eq 0) {Remove-Item function:\Write-Host -Force}}PoshGitPrompt}

```

dahlbyk:
Closes #239, inspired by need for #357 hack to fix #355.

Specific changes:

1. Chocolatey installer no longer adds code to `$PROFILE` that saves the pre–posh-git `prompt`
2. Chocolatey installer also strips that legacy code from `$PROFILE` (if you want to keep the old behavior, make sure `PoshGitPrompt` doesn't appear in those lines)
3. Added uninstaller that both deletes `%ChocolateyBinRoot%\poshgit` and removes lines we might have added to `$PROFILE`

We might consider reverting the second commit (and maybe the first) for now, to avoid a breaking change before  [1.0](https://github.com/dahlbyk/posh-git/issues/328) drops.

### DIFF
--- a/chocolatey/tools/chocolateyInstall.ps1
+++ b/chocolatey/tools/chocolateyInstall.ps1
@@ -1,9 +1,5 @@
 ﻿try {
     $binRoot = Get-BinRoot
-
-    $oldPromptOverride = "if(Test-Path Function:\Prompt) {Rename-Item Function:\Prompt PrePoshGitPrompt -Force}"
-    $newPromptOverride = "function Prompt() {if(Test-Path Function:\PrePoshGitPrompt){++`$global:poshScope; New-Item function:\script:Write-host -value `"param([object] ```$object, ```$backgroundColor, ```$foregroundColor, [switch] ```$nonewline) `" -Force | Out-Null;`$private:p = PrePoshGitPrompt; if(--`$global:poshScope -eq 0) {Remove-Item function:\Write-Host -Force}}PoshGitPrompt}"
-
     $poshgitPath = join-path $binRoot 'poshgit'
 
     try {
@@ -25,6 +21,8 @@
         #If old profile exists replace with new one and make sure prompt preservation function is on top
         $pgitExample = "$pgitDir\profile.example.ps1"
         foreach($line in $oldProfile) {
+            if ($line -like '*PoshGitPrompt*') { continue; }
+
             if($line.ToLower().Contains("$poshgitPath".ToLower())) {
                 $line = ". '$pgitExample'"
             }
--- a/chocolatey/tools/chocolateyInstall.ps1
+++ b/chocolatey/tools/chocolateyInstall.ps1
@@ -14,7 +14,7 @@
     $version = "v$Env:chocolateyPackageVersion"
     if ($version -eq 'v') { $version = 'master' }
     $poshGitInstall = if ($env:poshGit ) { $env:poshGit } else { "https://github.com/dahlbyk/posh-git/zipball/$version" }
-    Install-ChocolateyZipPackage 'poshgit' $poshGitInstall $poshgitPath
+    $zip = Install-ChocolateyZipPackage 'poshgit' $poshGitInstall $poshgitPath
     $currentVersionPath = Get-ChildItem "$poshgitPath\*posh-git*\" | Sort-Object -Property LastWriteTime | Select-Object -Last 1
 
     if(Test-Path $PROFILE) {
--- a/chocolatey/tools/chocolateyInstall.ps1
+++ b/chocolatey/tools/chocolateyInstall.ps1
@@ -1,8 +1,4 @@
-﻿function Insert-Script([ref]$originalScript, $script) {
-    if(!($originalScript.Value -Contains $script)) { $originalScript.Value += $script }
-}
-
-try {
+﻿try {
     $binRoot = Get-BinRoot
 
     $oldPromptOverride = "if(Test-Path Function:\Prompt) {Rename-Item Function:\Prompt PrePoshGitPrompt -Force}"
@@ -16,7 +12,7 @@ try {
         remove-item $poshgitPath -recurse -force
       }
     } catch {
-      Write-Host 'Could not remove poshgit folder'
+      Write-Host "Could not remove `'$poshgitPath`'"
     }
 
     $poshGitInstall = if($env:poshGit -ne $null){ $env:poshGit } else {'https://github.com/dahlbyk/posh-git/zipball/master'}
@@ -30,26 +26,15 @@ try {
         $pgitExample = "$pgitDir\profile.example.ps1"
         foreach($line in $oldProfile) {
             if($line.ToLower().Contains("$poshgitPath".ToLower())) {
-                Insert-Script ([REF]$newProfile) $oldPromptOverride
                 $line = ". '$pgitExample'"
             }
             if($line.Trim().Length -gt 0) {  $newProfile += $line }
         }
-        # Save any previous Prompt logic
-        Insert-Script ([REF]$newProfile) $oldPromptOverride
         Set-Content -path $profile -value $newProfile -Force
     }
 
     $installer = Join-Path $pgitDir 'install.ps1'
     & $installer
-
-    $newProfile = @(Get-Content $PROFILE)
-    Insert-Script ([REF]$newProfile) "Rename-Item Function:\Prompt PoshGitPrompt -Force"
-
-    # function that will run previous prompt logic and then the poshgit logic
-    # all output from previous prompts will be swallowed
-    Insert-Script ([REF]$newProfile) $newPromptOverride
-    Set-Content -path $profile  -value $newProfile -Force
 } catch {
   try {
     if($oldProfile){ Set-Content -path $PROFILE -value $oldProfile -Force }
--- a/chocolatey/tools/chocolateyInstall.ps1
+++ b/chocolatey/tools/chocolateyInstall.ps1
@@ -24,6 +24,9 @@
             if($line -like '. *posh-git*profile.example.ps1*') {
                 $line = ". '$currentVersionPath\profile.example.ps1'"
             }
+            if($line -like 'Import-Module *\src\posh-git.psd1*') {
+                $line = "Import-Module '$currentVersionPath\src\posh-git.psd1'"
+            }
             $newProfile += $line
         }
         Set-Content -path $profile -value $newProfile -Force
--- a/chocolatey/tools/chocolateyInstall.ps1
+++ b/chocolatey/tools/chocolateyInstall.ps1
@@ -11,7 +11,9 @@
       Write-Host "Could not remove `'$poshgitPath`'"
     }
 
-    $poshGitInstall = if ($env:poshGit ) { $env:poshGit } else { 'https://github.com/dahlbyk/posh-git/zipball/master' }
+    $version = "v$Env:chocolateyPackageVersion"
+    if ($version -eq 'v') { $version = 'master' }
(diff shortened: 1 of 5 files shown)


chocolatey/tools/chocolateyUninstall.ps1
```diff
@@ -0,0 +1,34 @@
...
+        $newProfile = @()
+
+        foreach($line in $oldProfile) {
+            if ($line -like '*PoshGitPrompt*') { continue; }
+            if ($line -like '*Load posh-git example profile*') { continue; }
+            if ($line -like '*posh-git*profile.example.ps1*') { continue; }
+
+            $newProfile += $line
+        }
+        Set-Content -path $profile -value $newProfile -Force
```
rkeithhill: Set-Content defaults to ASCII.  Either we should detect the current encoding like in the other install.ps1 or maybe use `-Encoding Utf8`?  But that will add the UTF-8 BOM.  File encodings are such a PITA.

rkeithhill: Also, should we backup the profile before we actually modify it - just in case?

chocolatey/tools/chocolateyInstall.ps1
```diff
@@ -29,29 +21,20 @@ try {
...
             if($line.ToLower().Contains("$poshgitPath".ToLower())) {
-                Insert-Script ([REF]$newProfile) $oldPromptOverride
                 $line = ". '$pgitExample'"
             }
-            if($line.Trim().Length -gt 0) {  $newProfile += $line }
+            $newProfile += $line
         }
-        # Save any previous Prompt logic
-        Insert-Script ([REF]$newProfile) $oldPromptOverride
         Set-Content -path $profile -value $newProfile -Force
```
rkeithhill: See comment below about file encoding and creating a backup.

chocolatey/tools/chocolateyUninstall.ps1
mwrock: Oh man yeah that BOM issue is a total pain. 👍 to backing up the old `$PROFILE`

dahlbyk: Conveniently enough, `install.ps1` already knows how to [detect encoding](https://github.com/dahlbyk/posh-git/blob/547d99a1465adbc80a5265a70f2e02d9e74b0af7/install.ps1#L23-L51). Reasonable to tackle that as a separate PR since this case is just copy/paste of `chocolateyInstall.ps1`?

dahlbyk:
I've brought this up-to-date after #361 and am working toward using `Add-PoshGitToProfile` and (new) `Remove-PoshGitFromProfile` for all activities that touch the profile.

dahlbyk:
> Yup, that's why I forced the result to be an array @() and then checked the count (length) of the array. Casting to [bool] like that works but isn't particularly "powershelly". And we typically don't use a return when it isn't needed. :-)

Yeah, I should have just gone to bed. I was trying to troubleshoot a false positive and just couldn't figure it out. Turns out `-match` returns a boolean when matching a string and acts as a filter when used against an array/list.

```powershell
PS> 'posh-git' -match 'posh-git'
True
PS> @('posh-git') -match 'posh-git'
posh-git
```

In retrospect, it's obvious: the false positive was a from a one-line `$PROFILE.CurrentUserAllHosts` I just happened to have created. So the fix is to force to array _before_ the `-match`: https://github.com/dahlbyk/posh-git/pull/358/commits/c8bd3b59d9e511536cf889cc4e6299f0e91d7de3. Added a test for this scenario, too.

rkeithhill:
> I should have just gone to bed

Heh, yeah that happens with me far too often.  Sometimes when I look at my "late night code" the next day or two - wow!  My wife keeps telling me - "you think you're twenty something but ... you're not!"  :-)

dahlbyk:
This is blocked until #376, which has absorbed this PR's non-Chocolatey changes, has been merged.

dahlbyk:
> This is blocked until #376, which has absorbed this PR's non-Chocolatey changes, has been merged.

Rebased on `master` and basically rewrote the entire branch with some refactoring, encoding awareness, etc. Haven't actually tested the Chocolatey package yet, but could use some feedback on the evolution.

chocolatey/tools/chocolateyInstall.ps1
```diff
@@ -1,55 +1,40 @@
...
     } catch {
-      Write-Host 'Could not remove poshgit folder'
+      Write-Host "Could not remove `'$poshgitPath`'"
     }
 
-    $poshGitInstall = if($env:poshGit -ne $null){ $env:poshGit } else {'https://github.com/dahlbyk/posh-git/zipball/master'}
+    $poshGitInstall = if ($env:poshGit ) { $env:poshGit } else { 'https://github.com/dahlbyk/posh-git/zipball/master' }
     Install-ChocolateyZipPackage 'poshgit' $poshGitInstall $poshgitPath
-    $pgitDir = Dir "$poshgitPath\*posh-git*\" | Sort-Object -Property LastWriteTime | Select -Last 1
+    $currentVersionPath = Get-ChildItem "$poshgitPath\*posh-git*\" | Sort-Object -Property LastWriteTime | Select-Object -Last 1
```
rkeithhill: I wonder if it would be better to sort on the directory name **if** it includes a lexicographic sortable version number?

dahlbyk: This is really just defensive programming. A few lines up we try to `remove-item $poshgitPath`, so this folder will almost always have a single subfolder.

dahlbyk:
Finally getting around to testing out the updated packaging.

* [ ] Clean install
* [x] Upgrade
* [x] Uninstall

WARNING: do not take a dependency on anything here, nor the `v0.7.0-pre1` tag. I'm hacking and slashing all over the place.

rkeithhill:
You've been busy.  ;-)

dahlbyk:
Local install/uninstall cycle worked (Chocolatey v0.9.9.6) so I've pushed to Chocolatey.org: https://chocolatey.org/packages/poshgit/0.7.0-pre1. `choco install poshgit -pre` worked for me as well.

HOWEVER, automated package verification failed with Chocolatey v0.10.3 with the same error reported in #381. I haven't had a chance to dig in, but the full error trace is available here: https://gist.github.com/choco-bot/8fd09f341ed80aa5cf6ae01f39b49763

dahlbyk:
> HOWEVER, automated package verification failed with Chocolatey v0.10.3 with the same error reported in #381. I haven't had a chance to dig in, but the full error trace is available here: https://gist.github.com/choco-bot/8fd09f341ed80aa5cf6ae01f39b49763

Fixed in https://github.com/dahlbyk/posh-git/pull/384

dahlbyk:
Added another "fix" for the latest error: https://github.com/dahlbyk/posh-git/issues/381#issuecomment-276192730

It looks like Chocolatey 0.10.* might have changed availability of `$PROFILE`?

rkeithhill:
Hmm, is Chocolatey implementing their own PowerShell host process (and not initializing $PROFILE)?  Hard to believe Chocolatey would just whack the $PROFILE variable but unfortunately that variable is not readonly, so it is possible.

src/Utils.ps1
```diff
@@ -118,6 +118,17 @@ function Add-PoshGitToProfile([switch]$AllHosts, [switch]$Force, [switch]$WhatIf
         }
     }
 
+    if (!$profilePath) {
+        Write-Warning "Skipping add of posh-git import; no profile found."
+        Write-Verbose "`$profilePath          = '$profilePath'"
+        Write-Verbose "`$PROFILE              = '$PROFILE'"
+        Write-Verbose "CurrentUserCurrentHost = '${PROFILE.CurrentUserCurrentHost}'"
+        Write-Verbose "CurrentUserAllHosts    = '${PROFILE.CurrentUserAllHosts}'"
```
rkeithhill: Here you want `"... '$($Profile.CurrentUserAllHosts)'"`  `${}` is handy when you want to separate the variable from the next text e.g. `"${foo}${bar}"`.

rkeithhill: Ditto for the other two below.

dahlbyk: Good catch, not sure what I was thinking.

dahlbyk:
> Hmm, is Chocolatey implementing their own PowerShell host process (and not initializing $PROFILE)? Hard to believe Chocolatey would just whack the $PROFILE variable but unfortunately that variable is not readonly, so it is possible.

At this point I just want it to work. I may bug @ferventcoder later to figure out why `$PROFILE` is missing its properties.

Check out https://github.com/dahlbyk/posh-git/pull/358/commits/4c8cbc6734bd5645cebb136b32a569df009ca20e for my workarounds. Everything finally seems to be working as expected with Chocolatey v0.10.3.

`choco install poshgit --version=v0.7.0-pre4`

rkeithhill:
I hear ya on the "just want it to work" sentiment. :-)  Otherwise, LGTM.

ferventcoder:
> `$PROFILE` is missing its properties.

Chocolatey indeed hosting our own PowerShell Host, since 0.9.10. I thought we implemented a way for `$PROFILE` to work similarly to how it did before, but it may not have much more than the one profile path as its settings. My understanding is that a profile path is dependent on the PowerShell Host and not guaranteed. At least from what I've read. Open to suggestions to improve the custom Choco Host though.

rkeithhill:
@ferventcoder thanks for confirming.  It makes more sense for us to handle this in the chocolatey install scripts.  Any CurrentHost Profile scripts provided by Chocolatey would only apply when the install package is running in Chocolatey and not later when the current host is PowerShell console or ISE.

dahlbyk:
> > `$PROFILE` is missing its properties.
> 
> Chocolatey indeed hosting our own PowerShell Host, since 0.9.10. I thought we implemented a way for $PROFILE to work similarly to how it did before, but it may not have much more than the one profile path as its settings. My understanding is that a profile path is dependent on the PowerShell Host and not guaranteed. At least from what I've read. Open to suggestions to improve the custom Choco Host though.

I think we're good now, but it wouldn't hurt for Chocolatey to learn about the extended `$PROFILE` properties: https://github.com/chocolatey/choco/issues/1154.

ferventcoder:
Not that you need it, but I just wanted to throw it out there that there is a chocolateyBeforeModify.ps1 (as of 0.9.10) that you can use before upgrade and before uninstall to ensure things are not running, etc.

PR #361, opened 2017-01-08
Implement Add-PoshGitToProfile command

rkeithhill:
Offer to update their current/all host profile if posh-git not detected in their profile scripts.  Note: we do not offer to update the "all users" profile scripts because that requires elevation and probably not worth it.

Add Pester tests for the new Utils.ps1 functions.

### DIFF
--- a/Utils.ps1
+++ b/Utils.ps1
@@ -16,7 +16,7 @@ function Invoke-NullCoalescing {
 
 Set-Alias ?? Invoke-NullCoalescing -Force
 
-function Invoke-Utf8ConsoleCommand ([ScriptBlock]$cmd) {
+function Invoke-Utf8ConsoleCommand([ScriptBlock]$cmd) {
     $currentEncoding = [Console]::OutputEncoding
     try {
         [Console]::OutputEncoding = [Text.Encoding]::UTF8
@@ -38,7 +38,7 @@ function Add-ImportModuleToProfile {
         [Parameter(Position = 1, Mandatory = $true)]
         [ValidateNotNullOrEmpty()]
         [string]
-        $ScriptRoot
+        $ModuleBasePath
     )
 
     # Check if profile script exists
@@ -55,11 +55,11 @@ function Add-ImportModuleToProfile {
     }
 
     # Check if the location of this module file is in the PSModulePath
-    if (Test-InModulePath $ScriptRoot) {
+    if (Test-InPSModulePath $ModuleBasePath) {
         $profileContent += "Import-Module posh-git"
     }
     else {
-        $profileContent += "Import-Module '$ScriptRoot\posh-git.psd1'"
+        $profileContent += "Import-Module '$ModuleBasePath\posh-git.psd1'"
     }
 
     Set-Content -LiteralPath $profilePath -Value $profileContent -Encoding $encoding
@@ -136,7 +136,11 @@ function Get-LocalOrParentPath($path) {
     return $null
 }
 
-function Test-InModulePath {
+function Get-PoshGitModulePath {
+    Get-Module posh-git -ListAvailable | ForEach-Object { $_.ModuleBase }
+}
+
+function Test-InPSModulePath {
     param (
         [Parameter(Position=0, Mandatory=$true)]
         [ValidateNotNull()]
@@ -144,7 +148,7 @@ function Test-InModulePath {
         $Path
     )
 
-    $modulePaths = $env:PSModulePath -split ';'
+    $modulePaths = Get-PoshGitModulePath
     if (!$modulePaths) { return $false }
 
     $pathStringComparison = Get-PathStringComparison
@@ -166,7 +170,7 @@ function Test-PoshGitImportedInScript {
     @((Get-Content $Path -ErrorAction SilentlyContinue) -match 'posh-git').Count -gt 0
 }
 
-function dbg ($Message, [Diagnostics.Stopwatch]$Stopwatch) {
+function dbg($Message, [Diagnostics.Stopwatch]$Stopwatch) {
     if ($Stopwatch) {
         Write-Verbose ('{0:00000}:{1}' -f $Stopwatch.ElapsedMilliseconds,$Message) -Verbose # -ForegroundColor Yellow
     }
--- a/Utils.ps1
+++ b/Utils.ps1
@@ -63,9 +63,14 @@ function Add-PoshGitToProfile([switch]$AllHosts, [switch]$Force) {
     $underTest = $false
 
     $profilePath = if ($AllHosts) { $PROFILE.CurrentUserAllHosts } else { $PROFILE.CurrentUserCurrentHost }
-    if ($args -gt 0) {
+
+    # Under test, we override some variables using $args as a backdoor.
+    if ($args.Count -gt 0) {
         $profilePath = [string]$args[0]
         $underTest = $true
+        if ($args.Count -gt 1) {
+            $ModuleBasePath = [string]$args[1]
+        }
     }
 
     if (!$Force) {
@@ -175,8 +180,9 @@ function Get-LocalOrParentPath($path) {
     return $null
 }
 
-function Get-PoshGitModulePath {
-    Get-Module posh-git -ListAvailable | ForEach-Object { $_.ModuleBase }
+function Get-PSModulePath {
+    $modulePaths = $Env:PSModulePath -split ';'
+    $modulePaths
 }
 
 function Test-InPSModulePath {
@@ -187,7 +193,7 @@ function Test-InPSModulePath {
         $Path
     )
 
-    $modulePaths = Get-PoshGitModulePath
+    $modulePaths = Get-PSModulePath
     if (!$modulePaths) { return $false }
 
     $pathStringComparison = Get-PathStringComparison
--- a/Utils.ps1
+++ b/Utils.ps1
@@ -51,7 +51,7 @@ function Add-ImportModuleToProfile {
         # Doesn't exist, so create it
         New-Item -Path $profilePath -ItemType File
         $profileContent = @()
-        $encoding = 'ascii'
+        $encoding = 'utf8'
     }
 
     # Check if the location of this module file is in the PSModulePath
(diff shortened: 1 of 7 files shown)


posh-git.psm1
```diff
@@ -92,6 +92,57 @@ if (!$currentPromptDef -or ($currentPromptDef -eq $defaultPromptDef)) {
     Set-Item Function:\prompt -Value $poshGitPromptScriptBlock
 }
 
+# If running interactive, check if user want's `import-module posh-git` added to their profile
+if (!$NoProfileCheck -and ($MyInvocation.ScriptName.Length -eq 0)) {
```
dahlbyk: Is the `Length` check more idiomatic than `!$MyInvocation.ScriptName`?

Also, s/want's/wants above. :trollface:

test/Utils.Tests.ps1
```diff
@@ -0,0 +1,122 @@
...
+                $scriptRoot = Split-Path $profilePath -Parent
+                Add-ImportModuleToProfile $profilePath $scriptRoot
+                Test-Path -LiteralPath $profilePath | Should Be $true
+                Get-FileEncoding $profilePath | Should Be 'ascii'
+                $content = Get-Content $profilePath
+                $content.Count | Should Be 1
+                @($content)[0] | Should BeExactly "Import-Module '$scriptRoot\posh-git.psd1'"
+            }
+            finally {
+                Remove-Item $profilePath -ErrorAction SilentlyContinue
```
dahlbyk:
```powershell
BeforeEach {
    $profilePath = [System.IO.Path]::GetTempFileName()
}
AfterEach {
    Remove-Item $profilePath -ErrorAction SilentlyContinue
}
```
❓

Utils.ps1
```diff
@@ -26,6 +27,98 @@ function Invoke-Utf8ConsoleCommand ([ScriptBlock]$cmd) {
...
+    if (Test-Path -LiteralPath $profilePath) {
+        $profileContent = @(Get-Content -LiteralPath $profilePath)
+        $profileContent += [string]::Empty # Empty line between previous profile script and import statement
+        $encoding = Get-FileEncoding $profilePath
+    }
+    else {
+        # Doesn't exist, so create it
+        New-Item -Path $profilePath -ItemType File
+        $profileContent = @()
+        $encoding = 'ascii'
```
rkeithhill: Check.

Utils.ps1
```diff
@@ -26,6 +27,98 @@ function Invoke-Utf8ConsoleCommand ([ScriptBlock]$cmd) {
...
+    }
+
+    # Check if the location of this module file is in the PSModulePath
+    if (Test-InModulePath $ScriptRoot) {
+        $profileContent += "Import-Module posh-git"
```
rkeithhill: I would avoid hard-coding the module path as a "bad practice" for PowerShell modules. I can switch to using Get-Module -ListAvailable.

install.ps1
```diff
@@ -1,5 +1,8 @@
 param([switch]$WhatIf = $false)
 
+# Dot source for Get-FileEncoding
+. $PSScriptRoot\Utils.ps1
```
rkeithhill: Thought I caught that.  We can just use Split-Path $MyInvocation.MyCommand.Path -Parent instead.

test/Utils.Tests.ps1
rkeithhill: Heh, I started with that.  Then wrote a few tests that didn't need $profilePath and removed Before/AfterEach.  Then I moved those tests to a different Context.  Fixed.

Utils.ps1
rkeithhill: Doh, that makes the Pester test a little harder.  Oh well, hopefully not too much harder.  :-)

rkeithhill: OK this is making the Pester tests a lot harder.  For some reason, Mock Get-Module is not working.  Will look at it more tomorrow.

install.ps1
```diff
@@ -1,5 +1,9 @@
 param([switch]$WhatIf = $false)
 
+# Dot source for Get-FileEncoding
+$scriptRoot = Split-Path $MyInvocation.MyCommand.Path -Parent
```
dahlbyk: We have `$installDir = Split-Path $MyInvocation.MyCommand.Path -Parent` on L22. Rename this one and delete that line?

dahlbyk:
Minor nitpick, otherwise revisions look good. Just need to fix up how we detect if we should `Import-Module posh-git`!

dahlbyk:
Is this ready for a final review? (Go ahead and remove WIP from title if so.)

rkeithhill:
Yes. Ready for final review.

rkeithhill:
Hmm, hold off on this a bit.  Joel says ScriptName is empty in implicit loading scenarios and he's right.  Let's see if we can use Get-PSCallstack instead.

rkeithhill:
I'm becoming quite nervous about this PR.  Using `Get-PSCallstack` has its issues as well.  In general I thought a stack count of 2 would be reliably indicative of importing from the prompt e.g. here is a manual import call stack:
```
PS C:\Users\Keith> ipmo posh-git

Command          : posh-git.psm1
Location         : posh-git.psm1: line 95
Arguments        : {}
ScriptName       : C:\Users\Keith\Documents\WindowsPowerShell\Modules\posh-git\0.7.0\posh-git.psm1
ScriptLineNumber : 95
InvocationInfo   : System.Management.Automation.InvocationInfo
Position         : Get-PSCallstack | fl * | Out-String
FunctionName     : <ScriptBlock>

Command          : <ScriptBlock>
Location         : <No file>
Arguments        : {}
ScriptName       :
ScriptLineNumber : 1
InvocationInfo   : System.Management.Automation.InvocationInfo
Position         : ipmo posh-git
FunctionName     : <ScriptBlock>
```
And when the PowerShell console profile runs, the stack depth is three:
```
Command          : posh-git.psm1
Location         : posh-git.psm1: line 95
Arguments        : {}
ScriptName       : C:\Users\Keith\Documents\WindowsPowerShell\Modules\posh-git\0.7.0\posh-git.psm1
ScriptLineNumber : 95
InvocationInfo   : System.Management.Automation.InvocationInfo
Position         : Get-PSCallstack | fl * | Out-String
FunctionName     : <ScriptBlock>

Command          : profile.ps1
Location         : profile.ps1: line 9
Arguments        : {}
ScriptName       : C:\Users\Keith\Documents\WindowsPowerShell\profile.ps1
ScriptLineNumber : 9
InvocationInfo   : System.Management.Automation.InvocationInfo
Position         : Import-Module posh-git
FunctionName     : <ScriptBlock>

Command          : <ScriptBlock>
Location         : <No file>
Arguments        : {}
ScriptName       :
ScriptLineNumber : 1
InvocationInfo   : System.Management.Automation.InvocationInfo
Position         : . 'C:\Users\Keith\Documents\WindowsPowerShell\profile.ps1'
FunctionName     : <ScriptBlock>
```
But then I started to notice that the PowerShell in VSCode on my machine keep stalling during init.  And of course it was stalling because of the query to modify the import.  Sigh...  There is something about the PowerShell extension custom host that results in a stack depth one less than expected when processing profiles e.g.:
```
Command          : posh-git.psm1
Location         : posh-git.psm1: line 96
Arguments        : {}
ScriptName       : C:\\Users\\Keith\\Documents\\WindowsPowerShell\\Modules\\posh-git\\0.7.0\\posh-git.psm1
ScriptLineNumber : 96
InvocationInfo   : System.Management.Automation.InvocationInfo
Position         : Get-PSCallstack | Fl * | Out-String
FunctionName     : <ScriptBlock>

Command          : profile.ps1
Location         : profile.ps1: line 10
Arguments        : {}
ScriptName       : C:\\Users\\Keith\\Documents\\WindowsPowerShell\\profile.ps1
ScriptLineNumber : 10
InvocationInfo   : System.Management.Automation.InvocationInfo
Position         : Get-GitStatus
FunctionName     : <ScriptBlock>
```
Another issue is that if posh-git hasn't been imported yet and you execute one of its commands, it will be auto-imported.  That will cause the Modify Profile prompt to appear. That could be confusing.

How about we rejigger this work into a command like `Add-PoshGitToProfile`?  It will result in a pretty simple install process e.g.:
```powershell
Install-Module posh-git -Scope User
Add-PoshGitToProfile
```
Otherwise I'm worried there will be some unanticipated scenario (other custom hosts, remoting, etc) that will result in posh-git stalling the script waiting for input.

rkeithhill:
Looking that the Docker instructions for [getting going with the posh-docker PowerShell module](https://docs.docker.com/docker-for-windows) see the section `Set up tab completion in PowerShell`.  Their profile mod approach is simply:
```powershell
Install-Module -Scope CurrentUser posh-docker -Force
Add-Content $PROFILE "`nImport-Module posh-docker"
```
I like the simplicity.  They could have also mentioned that if you want posh-docker loaded in all hosts, then use:
```powershell
Install-Module -Scope CurrentUser posh-docker -Force
Add-Content $PROFILE.CurrentUserAllHosts "`nImport-Module posh-docker"
```
And Add-Content seems to preserve the original file's encoding.

dahlbyk:
If you're nervous, I'm nervous. Let's skip the magic. I appreciate the simplicity of Docker's approach, but wouldn't mind being a bit more helpful. How about:

`Add-PoshGitToProfile [-AllHosts] [-Force]`

- `-AllHosts` would switch from `CurrentUserCurrentHost` to `CurrentUserAllHosts`.
- `-Force` would skip our check if posh-git is already being imported by a profile.
- We would keep our check to see if we can import by name or need to import the full path.
- The Chocolatey install would replace its complexity with an import and `Add-PoshGitToProfile`

> And Add-Content seems to preserve the original file's encoding.

And specifying `-Encoding UTF8` seems to provide the encoding to use if the file doesn't exist. :100:

We'll still need to handle encoding (we don't today) in Chocolatey install/uninstall, but `Add-Content` seems sufficient for `Add-PoshGitToProfile`.

rkeithhill:
> I appreciate the simplicity of Docker's approach, but wouldn't mind being a bit more helpful.

Agreed.  Everything else sounds good as well.  I'm starting on it.

rkeithhill:
I'm ready to merge this.

PR #362, opened 2017-01-08
Revert Get-SshPath back to use $Env:Home.

rkeithhill:
But still not use ~, use $home instead.  Also fixed a few more PSSA warnings.  I just .. can't .. help .. myself.  :-)


dahlbyk:
> Also fixed a few more PSSA warnings. I just .. can't .. help .. myself. :-)

No complaint here.

PR #336, opened 2016-12-30
Normalize line endings

dahlbyk:
We don't need to merge this before we ship 0.7 (#327), but this will close #334.

There is a good chance GitHub will refuse to merge the remaining open PRs. To resolve, we will want to set `git config --global merge.renormalize true`, ensure `master` is up-to-date, and then either:

1. `git rebase master` to replay changes after the normalization
2. `git merge master` to bring in these changes

PR #363, opened 2017-01-08
WIP allow prompt char > to reflect nesting level.

rkeithhill:
To be a good prompt citizen in PowerShell, the > (or whatever prompt char) should be displayed once for nesting level 0, twice for nesting level 1, and so on.

In order to make that happen, we have to allow for deferred expansion of the Prompt*Suffix settings.

See https://msdn.microsoft.com/en-us/powershell/reference/5.1/microsoft.powershell.core/about/about_prompts for more details.

### DIFF
--- a/posh-git.psm1
+++ b/posh-git.psm1
@@ -85,7 +85,7 @@ if (!$currentPromptDef -or ($currentPromptDef -eq $defaultPromptDef)) {
         }
 
         $global:LASTEXITCODE = $origLastExitCode
-        $promptSuffix
+        $ExecutionContext.SessionState.InvokeCommand.ExpandString($promptSuffix)
 '@)
 
     # Set the posh-git prompt as the default prompt
--- a/GitPrompt.ps1
+++ b/GitPrompt.ps1
@@ -96,8 +96,8 @@ $global:GitPromptSettings = New-Object PSObject -Property @{
 
     EnableWindowTitle                           = 'posh~git ~ '
 
-    PromptSuffix                                = '> '
-    PromptDebugSuffix                           = ' [DBG]>> '
+    PromptSuffix                                = '$(''>'' * ($nestedPromptLevel + 1)) '
+    PromptDebugSuffix                           = ' [DBG]$(''>'' * ($nestedPromptLevel + 1)) '
 
     Debug                                       = $false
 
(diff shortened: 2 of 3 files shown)


posh-git.psm1
```diff
@@ -38,7 +38,7 @@ $poshGitPromptScriptBlock = $null
 
 $currentPromptDef = if ($funcInfo = Get-Command prompt -ErrorAction SilentlyContinue) { $funcInfo.Definition }
 
-# HACK: If prompt is missing, create a global one we can overwrite with Set-Item
+# HACK: If prompt is missing, create a global one we can overwrite with Set-Item
```
rkeithhill: Sometimes GitHub diff gets a little whacked.  I didn't change that line at all.

dahlbyk: Apparently I [committed](https://github.com/dahlbyk/posh-git/commit/bd70123320f84bac803eed9ca1d54771b186df14) a LF here instead of CRLF:
notepad _2017-01-08_00-32-42

VS Code is fixing it. #336 will prevent this type of change in the future.

posh-git.psm1
```diff
@@ -85,7 +85,7 @@ if (!$currentPromptDef -or ($currentPromptDef -eq $defaultPromptDef)) {
         }
 
         $global:LASTEXITCODE = $origLastExitCode
-        $promptSuffix
+        $ExecutionContext.SessionState.InvokeCommand.ExpandString($promptSuffix)
```
dahlbyk: TIL

I probably would have done something way more involved like check if `$promptSuffix` was a `[ScriptBlock]` to execute every time.

dahlbyk:
> WIP allow prompt char

Commit message style note: WIP PRs are 100% approved, but generally I would avoid "WIP" in commit messages unless you're planning to rewrite history before we actually review and merge the code. Which I'm totally fine with (in fact, clean history is encouraged!), if we communicate as such. Unclear if that's your intent here, so thought I'd mention it.

posh-git.psm1
rkeithhill: Thanks.  I was a bit puzzled on that one.

rkeithhill:
> I assume ExpandString is available in PS v2?

Yes, I tested on v2 to make sure.

> I would avoid "WIP" in commit messages unless you're planning to rewrite history

So don't use WIP in commit msg unless it really is WIP?  :-)  Sorry about that. I thought it would change because I was thinking of replacing Prompt*Suffix with PromptChar.  But I've changed my mind since.

PR #364, opened 2017-01-08
Clean up Get-SshPath test.

rkeithhill:
Also experimenting with how to mock git.exe and add some Get-GitStatus tests.

### DIFF
--- a/test/Get-GitStatus.Tests.ps1
+++ b/test/Get-GitStatus.Tests.ps1
@@ -1,0 +1,88 @@
+# For info on Pester mocking see - http://www.powershellmagazine.com/2014/09/30/pester-mock-and-testdrive/
+Describe 'Get-GitStatus Tests' {
+    Context 'Get-GitStatus Working Directory Tests' {
+        BeforeAll {
+            function global:git {
+                $cmdline = "$args"
+                switch ($cmdline) {
+                    '--version' { 'git version 2.11.0.windows.1' }
+                    'help'      { Get-Content $PSScriptRoot\git-help.txt  }
+                    default     {
+                        $res = Invoke-Expression "git.exe $cmdline"
+                        $res
+                    }
+                }
+            }
+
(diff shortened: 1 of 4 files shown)


test/Get-GitStatus.Tests.ps1
```diff
@@ -0,0 +1,88 @@
...
+            $status.HasIndex | Should Be $false
+            $status.HasUntracked | Should Be $false
+            $status.Index.Count | Should Be 0
+            $status.Working.Count | Should Be 2
+            $status.Working[0] | Should Be "test/Foo.Tests.ps1"
+            $status.Working[1] | Should Be "test/Bar.Tests.ps1"
+            $status.AheadBy  | Should Be 0
+            $status.BehindBy | Should Be 0
+            $status.StashCount | Should Be 0
+            $status.Upstream -eq $null | Should Be $true
```
dahlbyk: We can discuss more as additional tests are written, but I try to limit asserts to only what I'm specifically testing, e.g. `HasIndex`, `HasUntracked`, `Index` and `Working` for this set. A different set of tests would have asserts for `Branch`, `AheadBy`, `BehindBy`, `Upstream` (and `UpstreamGone`).

rkeithhill: Fair enough.  I was a bit overexcited, getting the git.exe mocking to work.  :-)

rkeithhill: Also, was in mode of testing everything.  Feel free to tweak the assertions at any time.

dahlbyk: > I was a bit overexcited, getting the git.exe mocking to work. :-)

🎉  Understood 🎉 

> Feel free to tweak the assertions at any time.

I'm not worried about it today, unless you're planning to add more tests. What you have so far is good to merge as a starting point unless you'd rather keep going.

PR #365, opened 2017-01-08
Add AppVeyor Build

rkeithhill:
Of course this requires someone to setup an AppVeyor account - I think.
Not sure what's up with the line endings.  On my file system, this file has CRLF at the end of every line.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,33 @@
+os:
+  - "WMF 5"
+
+branches:
+  only:
+    - master
+
+install:
+  - ps: |
+      Install-PackageProvider -Name NuGet -MinimumVersion 2.8.5.201 -Force | Out-Null
+      Install-Module Pester -MinimumVersion 3.4.0 -Scope CurrentUser -Force | Out-Null
+
+build_script:
+  - ps: |
+      # Nothing to build at this point, just run Pester tests
+      $testResultsFile = Join-Path $pwd TestResults.xml
+      $testing = @{
+          Script       =
+          OutputFile   = $testResultsFile
+          OutputFormat = "NUnitXml"
+          PassThru     = $true
+          Verbose      = $VerbosePreference
+      }
+      Import-Module Pester
+      Invoke-Pester @testing
+      if (Test-Path $testResultsFile) {
+          (New-Object 'System.Net.WebClient').UploadFile("https://ci.appveyor.com/api/testresults/nunit/$($env:APPVEYOR_JOB_ID)", $testResultsFile)
+      }
+      else {
+          # Terminate the script to fail the build
+          $Error | Format-List * -Force
+          exit 1;
+      }
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -10,12 +10,14 @@ install:
       Install-PackageProvider -Name NuGet -MinimumVersion 2.8.5.201 -Force | Out-Null
       Install-Module Pester -MinimumVersion 3.4.0 -Scope CurrentUser -Force | Out-Null
 
-build_script:
+build: false
+
+test_script:
   - ps: |
-      # Nothing to build at this point, just run Pester tests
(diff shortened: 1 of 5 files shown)


dahlbyk:
YAML is rather specific about [EOL normalization](), so I'm assuming what you were seeing is a quirk of the syntax highligher. I've taken the liberty of rewriting this branch with `*.yml` added to `.gitattributes` and the new file normalized to LF. Go ahead and `fetch` then `reset` your local branch to the remote (`git reset "@{u}"` if you set up tracking with `git push -u`).

AppVeyor is set up for this project and the PR is building. :tada:

appveyor.yml
```diff
@@ -0,0 +1,33 @@
...
+  - ps: |
+      Install-PackageProvider -Name NuGet -MinimumVersion 2.8.5.201 -Force | Out-Null
+      Install-Module Pester -MinimumVersion 3.4.0 -Scope CurrentUser -Force | Out-Null
+
+build_script:
+  - ps: |
+      # Nothing to build at this point, just run Pester tests
+      $testResultsFile = Join-Path $pwd TestResults.xml
+      $testing = @{
+          Script       =
```
dahlbyk: Build failed because no value is specified here:

> OutputFile : The term 'OutputFile' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the spelling of the name, or if a path was included, verify that the path is correct and try again.

dahlbyk:
A bit of searching on using Pester with AppVeyor turned up this series:

* http://ramblingcookiemonster.github.io/GitHub-Pester-AppVeyor/
* http://ramblingcookiemonster.github.io/Github-Pester-AppVeyor-Part-2/
* https://github.com/RamblingCookieMonster/Invoke-Parallel/blob/master/appveyor.yml
* https://github.com/RamblingCookieMonster/Invoke-Parallel/blob/master/Tests/appveyor.pester.ps1

Of particular interest is that the tests are also run against PowerShell 2.0.

rkeithhill:
Thanks for the links!  I wonder if the `[Runspace]::DefaultRunspace.InitialSessionSta ...` issue is due to v3.  I thought I had tested that on v3.  If it had be executed with v2 it wouldn't have executed that line.

rkeithhill:
Hmm, the runspace line works on PS v3 in my VM.  Grrr..

rkeithhill:
OK,the test failure now is legit.  I think I need to update my Get-SshPath tests to reflect changing that function back to using $env:home and Resolve-Path.  The think about Resolve-Path is that it returns nothing for a path that doesn't exist.  Is that the correct semantics for this function?

rkeithhill:
Aha, looks like folks view an empty result from Get-SshPath as a bug #268.  @dahlbyk Can you confirm that?  If so then `Resolve-Path` is not the way to go.  It will error and not return anything if the path doesn't exist.

wekempf:
I've never liked Resolve-Path as it does two things you rarely want to do together (normalize the path AND validate it exists). What's not clear here to me is why you're using Resolve-Path? In fact, I'm not sure what this Get-SshPath command does. There's no help for it, and the result when I run it is the path to *a* key file. What's the point of that?

Anyway, if you need to normalize (make absolute) a path without validating it exists, I've got code that does that. Of course, since this will always be in the file system, it actually should be fairly trivial to implement using System.IO.Path, while my code works even if you've set-location to some other provider.

rkeithhill:
> What's not clear here to me is why you're using Resolve-Path

I was putting the code back to a state - more or less - before I changed it.  However I did change it to **not** use `~`  which isn't valid in all drives.  So I think the fix here is to simply remove `Resolve-Path` as it no longer needs to resolve `~`.

dahlbyk:
I'm good with removing `Resolve-Path`.

rkeithhill:
@dahlbyk I'm happy with this.  OK if I merge it?

PR #366, opened 2017-01-12
Add Worktree tab completion

Flueworks:
Git worktrees have received some love in recent releases of git.
This implements a simple tab completion for worktrees.
Autocompletes path and branch: git worktree <path> <ref>


dahlbyk:
Thanks!

PR #367, opened 2017-01-16
Exclude detached HEAD "branch" from tab completion

dahlbyk:
When `HEAD` is detached, `git branch` includes a listing along the lines of `(HEAD detached at deadbeef)`. This shouldn't be included when we're completing branch names.

### DIFF
--- a/GitTabExpansion.ps1
+++ b/GitTabExpansion.ps1
@@ -75,7 +75,7 @@ function script:gitBranches($filter, $includeHEAD = $false, $prefix = '') {
         $prefix += $matches['from']
         $filter = $matches['to']
     }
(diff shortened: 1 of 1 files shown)


GitTabExpansion.ps1
```diff
@@ -75,7 +75,7 @@ function script:gitBranches($filter, $includeHEAD = $false, $prefix = '') {
         $prefix += $matches['from']
         $filter = $matches['to']
     }
-    $branches = @(git branch --no-color | ForEach-Object { if($_ -match "^\*?\s*(?<ref>.*)") { $matches['ref'] } }) +
+    $branches = @(git branch --no-color | ForEach-Object { if(!($_ -match "^\* \(HEAD detached .+\)$") -and ($_ -match "^\*?\s*(?<ref>.*)")) { $matches['ref'] } }) +
```
rkeithhill: You could use `-notmatch` here otherwise looks good!

PR #368, opened 2017-01-16
Add more Get-GitStatus Pester tests

rkeithhill:


rkeithhill:
Since these are just tests, I think it's pretty safe to merge in.

PR #371, opened 2017-01-16
Worth adding optional timing output to def prompt?

rkeithhill:
This is enabled with `$GitPromptSettings.DefaultPromptEnableTiming=$true` and then the prompt function outputs `~\GitHub\rkeithhill\posh-git [rkeithhill/enable-default-prompt-timing +0 ~2 -0 !] 55 mS>`.

It also occurred to me that the GitPromptSettings should be prefixed "DefaultPrompt" so folks don't think they apply to their own custom prompt functions.

I'm not entirely sure this is worth it but I could see it coming in handy to get "hard" perf data from users of the default prompt.  If  we decide not to go with this PR, I still think we should consider adding the "Default" prefix to the existing two Prompt*Suffix settings.

In the VSCode dir on my machine (status equiv to origin), the prompt takes 150 mS.  VSCode has 6466 files in the repo.

### DIFF
--- a/posh-git.psm1
+++ b/posh-git.psm1
@@ -92,7 +92,7 @@ if (!$currentPromptDef -or ($currentPromptDef -eq $defaultPromptDef)) {
         if ($GitPromptSettings.DefaultPromptEnableTiming) {
             $sw.Stop()
             $elapsed = $sw.ElapsedMilliseconds
-            Write-Host " $($elapsed)ms" -NoNewline
+            Write-Host " ${elapsed}ms" -NoNewline
         }
 
         $global:LASTEXITCODE = $origLastExitCode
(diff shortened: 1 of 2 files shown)


dahlbyk:
I like the idea. Added commit to compress the display: `[branch] 55ms>`...eh?

posh-git.psm1
```diff
@@ -92,7 +92,7 @@ if (!$currentPromptDef -or ($currentPromptDef -eq $defaultPromptDef)) {
         if ($GitPromptSettings.DefaultPromptEnableTiming) {
             $sw.Stop()
             $elapsed = $sw.ElapsedMilliseconds
-            Write-Host " $elapsed mS" -NoNewline
+            Write-Host " $($elapsed)ms" -NoNewline
```
rkeithhill: Works for me.  You can simplify that a bit `" ${elapsed}ms"`.  In interpolated strings in PowerShell, you can delineate variable names with `${<variable-name>}`.

PR #372, opened 2017-01-17
Report UpstreamGone from cache responses.

cmarcusreid:
Latest GitStatusCache executable can be pulled with Update-GitStatusCache. Functionality won't light up on existing clients without pulling it down.


dahlbyk:
Thanks!

Issue #338, opened 2016-12-30
Spin off SSH Agent module?


dahlbyk:
As I alluded to in https://github.com/dahlbyk/posh-git/issues/328#issuecomment-269794084, a [nontrivial number of issues and open PRs here](https://github.com/dahlbyk/posh-git/search?q=agent&type=Issues) are dedicated to managing `ssh-agent` and Pageant.

Since that problem has nothing to do with Git other than Git for Windows being a common distribution path for MSYS, I'd like to spin the SSH Agent bits off into a separate module. Is anyone interested in taking that on?

cc people who have tried to contribute SSH fixes: @theaquamarine (#133) @paulmarsy (#162) @c-mu (#195) @hjoelr (#297) @StefanScherer (#238) @Ventajou (#148) @nitin88 (#295)

dahlbyk:
Also cc @pcx (#155) @GrahamTheCoder (#211) @kayukin (#316)

pcx:
@dahlbyk I guess you wanted to cc: @pxc :)

dahlbyk:
> @dahlbyk I guess you wanted to cc: @pxc :)

I suppose I did. You're welcome to help, too. 😉

rkeithhill:
@dahlbyk do you want to host this?  If so, what would we call it?  `posh-git-ssh`?  There is already a [posh-ssh](https://www.powershellgallery.com/packages/Posh-SSH/1.7.7) on the PowerShell Gallery but it doesn't seem to deal with the agent or keys.

dahlbyk:
> @dahlbyk do you want to host this? If so, what would we call it? posh-git-ssh? There is already a posh-ssh  on the PowerShell Gallery but it doesn't seem to deal with the agent or keys.

I might as well. @drusellers suggested `posh-sshit`, but that may be slightly inappropriate. Maybe `posh-ssh-agent`? [Twitter can decide.](https://twitter.com/dahlbyk/status/834118721983676417)

> Also, will the upcoming Win32-OpenSSH implementation impact this?

Several of the open issues here are or have been about Win32-OpenSSH, so yes that's in scope.

rkeithhill:
+1 for `posh-ssh-agent`.

nigurr:
Posh-ssh-agent sounds good. :)

nitin88:
I am very bad at naming 😛
`posh-ssh-agent` sounds good 👍

tamj0rd2:
[...]

PR #376, opened 2017-01-26
Simplify install.ps1 and move core module files under src/

rkeithhill:
This moves the core module files under a src dir.  BTW can we move install.ps1 to the chocolatey dir?  Shipping this file in the module may lead to confusion.  Also both it and CheckVersion.ps1 check if git is installed.


test/Utils.Tests.ps1
```diff
@@ -21,7 +22,7 @@ Describe 'Utils Function Tests' {
             Get-FileEncoding $profilePath | Should Be 'utf8'
             $content = Get-Content $profilePath
             $content.Count | Should Be 2
-            $modulePath = Resolve-Path $PSScriptRoot\..
+            # $modulePath = Resolve-Path $PSScriptRoot\..
```
dahlbyk: Delete?

rkeithhill:
Will this cause merge headaches for the outstanding PRs?

dahlbyk:
> Will this cause merge headaches for the outstanding PRs?

Kind of. I wasn't comfortable with the overlap between this and #358, so I've overwritten your changes here based on the non-Chocolatey changes that ended up over there. Net result:

- `profile.example.ps1` has been moved back up to the project root, to avoid breaking those who have used `install.ps1` or Chocolatey until now.
- I also added a warning to that file indicating it will be removed in a future version (presumably 1.0).
- `install.ps1` no longer uses `profile.example.ps1`, instead calling `Import-Module` directly.
  - This is kind of a breaking change, as fresh installs will no longer call `Start-SshAgent -Quiet`. Thoughts? (cc #338)
- Finally, the module files are moved under `src/` along with some missing fixes for references to those files.

Note: there are currently a few commits in this branch marked as `fixup!`. If you see that, it should be a clear indication that the branch is meant to be `rebase -i`'d before it's merged (autosquash FTW).

rkeithhill:
So should I run `git rebase -i master` on this branch and then `git push --force-with-lease` before merging it?

dahlbyk:
> So should I run `git rebase -i master` on this branch and then `git push --force-with-lease` before merging it?

Yep, looks good as merged.

PR #378, opened 2017-01-27
Prep for 0.7.0 release.

rkeithhill:


rkeithhill:
I want to merge this quickly as having the version number bumped to 0.7.0 makes it easier to tell which version of posh-git you have imported.  This will be important if we can get some folks to test posh-git by cloning the repo - just to make sure they have the new bits loaded and not the bits from their installed modules folder.

PR #379, opened 2017-01-28
Improve tab expansion to handle params for push.

### LINKED ISSUE #234: Incorrect tab expansion for `git push --option <remote>`
### Alias

Define the alias `ps` as `push --follow-tags`
### Execution

I would delete the remote branch named `issue-1234`

**Expected**

``` bash
git push [TAB]=>origin [TAB]=>develop...
```

**Current**

```
git ps [TAB]=>develop[TAB]=>develop
```

> Is this the current behavior? Should we strip off the arguments on tab completion?

rkeithhill:
Fix #234  BTW if we decide this is an OK fix (need another set of eyes on my $gitParams regex), I'd like to review the fact that all these switch regex case statements fallthrough to the next (observed while debugging tab expansion).  I don't believe that is necessary.  I'm thinking that most (all) of these should have a break statement.

Started a minimum set of Pester tests for tab completion.  Would be good to flesh these out over time.

BTW the latest PowerShell extension for VS Code release supports "attach to PS host process" which makes debugging tab expansion a breeze.  VS Code FTW!  :-)

### DIFF
--- a/src/GitTabExpansion.ps1
+++ b/src/GitTabExpansion.ps1
@@ -258,14 +258,14 @@ function GitTabExpansionInternal($lastBlock) {
 
         # Handles git push remote <ref>:<branch>
         # Handles git push remote +<ref>:<branch>
-        "^push${gitParams}\s+(?<remote>[^\s-]\S*)(${gitParams}\s+(?<force>\+?)(?<ref>[^\s\:]*\:)(?<branch>\S*))+$" {
+        "^push${gitParams}\s+(?<remote>[^\s-]\S*).*\s+(?<force>\+?)(?<ref>[^\s\:]*\:)(?<branch>\S*)$" {
             gitRemoteBranches $matches['remote'] $matches['ref'] $matches['branch'] -prefix $matches['force']
         }
 
         # Handles git push remote <ref>
         # Handles git push remote +<ref>
         # Handles git pull remote <ref>
-        "^(?:push|pull)${gitParams}\s+(?<remote>[^\s-]\S*)(${gitParams}\s+(?<force>\+?)(?<ref>[^\s\:]*))+$" {
+        "^(?:push|pull)${gitParams}\s+(?<remote>[^\s-]\S*).*\s+(?<force>\+?)(?<ref>[^\s\:]*)$" {
             gitBranches $matches['ref'] -prefix $matches['force']
             gitTags $matches['ref'] -prefix $matches['force']
         }
--- a/src/GitTabExpansion.ps1
+++ b/src/GitTabExpansion.ps1
@@ -29,10 +29,8 @@ $gitflowsubcommands = @{
 }
 
 function script:gitCmdOperations($commands, $command, $filter) {
-    $commands.$command -split ' ' |
-        Where-Object { $_ -like "$filter*" }
+    $commands.$command -split ' ' | Where-Object { $_ -like "$filter*" }
 }
-
 
 $script:someCommands = @('add','am','annotate','archive','bisect','blame','branch','bundle','checkout','cherry',
                          'cherry-pick','citool','clean','clone','commit','config','describe','diff','difftool','fetch',
@@ -40,9 +38,9 @@ $script:someCommands = @('add','am','annotate','archive','bisect','blame','branc
                          'notes','prune','pull','push','rebase','reflog','remote','rerere','reset','revert','rm',
                          'shortlog','show','stash','status','submodule','svn','tag','whatchanged', 'worktree')
 try {
-  if ($null -ne (git help -a 2>&1 | Select-String flow)) {
-      $script:someCommands += 'flow'
-  }
+    if ($null -ne (git help -a 2>&1 | Select-String flow)) {
+        $script:someCommands += 'flow'
+    }
 }
 catch {
     Write-Debug "Search for 'flow' in 'git help' output failed with error: $_"
@@ -52,7 +50,8 @@ function script:gitCommands($filter, $includeAliases) {
     $cmdList = @()
     if (-not $global:GitTabSettings.AllCommands) {
         $cmdList += $someCommands -like "$filter*"
-    } else {
+    }
+    else {
         $cmdList += git help --all |
             Where-Object { $_ -match '^  \S.*' } |
             ForEach-Object { $_.Split(' ', [StringSplitOptions]::RemoveEmptyEntries) } |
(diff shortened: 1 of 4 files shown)


src/GitTabExpansion.ps1
```diff
@@ -29,20 +29,18 @@ $gitflowsubcommands = @{
...
-        Where-Object { $_ -like "$filter*" }
+    $commands.$command -split ' ' | Where-Object { $_ -like "$filter*" }
 }
 
-
 $script:someCommands = @('add','am','annotate','archive','bisect','blame','branch','bundle','checkout','cherry',
                          'cherry-pick','citool','clean','clone','commit','config','describe','diff','difftool','fetch',
-                         'format-patch','gc','grep','gui','help','init','instaweb','log','merge','mergetool','mv',
-                         'notes','prune','pull','push','rebase','reflog','remote','rerere','reset','revert','rm',
+                         'format-patch','gc','grep','gui','help','init','instaweb','ls-files','log','merge','mergetool',
```
dahlbyk: Out of curiosity, why add `ls-files`? You can get tab expansion for all commands (including plumbing like `ls-files`) by setting `$Global:GitTabSettings.AllCommands = $true`, but I'm not sure this belongs with the porcelain commands.

test/TabExpansion.Tests.ps1
```diff
@@ -0,0 +1,38 @@
...
+            $result = & $module GitTabExpansionInternal 'git push  --follow-tags  -u   origin '
+            $result -contains 'master' | Should Be $true
+            $result -contains 'origin/master' | Should Be $true
+            $result -contains 'origin/HEAD' | Should Be $true
+        }
+        It 'Tab completes remote and matching branch with preceding parameters' {
+            $result = & $module GitTabExpansionInternal 'git push  --follow-tags  -u   origin ma'
+            $result | Should BeExactly 'master'
+        }
+    }
```
dahlbyk: Add tests to assert empty result when expanding a missing remote, branch, or remote/branch combination?

It wouldn't be a terrible idea to set a remote and/or branch name with a `-` in it, as a sanity check for the `gitParams` regex.

src/GitTabExpansion.ps1
```diff
@@ -258,22 +258,22 @@ function GitTabExpansionInternal($lastBlock) {
...
-        "^push.* (?<remote>\S+) (?<force>\+?)(?<ref>[^\s\:]*\:)(?<branch>\S*)$" {
+        "^push.* ${gitParams}(?<remote>\S+) (?<force>\+?)(?<ref>[^\s\:]*\:)(?<branch>\S*)$" {
             gitRemoteBranches $matches['remote'] $matches['ref'] $matches['branch'] -prefix $matches['force']
         }
 
         # Handles git push remote <ref>
         # Handles git push remote +<ref>
         # Handles git pull remote <ref>
-        "^(?:push|pull).* (?:\S+) (?<force>\+?)(?<ref>[^\s\:]*)$" {
+        "^(?:push|pull).* ${gitParams}(?<remote>[^\s-]\S*) (?<force>\+?)(?<ref>[^\s\:]*)$" {
```
dahlbyk: Not sure how to keep this maintainable over time (other than a proper Git arg parser 🙄 ), but it's legal to have Git parameters between the remote and ref specs, and between individual ref specs (e.g. `git push origin -f master -u my-branch`). I'm fine with this incremental improvement for now, but food for thought.

src/GitTabExpansion.ps1
rkeithhill: I'll remove this.  Didn't know about `AllCommands`.  Thanks.

test/TabExpansion.Tests.ps1
rkeithhill: I can add some more tests.  Can a branch name start with - or --?

src/GitTabExpansion.ps1
rkeithhill: Well, we could try `"^(?:push|pull).* ${gitParams}(?<remote>[^\s-]\S*)${gitParams} (?<force>\+?)(?<ref>[^\s\:]*)$"` and add some more tests? I'm OK with the incremental improvement unless we think it would be as simple as the above tweaks to handle this situation.

test/TabExpansion.Tests.ps1
dahlbyk: No, that's one of the few non-system restrictions on branch/remote names.

src/GitTabExpansion.ps1
dahlbyk: Something that like that seems like it should do the trick - give it a try?

rkeithhill:
@dahlbyk I don't have a good test case for the remote refspec (with `:` in it).  Can you provide me with such a refspec that returns results?

test/TabExpansion.Tests.ps1
```diff
@@ -0,0 +1,102 @@
...
+        It 'Tab completes all branches' {
+            $result = & $module GitTabExpansionInternal 'git push origin '
+            $result -contains 'master' | Should Be $true
+            $result -contains 'origin/master' | Should Be $true
+            $result -contains 'origin/HEAD' | Should Be $true
+        }
+        It 'Tab completes all :branches' {
+            $result = & $module GitTabExpansionInternal 'git push origin :'
+            $result -contains ':master' | Should Be $true
+            # Is the following a valid branch name in a refspec?
```
rkeithhill: Thanks.  I wasn't sure about that but it didn't look right (with spaces, ->).  Thanks for updating the sample specs.  I "think" the multiple ref specs should be easy to implement/test.  I'll look at that tonight.

rkeithhill:
> Stretch goal, since you're already up to your neck in this

Oh sure, goad me on.  :-)  OK last commit addresses this I think.

dahlbyk:
> > Stretch goal, since you're already up to your neck in this
> 
> Oh sure, goad me on. :-) OK last commit addresses this I think.

Not quite, since `ref:branch` can be mixed with other forms. I added/adjusted the tests and made the regex dumber between the remote and force/ref, so I think we're good.

rkeithhill:
Yeah, getting into usages of git I'm not so familiar with.  Well, you've fixed it.  Want me to rebase or merge as is (didn't see any fixup)?

rkeithhill:
Nevermind, there is a fixup.

rkeithhill:
@dahlbyk FYI I rebased -i --autosquash this branch and pushed with --force-with-lease in case you want to make more changes.  Otherwise, I'm feeling pretty good about merging this in.

Issue #344, opened 2016-12-31
Replace or improve $GitPromptSettings


rkeithhill:
It is really hard to find a setting in the output of $GitPrompSettings:
```
LocalWorkingStatusForegroundColor         : DarkRed
FileModifiedText                          : ~
AfterStashBackgroundColor                 : Black
WorkingForegroundBrightColor              : Red
RepositoriesInWhichToDisableFileStatus    : {}
IndexForegroundColor                      : DarkGreen
BranchForegroundColor                     : Cyan
BeforeStashForegroundColor                : Red
EnablePromptStatus                        : True
AfterBackgroundColor                      : Black
FileRemovedText                           : -
BeforeIndexBackgroundColor                : Black
AfterText                                 : ]
BeforeForegroundColor                     : Yellow
LocalWorkingStatusSymbol                  : !
BranchAheadStatusSymbol                   : ↑
LocalDefaultStatusSymbol                  :
IndexForegroundBrightColor                : Green
IndexBackgroundColor                      : Black
LocalStagedStatusForegroundColor          : Cyan
ShowStatusWhenZero                        : True
WorkingForegroundColor                    : DarkRed
BranchIdenticalStatusToBackgroundColor    : Black
BeforeIndexForegroundColor                : DarkGreen
BranchIdenticalStatusToSymbol             : ≡
BranchBehindAndAheadStatusForegroundColor : Yellow
LocalDefaultStatusForegroundColor         : DarkGreen
DescribeStyle                             :
BranchBackgroundColor                     : Black
LocalStagedStatusBackgroundColor          : Black
BranchIdenticalStatusToForegroundColor    : Cyan
AfterStashText                            : )
BeforeBackgroundColor                     : Black
BeforeIndexForegroundBrightColor          : Green
EnableStashStatus                         : False
Debug                                     : False
LocalWorkingStatusForegroundBrightColor   : Red
LocalWorkingStatusBackgroundColor         : Black
LocalStagedStatusSymbol                   : ~
DelimText                                 :  |
BranchBehindAndAheadStatusSymbol          : ↕
WorkingBackgroundColor                    : Black
BeforeStashBackgroundColor                : Black
DefaultForegroundColor                    : Gray
BranchNameLimit                           : 0
AfterForegroundColor                      : Yellow
DelimBackgroundColor                      : Black
BranchBehindStatusForegroundColor         : Red
AfterStashForegroundColor                 : Red
LocalDefaultStatusForegroundBrightColor   : Green
[...]

PR #382, opened 2017-01-30
Use pscustomobject cast for sorted output of $GitPromptSettings

lzybkr:
To get output sorted like the original source for `$GitPromptSettings`,
we can use a cast instead of `New-Object`.

This does change the type of the object in V2 from `PSObject` to
`hashtable`, but it doesn't appear to cause any problems because
property syntax works either way.


rkeithhill:
Verified.  V2 still works (although the properties don't display in definition order).  OTOH V5 (and probably V3/4) do display the properties in the order they were defined.

rkeithhill:
Sufficiently fixes #344 for 0.7.0.

PR #384, opened 2017-01-30
Remove Mandatory from Test-PoshGitImportedInScript -Path

### LINKED ISSUE #381: Can't install latest release
Hi,

I can't install posh-git via choco or `install-package/install-module`.
When I install using `install-package` or `install-module` posh-git gets installed (I can see it in 'modules' folder), but `import-module posh-git` doesn't bring it.
If I install using choco setup fails:
```
Get-BinRoot is going to be deprecated in v1 and removed in v2. It has been replaced with Get-ToolsLocation (starting with v0.9.10), however many packages no longer require a special separate directory since package folders no longer have versions on them. Some do though and should continue to use Get-ToolsLocation.
Extracting C:\Users\username\AppData\Local\Temp\chocolatey\poshgit\0.6.1.20160330\dahlbyk-posh-git-v0.6.1-224-g5ada37f.zip to C:\tools\poshgit...                                                                                                                                      
C:\tools\poshgit
ERROR: Cannot bind argument to parameter 'Path' because it is an empty string.                                                              
The install of poshgit was NOT successful.                                                                                                  
Error while running 'C:\ProgramData\chocolatey\lib\poshgit\tools\chocolateyInstall.ps1'.                                                    
```

I only managed to install it by first installing PsGet and then running `install-module`, then manually addding `import-module` to profile. But in this case I've got version 0.7.0.

System info:
```
posh-git version: 0.6.1, PSVersion: 5.1.14393.693, git version 2.10.1.windows.1, OS: Microsoft Windows NT 10.0.14393.0
```

dahlbyk:
Fix #381 

> full error trace is available here: https://gist.github.com/choco-bot/8fd09f341ed80aa5cf6ae01f39b49763

```
2017-01-30 18:46:34,436 [ERROR] - ERROR: Cannot bind argument to parameter 'Path' because it is an empty string.
 at Add-PoshGitToProfile, C:\tools\poshgit\dahlbyk-posh-git-9f6b256\src\Utils.ps1: line 98
at <ScriptBlock>, C:\tools\poshgit\dahlbyk-posh-git-9f6b256\install.ps1: line 6
at <ScriptBlock>, C:\ProgramData\chocolatey\lib\poshgit\tools\chocolateyInstall.ps1: line 42
at <ScriptBlock>, C:\ProgramData\chocolatey\helpers\chocolateyScriptRunner.ps1: line 48
at <ScriptBlock>, <No file>: line 1
```

As of https://github.com/dahlbyk/posh-git/pull/376, `install.ps1` passes `$WhatIf` on to `Add-PoshGitToProfile`, which matches a hack added to let tests override `$profilePath`. Ultimately this causes `$profilePath` to be empty when passed as a `Mandatory` parameter to `Test-PoshGitImportedInScript`. Two-part fix:

1. `Test-PoshGitImportedInScript` can correctly handle a `$null`/empty `$Path`, so it's safer to remove `Mandatory`.
2. We can skip the test argument hack by checking the argument type.

### DIFF
--- a/src/Utils.ps1
+++ b/src/Utils.ps1
@@ -205,7 +205,7 @@ function Test-InPSModulePath {
 
 function Test-PoshGitImportedInScript {
     param (
-        [Parameter(Position=0, Mandatory=$true)]
+        [Parameter(Position=0)]
         [string]
         $Path
     )
--- a/src/Utils.ps1
+++ b/src/Utils.ps1
@@ -84,7 +84,8 @@ function Add-PoshGitToProfile([switch]$AllHosts, [switch]$Force, [switch]$WhatIf
     $profilePath = if ($AllHosts) { $PROFILE.CurrentUserAllHosts } else { $PROFILE.CurrentUserCurrentHost }
 
     # Under test, we override some variables using $args as a backdoor.
-    if ($args.Count -gt 0) {
+    # TODO: Can we just turn these into optional parameters with well-defined behavior?
+    if (($args.Count -gt 0) -and ($args[0] -is [string])) {
         $profilePath = [string]$args[0]
         $underTest = $true
         if ($args.Count -gt 1) {


src/Utils.ps1
```diff
@@ -84,7 +84,8 @@ function Add-PoshGitToProfile([switch]$AllHosts, [switch]$Force, [switch]$WhatIf
     $profilePath = if ($AllHosts) { $PROFILE.CurrentUserAllHosts } else { $PROFILE.CurrentUserCurrentHost }
 
     # Under test, we override some variables using $args as a backdoor.
-    if ($args.Count -gt 0) {
+    # TODO: Can we just turn these into optional parameters with well-defined behavior?
```
dahlbyk: @rkeithhill this question is for you. Can we just make `$ProfilePath` and real (optional) argument that is documented to override `$AllHosts`?

dahlbyk: Alternatively, we could use environment variables to override.

rkeithhill: I'd rather not make the public parameters any more complicated for the sake of tests.  Yes, we could use env vars but those are global to the process and could impact tests if they ever ran in parallel (which they currently don't).

PR #387, opened 2017-01-31
Add AbbreviateHomeDirectory setting

### LINKED ISSUE #386: Add option to abbreviate home directory in prompt
Currently, posh-git abbreviates my home directory to `~` in the prompt. By default, PowerShell prints out the whole path, which I prefer.

`PS C:\Users\kshaw>` becomes `~>` with posh-git

After upgrading recently (thanks to Github for Windows), I noticed that my home directory is now abbreviated, and went looking for a `GitPromptSetting` that would allow me to turn off the abbreviation, but to no avail.

I'd like to see an `AbbreviateHomeDirectory` (or similar) setting in the `GitPromptSettings` that would allow me to enable or disable this home directory abbreviation.

shawmanz32na:
Adds a setting to optionally disable the home directory abbreviation in
the prompt. By default, this is set to true, preserving the existing
behavior.

Closes #386

### DIFF
--- a/src/GitPrompt.ps1
+++ b/src/GitPrompt.ps1
@@ -100,7 +100,7 @@ $global:GitPromptSettings = [pscustomobject]@{
     DefaultPromptSuffix                         = '$(''>'' * ($nestedPromptLevel + 1)) '
     DefaultPromptDebugSuffix                    = ' [DBG]$(''>'' * ($nestedPromptLevel + 1)) '
     DefaultPromptEnableTiming                   = $false
-    DefaultPromptAbbreviateHomeDirectory        = $true
+    DefaultPromptAbbreviateHomeDirectory        = $false
 
     Debug                                       = $false
 
--- a/src/GitPrompt.ps1
+++ b/src/GitPrompt.ps1
@@ -101,6 +101,8 @@ $global:GitPromptSettings = [pscustomobject]@{
     DefaultPromptDebugSuffix                    = ' [DBG]$(''>'' * ($nestedPromptLevel + 1)) '
     DefaultPromptEnableTiming                   = $false
 
+    AbbreviateHomeDirectory                     = $true
+
     Debug                                       = $false
 
     BranchNameLimit                             = 0
--- a/src/GitPrompt.ps1
+++ b/src/GitPrompt.ps1
@@ -100,8 +100,7 @@ $global:GitPromptSettings = [pscustomobject]@{
     DefaultPromptSuffix                         = '$(''>'' * ($nestedPromptLevel + 1)) '
     DefaultPromptDebugSuffix                    = ' [DBG]$(''>'' * ($nestedPromptLevel + 1)) '
     DefaultPromptEnableTiming                   = $false
-
-    AbbreviateHomeDirectory                     = $true
+    DefaultPromptAbbreviateHomeDirectory        = $true
 
     Debug                                       = $false
 
--- a/src/posh-git.psm1
+++ b/src/posh-git.psm1
@@ -66,7 +66,7 @@ if (!$currentPromptDef -or ($currentPromptDef -eq $defaultPromptDef)) {
         }
 
         # Abbreviate path by replacing beginning of path with ~ *iff* the path is in the user's home dir
-        if ($currentPath -and $currentPath.StartsWith($Home, $stringComparison))
+        if ($GitPromptSettings.AbbreviateHomeDirectory -and $currentPath -and $currentPath.StartsWith($Home, $stringComparison))
         {
             $currentPath = "~" + $currentPath.SubString($Home.Length)
         }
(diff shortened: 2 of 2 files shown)


src/GitPrompt.ps1
```diff
@@ -101,6 +101,8 @@ $global:GitPromptSettings = [pscustomobject]@{
     DefaultPromptDebugSuffix                    = ' [DBG]$(''>'' * ($nestedPromptLevel + 1)) '
     DefaultPromptEnableTiming                   = $false
 
+    AbbreviateHomeDirectory                     = $true
```
rkeithhill: Can we change this to `DefaultPromptAbbrevHomeDir` or something with the `DeafultPrompt` prefix?

shawmanz32na: After re-reading, this change only affects the default prompt, so sure.

Would `DefaultPromptAbbreviateHomeDirectory` be appropriate?

shawmanz32na:
I looked through the test code, but didn't see any tests for the default prompt. Did I miss anything?

Creating a new test suite for the default prompt seems out of the scope of what I'm trying to accomplish here, so if I'm correct that there is not an existing test suite for this particular feature, I'd like to skip writing a test for this, since it's so simple, and since making this change in my posh-git installation results in the expected behavior ;-) .

src/GitPrompt.ps1
rkeithhill: It's a bit long but still not the longest property name.  :-)  Folder is a bit shorter than Directory but I guess it doesn't matter that much.

rkeithhill:
We just started adding Pester tests.  I believe the built-in prompt could easily be tested.  I'll look into adding some prompt tests tonight.

src/GitPrompt.ps1
shawmanz32na: Given the option, I personally prefer 'Directory', so unless you feel strongly about using `Folder` to save the three characters, I'd prefer to use 'Directory'.

I'll change to `DefaultPromptAbbreviateHomeDirectory`, and you can veto as necessary.

shawmanz32na:
^ It appears that I'm a dummy and fat-fingered the commit message.... I suppose I'll just hope you squash it when it gets merged in.

rkeithhill:
Squash and merge isn't enabled for this repo but we'll get by with the typo.  Let me think about the default for this setting a bit.

src/GitPrompt.ps1
```diff
@@ -100,6 +100,7 @@ $global:GitPromptSettings = [pscustomobject]@{
     DefaultPromptSuffix                         = '$(''>'' * ($nestedPromptLevel + 1)) '
     DefaultPromptDebugSuffix                    = ' [DBG]$(''>'' * ($nestedPromptLevel + 1)) '
     DefaultPromptEnableTiming                   = $false
+    DefaultPromptAbbreviateHomeDirectory        = $true
```
dahlbyk: I probably sound like a broken record, but let's default to `$false` for folks who are used to posh-git uncustomized.

Bonus points if you find a place to mention this new setting in the README.

dahlbyk:
> ^ It appears that I'm a dummy and fat-fingered the commit message.... I suppose I'll just hope you squash it when it gets merged in.

You are welcome to fix up the commit message and force push to this branch to update the PR before we merge. Or not, up to you what you want to be recorded for all time. 😀

rkeithhill:
>  but let's default to $false for folks who are used to posh-git uncustomized.

Agreed. Not only posh-git but PowerShell uncustomized as well.

dahlbyk:
> You are welcome to fix up the commit message and force push to this branch to update the PR before we merge.

Nevermind, I'm too eager to merge. Let the record show the misleading commit message is my fault.

shawmanz32na:
> Nevermind, I'm too eager to merge. Let the record show the misleading commit message is my fault.

It was a close race - I was force pushing as you were merging, but it appears I lost.

What do we want to do about the extra settings discussed here? Should I add them (and documentation)? How about tests? Did anyone find some time to get the base test written so I can add some more tests for this functionality?

rkeithhill:
I'll handle the other setting and tests in a separate PR.

shawmanz32na:
Is there an opportunity for me to help, or am I just going to slow you guys down?

rkeithhill:
I've got it.  Besides the setting and tests there are several places of documentation I need to clean up / add info on this (and the other new setting) as well.  BTW thanks for your PR and getting us thinking about how the default posh-git prompt should be configured.

shawmanz32na:
Alrighty. Cheers to you and the team for all the great work. Glad I could contribute a bit.

dahlbyk:
> Is there an opportunity for me to help, or am I just going to slow you guys down?

Sounds like @rkeithhill has this under control, but check out our [up-for-grabs](https://github.com/dahlbyk/posh-git/labels/Up-for-Grabs) issues for other opportunities to contribute! :tada: (And check http://up-for-grabs.net/ to find other projects looking for help!)

PR #394, opened 2017-02-02
Adds aliasing support for TortoiseGit commands

eamodio:
Also allows configuration of TortoiseGit commands

### DIFF
--- a/src/GitTabExpansion.ps1
+++ b/src/GitTabExpansion.ps1
@@ -194,7 +194,7 @@ function GitTabExpansionInternal($lastBlock) {
     # Handles tgit <command> (tortoisegit)
     if ($lastBlock -match "^$(Get-AliasPattern tgit) (?<cmd>\S*)$") {
         # Need return statement to prevent fall-through.
-        return $tortoiseGitCommands | Where-Object { $_ -like "$($matches['cmd'])*" }
+        return $Global:TortoiseGitSettings.TortoiseGitCommands.Keys.GetEnumerator() | sort | Where-Object { $_ -like "$($matches['cmd'])*" }
     }
 
     # Handles gitk
--- a/src/GitTabExpansion.ps1
+++ b/src/GitTabExpansion.ps1
@@ -194,7 +194,7 @@ function GitTabExpansionInternal($lastBlock) {
     # Handles tgit <command> (tortoisegit)
     if ($lastBlock -match "^$(Get-AliasPattern tgit) (?<cmd>\S*)$") {
         # Need return statement to prevent fall-through.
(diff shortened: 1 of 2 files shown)


src/GitTabExpansion.ps1
```diff
@@ -194,7 +194,7 @@ function GitTabExpansionInternal($lastBlock) {
     # Handles tgit <command> (tortoisegit)
     if ($lastBlock -match "^$(Get-AliasPattern tgit) (?<cmd>\S*)$") {
         # Need return statement to prevent fall-through.
-        return $tortoiseGitCommands | Where-Object { $_ -like "$($matches['cmd'])*" }
+        return $Global:TortoiseGitSettings.TortoiseGitCommands.Keys.GetEnumerator() | sort | Where-Object { $_ -like "$($matches['cmd'])*" }
```
dahlbyk: s/sort/Sort-Object/

(also in TortoiseGit.ps1)

src/TortoiseGit.ps1
```diff
@@ -11,60 +11,74 @@ function private:Get-TortoiseGitPath {
...
+    "config" = "settings";
+    "stash" = "stash";
+    "stashapply" = "stashapply";
+    "stashsave" = "stashsave";
+    "subadd" = "subadd";
+    "subsync" = "subsync";
+    "subupdate" = "subupdate";
+    "switch" = "switch";
+    "checkout" = "switch";
+    "sync" = "sync";
```
dahlbyk: Would it make sense to alias `fetch` to `sync`?

src/TortoiseGit.ps1
```diff
@@ -11,60 +11,74 @@ function private:Get-TortoiseGitPath {
...
-"stashapply",
-"subadd",
-"subupdate",
-"subsync",
-"reflog",
-"refbrowse",
-"sync",
-"repostatus"
-) | sort
+}
```
dahlbyk: EOL @ EOF, please!

(Tip: we have a [.editorconfig](http://editorconfig.org/) that will enforce this for supported editors.)

src/GitTabExpansion.ps1
rkeithhill: Good catch.  PowerShell Core running on Linux or macOS does not have the sort alias defined for Sort-Object.  So calling sort would execute the native sort utility.

src/TortoiseGit.ps1
eamodio: Fixed :)

src/TortoiseGit.ps1
eamodio: Added.

src/GitTabExpansion.ps1
eamodio: Fixed.

dahlbyk:
:100: :clap: :tada:

PR #396, opened 2017-02-03
Fixed link to git-scm.com

KanjiBates:
Link to git-scm.com was relative, taking you to a non-existing page instead of git-scm.com.


dahlbyk:
Oops. Put text in URL and URL in text. My bad, thanks for the fix!

PR #402, opened 2017-02-06
Add more info including thank you section.

rkeithhill:
Whipped this out fast.  There may be some typos.

PR #403, opened 2017-02-06
Fix tab expansion for non-ASCII file paths

### LINKED ISSUE #64: Issue with tab expansion and non-English file names
Repro steps:
1. `touch 慣用音.txt`
2. `git add {tab}`

Expected: it expands the filename. 
Actual: it expands to:

```
git add "\346\205\243\347\224\250\351\237\263.txt"
```

That wouldn't be a problem except it doesn't work.
1. `git config core.quotepath true`
2. git add {tab}

Result:

dahlbyk:
Fix #64 now that #359 and #397 have fixed the [encoding issue I encountered at the time](https://github.com/dahlbyk/posh-git/issues/64#issuecomment-7152220).

non-ascii-paths

cc @Haacked

### DIFF
--- a/src/GitTabExpansion.ps1
+++ b/src/GitTabExpansion.ps1
@@ -136,19 +136,19 @@ function script:gitFiles($filter, $files) {
         ForEach-Object { if ($_ -like '* *') { "'$_'" } else { $_ } }
 }
 
-function script:gitIndex($filter) {
+function script:gitIndex($GitStatus, $filter) {
     gitFiles $filter $GitStatus.Index
 }
 
-function script:gitAddFiles($filter) {
+function script:gitAddFiles($GitStatus, $filter) {
     gitFiles $filter (@($GitStatus.Working.Unmerged) + @($GitStatus.Working.Modified) + @($GitStatus.Working.Added))
 }
 
-function script:gitCheckoutFiles($filter) {
+function script:gitCheckoutFiles($GitStatus, $filter) {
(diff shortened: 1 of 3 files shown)


dahlbyk:
> Don't forget to add this to the CHANGELOG.md. :-)

Do I fail at reading comprehension, or does CHANGELOG (even after #402) not mention the other encoding fixes either? Regardless, to avoid conflicts I'll add a note about this to #402 after this gets merged.

dahlbyk:
@rkeithhill do you think we would be able to reasonably test this particularly quote/encoding round-trip?

rkeithhill:
> Do I fail at reading comprehension, or does CHANGELOG (even after #402) not mention the other encoding fixes either

I suspect not.  ;-)  I added stuff I found but no guarantees that I found everything.  

> do you think we would be able to reasonably test this particularly quote/encoding round-trip?

Not sure.  Would have to experiment with it.

dahlbyk:
Figured out a failing test that is made to pass:

```
   Context Add/Reset/Checkout TabExpansion Tests
    [-] Tab completes non-ASCII file name 782ms
      at <ScriptBlock>, C:\Dev\OSS\posh-git\test\TabExpansion.Tests.ps1: line 144
      Expected string length 12 but was 25. Strings differ at index 0.
      Expected: {poshgit.txt}
      But was:  {"posh\342\200\242git.txt"}
      -----------^
      144:             $result | Should BeExactly 'poshgit.txt'
```

Oddly, VS Code is showing `Expected: {poshgit.txt}` when it should be `Expected: {posh•git.txt}` (i.e the UTF-8 bullet is invisble), but it seems to work. ¯\_(ツ)_/¯

PR #406, opened 2017-02-08
DefaultRunspace.Debugger.Inbreakpoint check

skeept:
If I don't add that check, my prompt will always end in PS> which is quite annoying.

If I check which properties are available I get:
C:\opt [master]> [runspace]::DefaultRunspace.Debugger | fl *

IsActive  : False
DebugMode : LocalScript, RemoteScript

So it doesn't include InBreakpoint. Maybe it is only enabled when IsActive is true?

### DIFF
--- a/src/posh-git.psm1
+++ b/src/posh-git.psm1
@@ -87,7 +87,7 @@ if ($ForcePoshGitPrompt -or !$currentPromptDef -or ($currentPromptDef -eq $defau
         Write-VcsStatus
 
         # If stopped in the debugger, the prompt needs to indicate that in some fashion
-        $hasInBreakpoint = [runspace]::DefaultRunspace.Debugger | Get-member -Name InBreakPoint -MemberType property
+        $hasInBreakpoint = [runspace]::DefaultRunspace.Debugger | Get-Member -Name InBreakPoint -MemberType property
         $debugMode = (Test-Path Variable:/PSDebugContext) -or ($hasInBreakPoint -and [runspace]::DefaultRunspace.Debugger.InBreakpoint)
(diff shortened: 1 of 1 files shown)


src/posh-git.psm1
```diff
@@ -87,7 +87,8 @@ if ($ForcePoshGitPrompt -or !$currentPromptDef -or ($currentPromptDef -eq $defau
         Write-VcsStatus
 
         # If stopped in the debugger, the prompt needs to indicate that in some fashion
-        $debugMode = (Test-Path Variable:/PSDebugContext) -or [runspace]::DefaultRunspace.Debugger.InBreakpoint
+        $hasInBreakpoint = [runspace]::DefaultRunspace.Debugger | Get-member -Name InBreakPoint -MemberType property
```
rkeithhill: Minor nit, can you upcase the `m` in `M`.   Also what version of PowerShell are you using?  Otherwise LGTM.

rkeithhill: Nevermind on the version, I see you're on 4.0.

src/posh-git.psm1
```diff
@@ -87,7 +87,8 @@ if ($ForcePoshGitPrompt -or !$currentPromptDef -or ($currentPromptDef -eq $defau
         Write-VcsStatus
 
         # If stopped in the debugger, the prompt needs to indicate that in some fashion
-        $debugMode = (Test-Path Variable:/PSDebugContext) -or [runspace]::DefaultRunspace.Debugger.InBreakpoint
+        $hasInBreakpoint = [runspace]::DefaultRunspace.Debugger | Get-Member -Name InBreakPoint -MemberType property
```
rkeithhill: Sorry, one more minor nit.  The name the `p` in `Breakpoint` should be lower case everywhere.  There is one instance on this line and one on the next line.  Again, sorry I didn't catch this the first time.

PR #412, opened 2017-02-11
Add debug script to see what appveyor issue is.

rkeithhill:


rkeithhill:
Found this online - http://git.661346.n2.nabble.com/git-push-output-goes-into-stderr-td6758028.html
```
> When I create a local branch and then push it to remote. I find that 
> the output without error goes into stderr, is this expected? 

Progress output are sent to the stderr stream. 

In general, any program or script is buggy if it assumes that some output 
that are emitted to the standard error output from programs it invokes 
indicates an error (IIRC, that includes tcl/tk). 
```
and
```
when the two sides of the conversation have an established protocol to 
talk to each other and exchange data, any extraneous stuff (progress 
message, tracing/debugging logs...) will confuse the protocol. 
Shunting that aside to STDERR helps avoid protocol confusion. 
```
and
```
Stdout has always been about "the main program output" and stderr about 
diagnostic messages. With a program whose main function is to generate 
output (e.g., "git tag -l", it's very easy to know that the list of tags 
is the main program output (which you don't want to pollute with 
anything else), and any problems or even general chattiness goes to 
stderr. 

But with a program whose main function is to perform an action, like 
"git push", I think there are really two ways to look at it: 

  1. There is no main output; any progress or status update is just 
     diagnostic chat, and should go to stderr. 

  2. The main output is the status report; it goes to stdout, and 
     progress updates go to stderr. 
```
Normally a PowerShell host like the console host will just hand over the console stream handles to the native app so it can use them directly (enables old char mod graphics apps like edit.com).  However, if you redirect a stream, the PowerShell console host no longer hands over the handles, PowerShell gets back stderr and generates ErrorRecords (red text).  

I believe the "RemoteHostImplementation" host used by appveyor (and indeed any remote host) will not attempt to give stream handles to the native app but handle stderr directly and generate ErrorRecords. 

So for this test, running on AppVeyor, I'll just redirect stderr to $null.

PR #413, opened 2017-02-12
Quote tab completion items with PS special chars.

### LINKED ISSUE #293: Tab expansion starting with '#' needs to be escaped
When using tab-expansion on branch-names, if they start with '#' then it, and the rest of the line will get ignored by powershell since '#' marks the start of a line comment. Escaping it with ` does help though, e.g.:

`
git checkout `#1-branch-name
`

This is only necessary when the name starts with the pound-sign though, as names like `feature/#1-branch-name` works fine without any escaping.

rkeithhill:
Fix #293

### DIFF
--- a/src/GitTabExpansion.ps1
+++ b/src/GitTabExpansion.ps1
@@ -51,6 +51,15 @@ catch {
     Write-Debug "Search for 'flow' in 'git help' output failed with error: $_"
 }
 
+filter quoteStringWithSpecialChars {
+    if ($_ -and ($_ -match '\s+|#|@|\$|;|\{|\}|\(|\)')) {
+        "'" + $_ + "'"
+    }
+    else {
+        $_
+    }
+}
+
 function script:gitCommands($filter, $includeAliases) {
     $cmdList = @()
     if (-not $global:GitTabSettings.AllCommands) {
@@ -86,7 +95,8 @@ function script:gitBranches($filter, $includeHEAD = $false, $prefix = '') {
 
     $branches |
         Where-Object { $_ -ne '(no branch)' -and $_ -like "$filter*" } |
-        ForEach-Object { $prefix + $_ }
+        ForEach-Object { $prefix + $_ } |
+        quoteStringWithSpecialChars
 }
 
 function script:gitRemoteUniqueBranches($filter) {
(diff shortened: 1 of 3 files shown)


rkeithhill:
If we merge this for 0.7.0, don't forget to update CHANGELOG.md.

src/GitTabExpansion.ps1
```diff
@@ -51,6 +51,15 @@ catch {
     Write-Debug "Search for 'flow' in 'git help' output failed with error: $_"
 }
 
+filter quoteStringWithSpecialChars {
+    if ($_ -and ($_ -match '\s+|#|@|\$|;|\{|\}|\(|\)')) {
+        "'" + $_ + "'"
```
dahlbyk: 1. Why not `"'$_'"`?
2. We should also handle values that include `'`.

src/GitTabExpansion.ps1
```diff
@@ -51,6 +51,15 @@ catch {
     Write-Debug "Search for 'flow' in 'git help' output failed with error: $_"
 }
 
+filter quoteStringWithSpecialChars {
+    if ($_ -and ($_ -match '\s+|#|@|\$|;|\{|\}|\(|\)')) {
```
dahlbyk: Missing `,`

test/TabExpansion.Tests.ps1
```diff
@@ -137,16 +137,70 @@ Describe 'TabExpansion Tests' {
...
-            $result = & $module GitTabExpansionInternal 'git add ' $GitStatus
+            $gitStatus = & $module Get-GitStatus
 
+            $result = & $module GitTabExpansionInternal 'git add ' $gitStatus
             $result | Should BeExactly $fileName
         }
     }
+
+    Context 'PowerShell Special Chars Tests' {
+        BeforeAll {
```
dahlbyk: This is bound to be a pretty common pattern. Is it possible to write a Pester helper that would encapsulate a `BeforeAll`/`AfterAll` pairing that initializes and cleans up a Git repo, with its path left in `$repoPath`? Something like:

```powershell
Context 'Awesome test' {
    TempRepo {
        # Set up the repo
    }
    It 'Uses $repoPath' {
        # ...
    }
}
```

rkeithhill:
I think I've addressed all of the issues except the last one the Pester helper.  That said, I did move common code into reusable functions (in Shared.ps1).

PR #422, opened 2017-02-16
Fix issue with tab expansion of duplicate aliases.

### LINKED ISSUE #421: When more than one identical alias is defined tab completion fails
### System Details

- posh-git version: 0.7.0
- PowerShell version: 5.1.14393.693
- git version 2.11.0.windows.1
- OS: Microsoft Windows NT 10.0.14393.0

### Issue Description

If there is both a global and local alias for say `co -> checkout`, then tab completion fails, e.g.: `git co ma<tab>` doesn't complete to `master`.  That's because the `expandGitAlias` function is not expecting more than one string that matches.

rkeithhill:
Add Pester test for bug.  Fix #421.

Also, fix that duplicate aliases would appear in the command tab expansion list.

### DIFF
--- a/src/GitTabExpansion.ps1
+++ b/src/GitTabExpansion.ps1
@@ -192,9 +192,9 @@ function script:gitAliases($filter) {
 }
 
 function script:expandGitAlias($cmd, $rest) {
-    $gitAliases = @(git config --get-regexp "^alias\.$cmd`$")
-    if ($gitAliases[0] -match "^alias\.$cmd (?<cmd>[^!].*)`$") {
-        return "git $($Matches['cmd'])$rest"
+    $alias = git config "alias.$cmd"
+    if ($alias) {
+        return "git $alias$rest"
     }
     else {
         return "git $cmd$rest"
--- a/src/GitTabExpansion.ps1
+++ b/src/GitTabExpansion.ps1
@@ -192,7 +192,8 @@ function script:gitAliases($filter) {
 }
 
 function script:expandGitAlias($cmd, $rest) {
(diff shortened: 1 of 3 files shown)


dahlbyk:
Added a few commits to:

1. Make tests less environment-specific (initially failed for me)
2. Use `--get-all` instead of `--get-regexp` for exact matches
3. Refactored alias test setup/teardown (I try to avoid `try`/`finally` in tests, but would love feedback on this change)
4. Tested/fixed that duplicate aliases appeared twice in the command completion list

src/GitTabExpansion.ps1
```diff
@@ -188,12 +188,13 @@ function script:gitAliases($filter) {
                 $alias
             }
         }
-    } | Sort-Object
+    } | Get-Unique | Sort-Object
```
rkeithhill: It is probably a bit better to use `Sort-Object -Unique`.  And yes, that is available on v2 (just checked).

test/TabExpansion.Tests.ps1
```diff
@@ -142,14 +142,60 @@ Describe 'TabExpansion Tests' {
...
+            $repoPath = NewGitTempRepo -MakeInitialCommit
+        }
+        AfterAll {
+            $addedAliases | Where-Object { $_ } | ForEach-Object {
+                git.exe config --global --unset "alias.$_" 2>$null
+            }
+
+            RemoveGitTempRepo $repoPath
+        }
+        It 'Command completion includes unique list of aliases' {
```
rkeithhill: Thanks for fleshing out these tests!

src/GitTabExpansion.ps1
dahlbyk: TIL

Issue #423, opened 2017-02-16
Bad performance when EnableFileStatusFromCache is true


Tasteful:
### System Details

- posh-git version: 0.7.0
- PowerShell version: 5.1.14393.576
- git version 2.11.0.windows.1
- OS: Microsoft Windows NT 10.0.14393.0
- disc: SSD
- git-status-cache-posh-client: 1.0.0

### Issue Description
After installing `git-status-cache-posh-client` with `choco install git-status-cache-posh-client` the performance for posh-git is bad and it takes around 5-8 seconds to get the prompt up after any changes in git (ex `git add -A` or `git commit -m "..."`).

After turning off the git-status-cache with the settings `$Global:GitPromptSettings.EnableFileStatusFromCache=$false` everything is running fast again and the prompt is up after less than 1 second.

- File structure contains 96975 files, some of them in folder that is excluded in `.gitignore`.
- the GIT-LFS extension is used that also contains some larger files.

dahlbyk:
Ping @cmarcusreid. I assume this is an inefficiency in libgit2's status implementation?

@Tasteful: I assume the repo in question is private? Does the cache seem to improve performance on your other repositories?

Tasteful:
Yes, the repository is private (corporate). I have a smaller repository (also private) whit 18473 files and the performance is worse with `git-status-cache-posh-client` than without. When I have the cache turned off the prompt is returned directly, when I turn the cache on it takes about 1 second for the prompt to return.

I was setting `$Global:GitPromptSettings.DefaultPromptEnableTiming=$true` but I didn't get any timing out, do I need to set any other parameter to enable the timing also?

dahlbyk:
> I was setting `$Global:GitPromptSettings.DefaultPromptEnableTiming=$true` but I didn't get any timing out, do I need to set any other parameter to enable the timing also?

That should be sufficient. Curiously, I am able to reproduce your missing timing in GitHub Desktop's Git Shell, but not in a normal PowerShell window. Odd...

Tasteful:
> That should be sufficient. Curiously, I am able to reproduce your missing timing in GitHub Desktop's Git Shell, but not in a normal PowerShell window. Odd...

I'm using normal PowerShell window, started with `Powershell` in run-menu. I was testing also from the GitHub Desktop's Git Shell and that have same behavior with bad performance for me.

dahlbyk:
Out of curiosity, what does this give you?

```powershell
[...]

PR #425, opened 2017-02-17
Ignore pre-0.7 posh-git prompt

dahlbyk:
Inspired by #423, here's a relatively simple fix for folks who copied the old `prompt` function into their `$PROFILE`. I've tried to make it relatively tolerant of tabs/spaces, CRLF/LF, etc.

### DIFF
--- a/src/posh-git.psm1
+++ b/src/posh-git.psm1
@@ -38,6 +38,14 @@ else {
 $poshGitPromptScriptBlock = $null
 
 $currentPromptDef = if ($funcInfo = Get-Command prompt -ErrorAction SilentlyContinue) { $funcInfo.Definition }
+
+# If prompt matches pre-0.7 posh-git prompt, ignore it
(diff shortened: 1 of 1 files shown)


dahlbyk:
I don't think there is much we can do about that, short of teaching `Write-GitSatus` to check `prompt`. That doesn't feel appropriate.

It may be that this change doesn't fix enough problems for us to bother maintaining.

rkeithhill:
> It may be that this change doesn't fix enough problems for us to bother maintaining.

Yup, although we could add a command to "fix up" the user's profile to remove the old stuff.  Or maybe even modify `Add-PoshGitToProfile` to do that, perhaps with an `-Update` or `-Force` parameter.

dahlbyk:
Ambivalence is fun! Leaning toward leaving it because it's not much code, and it at least does warn me locally that the "wrong" `prompt` is being used, even though it's not able to actually override that `prompt` (this is in GitHub Desktop). Doesn't catch everything, but it seems better than not reporting a problematic situation that we _can_ detect.

PR #427, opened 2017-02-18
Remove PowerShell v2 support.

rkeithhill:
Also doing some miscellaneous cleanup (remove Invoke-NullCoalescing, remove testDebugHarness.ps1 in favor of using the Pester Tests debug configuration, rename CheckVersion.ps1 to CheckRequirements.ps1).

Checkout the version number (1.0.0-pre00) in the chocolatey/poshgit.nuspec file.  Not sure which version of NuGet is being used but if v2, I think the prelease string uses a simple lexicographic sort so that pre10 would sort below pre9.  I suspect 1.0 could have more than 10 prerelease choco drops.  I hope not more than 100 though.  :-)

cc https://github.com/dahlbyk/posh-git/issues/163

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -3,7 +3,7 @@ os:
 
 branches:
   only:
-    - master
+    - develop
 
 init:
   - git config --global core.autocrlf true
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -4,6 +4,7 @@ os:
 branches:
   only:
     - develop
+    - master
 
 init:
   - git config --global core.autocrlf true
--- a/chocolatey/poshgit.nuspec
+++ b/chocolatey/poshgit.nuspec
@@ -3,7 +3,7 @@
   <metadata>
     <id>poshgit</id>
     <title>posh-git</title>
-    <version>0.7.0</version>
+    <version>1.0.0-pre00</version>
     <authors>Keith Dahlby, Mark Embling, Jeremy Skinner, Keith Hill</authors>
     <owners>Keith Dahlby</owners>
     <description>### posh-git
--- a/src/posh-git.psd1
+++ b/src/posh-git.psd1
@@ -4,7 +4,7 @@
 RootModule = 'posh-git.psm1'
 
 # Version number of this module.
-ModuleVersion = '1.0.0'
+ModuleVersion = '1.0.0.0'
 
 # ID used to uniquely identify this module
 GUID = '74c9fd30-734b-4c89-a8ae-7727ad21d1d5'
@@ -24,9 +24,6 @@ PowerShellVersion = '3.0'
 # Functions to export from this module
 FunctionsToExport = @(
     'Add-PoshGitToProfile',
-    'Enable-GitColors',
-    'Get-AliasPattern',
-    'Get-GitBranch',
     'Get-GitStatus',
     'Get-GitDirectory',
     'Update-AllBranches',
--- a/src/posh-git.psd1
+++ b/src/posh-git.psd1
@@ -69,7 +69,7 @@ PrivateData = @{
         ReleaseNotes = 'https://github.com/dahlbyk/posh-git/blob/develop/CHANGELOG.md'
 
         # TODO: REMOVE BEFOE RELEASE
-        PrereleaseVersion = 'pre00'
+        PreReleaseVersion = 'pre00'
     }
 
 }
--- a/src/posh-git.psd1
+++ b/src/posh-git.psd1
@@ -1,10 +1,10 @@
 @{
 
 # Script module or binary module file associated with this manifest.
-ModuleToProcess = 'posh-git.psm1'
+RootModule = 'posh-git.psm1'
 
 # Version number of this module.
-ModuleVersion = '0.7.0'
+ModuleVersion = '1.0.0'
 
 # ID used to uniquely identify this module
 GUID = '74c9fd30-734b-4c89-a8ae-7727ad21d1d5'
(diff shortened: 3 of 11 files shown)


chocolatey/poshgit.nuspec
```diff
@@ -3,7 +3,7 @@
   <metadata>
     <id>poshgit</id>
     <title>posh-git</title>
-    <version>0.7.0</version>
+    <version>1.0.0-pre00</version>
```
dahlbyk: 👍

src/posh-git.psd1
```diff
@@ -50,7 +49,7 @@ CmdletsToExport = @()
 VariablesToExport = @()
 
 # Aliases to export from this module
-AliasesToExport = @('??')
+AliasesToExport = @()
```
lzybkr: An empty array helps with overall system performance.

If you omit the entry, PowerShell does more expensive work to see if you meant to export any aliases.

appveyor.yml
```diff
@@ -3,7 +3,7 @@ os:
 
 branches:
   only:
-    - master
```
rkeithhill: Ideally there would be a way in appveyor.yml to specify the "current target branch".  So if we put in both `master ` and `develop` will that attempt to build a PR meant for develop on master also?  If so, that would likely fail.  Then again, I'm a total AppVeyor n00b.

src/posh-git.psd1
```diff
@@ -1,10 +1,10 @@
 @{
 
 # Script module or binary module file associated with this manifest.
-ModuleToProcess = 'posh-git.psm1'
+RootModule = 'posh-git.psm1'
 
 # Version number of this module.
-ModuleVersion = '0.7.0'
+ModuleVersion = '1.0.0'
```
rkeithhill: '1.0.0' comes out as version 1.0.0.-1 so the order between release and pre is wrong.  :-(
We can add a field to the module manifest private data to indicate a PrereleaseVersion string that we could retrieve with `(gmo posh-git).PrivateData.PSData.PrereleaseVersion`.

appveyor.yml
dahlbyk: It just means any change in `master` or `develop` will be built. We will probably want to add a badge for `develop`, too.

PRs are built regardless of target; this setting is (ideally) only for merged PRs.

src/posh-git.psd1
dahlbyk: I am mostly concerned with the issue template version script showing a difference between loading from `develop`/`master` and from Chocolatey or the Gallery.

An alternative would be to include the posh-git module's path, perhaps with a replacement of `$Home` with `~`?

dahlbyk: > '1.0.0' comes out as version 1.0.0.-1 so the order between release and pre is wrong.

It may actually be desirable for a local version (think fork with personalized features) to win out over an official version (e.g. inherited from GitHub Desktop). Maybe? Of course the personalized fork could just bump the version...

rkeithhill: The way I handle that scenario is to import by path e.g:
```powershell
Import-Module $Home\GitHub\dahlbyk\posh-git\src\posh-git.psd1
```
Development is in the works for PowerShellGet/PSGallery to support prereleases.  I'm hoping this problem will get solved in the coming months.  https://github.com/PowerShell/PowerShell-RFC/issues/42

dahlbyk: Yes, but doing that if posh-git has already been loaded, e.g. in the GitHub Desktop shell, leaves you with two posh-git modules loaded. If one of those is 0.7.0 and the other is a personal fork that also reports 0.7.0, I have to check path to differentiate. Thus my suggestion of the extra `.0` as a hint that this isn't an official version.

That said, also including `PrereleaseVersion` makes sense… add that to the issue template script?

rkeithhill: OK.  But I'll update the issue template file on master in a separate PR.  Also, regarding those extra commands to not export.  I'll remove those.  I'm guilty of adding Get-GitBranch for testing purposes but I think Pesters's InModuleScope will get me access to this function.

rkeithhill: What if we leave in `Write-Prompt` to abstract the difference between using Write-Host and ANSI escape sequences.  Folks customizing their own prompt function might find that useful.

dahlbyk:
So build is green here, but [failed on `develop`](https://ci.appveyor.com/project/dahlbyk/posh-git/build/0.7.237). @rkeithhill can you take a look?

rkeithhill:
Yup.  Could be the switch to using InModuleScope to access the Get-GitBranch function (as opposed to exporting it).

dahlbyk:
> Could be the switch to using InModuleScope to access the Get-GitBranch function (as opposed to exporting it).

I wouldn't expect that to have passed in the PR build. My guess is it's due to building on `develop` instead of `master`?

rkeithhill:
Was just looking more closely at the error messages. It doesn't like creating "#develop". I wouldn't think that would interfere with the existing `develop` branch.

rkeithhill:
I can run these tests locally on both master and develop and they all pass.  This must have something to do with how AppVeyor sets up the build/test environment. For instance, when it uses develop, I see failures like this:
```
        Expected exactly: {master}

        But was: {}
```
Is it fetching develop only and not master as well?

dahlbyk:
`Context 'Fetch/Push/Pull TabExpansion Tests'` doesn't use a test repo; it's using the posh-git repo. I get the same set of failures if I delete my local `master` branch. PR incoming.

rkeithhill:
Yeah, my bad.  Thanks for the help on this!

dahlbyk:
PR opened into `master`. I'll open another PR from `master` into `develop` when this is merged.

PR #428, opened 2017-02-18
Add-PoshGitToProfile should not modify signed prof

rkeithhill:
Forgot to address this issue before shipping 0.7.0 - https://github.com/dahlbyk/posh-git/issues/328#issuecomment-269740506

Merge to develop.

### DIFF
--- a/src/Utils.ps1
+++ b/src/Utils.ps1
@@ -142,7 +142,7 @@ function Add-PoshGitToProfile {
     }
 
     if (!$profilePath) {
-        Write-Warning "Skipping add of posh-git import; no profile found."
+        Write-Warning "Skipping add of posh-git import to profile; no profile found."
         Write-Verbose "`$profilePath          = '$profilePath'"
         Write-Verbose "`$PROFILE              = '$PROFILE'"
         Write-Verbose "CurrentUserCurrentHost = '$($PROFILE.CurrentUserCurrentHost)'"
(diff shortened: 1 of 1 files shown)


src/Utils.ps1
```diff
@@ -138,6 +152,16 @@ function Add-PoshGitToProfile([switch]$AllHosts, [switch]$Force, [switch]$WhatIf
         return
     }
 
+    # If the profile script exists and is signed, then we should not modify it
+    if (Test-Path -LiteralPath $profilePath) {
+        $sig = Get-AuthenticodeSignature $profilePath
+        if ($sig.Status -eq [System.Management.Automation.SignatureStatus]::Valid) {
```
dahlbyk: Should we warn/abort if `$sig.Status` is anything other than `NotSigned`?

src/Utils.ps1
```diff
@@ -82,18 +82,32 @@ function Invoke-Utf8ConsoleCommand([ScriptBlock]$cmd) {
 .OUTPUTS
     None.
 #>
-function Add-PoshGitToProfile([switch]$AllHosts, [switch]$Force, [switch]$WhatIf) {
+function Add-PoshGitToProfile {
+    [CmdletBinding(SupportsShouldProcess)]
```
dahlbyk: 👍 Always forget about this.

src/Utils.ps1
rkeithhill: Yup. I'll make that change.

rkeithhill: Well that turned out to be a little ugly as there are three values that can represent an unsigned file (NotValid, UnknownError, IncompatibleFileType) and four values that can represent a signed file.  Switching to testing the SignatureType field.

PR #441, opened 2017-02-23
Don't rerun pageant if there are no keys to add

somewhatabstract:
Currently, if there is no key file, posh-git's Start-SshAgent runs pageant twice which causes pageant to show an error dialog about already running. This simple change stops that happening.


dahlbyk:
Thanks!

PR #444, opened 2017-02-25
Hide profile.example.ps1 warning for Chocolatey

### LINKED ISSUE #442: Failed to import posh-git
### System Details
- posh-git version/path: 0.7.0  C:\tools\poshgit\dahlbyk-posh-git-7d93c81\src
- PowerShell version: 3.0
- git version 1.9.5.msysgit.1
- OS: Microsoft Windows NT 6.1.7601 Service Pack 1

### Issue Description

Amateur here who uses posh-git for hobby projects. posh-git's been working fine. On a whim, I updated the other day, now when I start my powershell I get a _yellow warning_ saying I should copy the contents of the profile.example.ps1 to the $PROFILE.  When I copy the contents of the profile.example.ps1 into my Microsoft.PowerShell_profile.ps1 file, I get the error message "Failed to import posh-git" and posh-git will not work. 
I thought I  might type in the commandlet to see what I get:
`get-module posh-git -ListAvailable`
But I get nothing.  If I restore my Powershell profile back so it points to the original profile.example.ps1, it again works fine. It's as if the import only works when it is in the original location.  Basically, git and posh-git are the only things I really use Powershell for at the present time.  Am I misunderstanding the instructions? What am I missing? 
[Microsoft.PowerShell_profile.txt](https://github.com/dahlbyk/posh-git/files/800091/Microsoft.PowerShell_profile.txt) <--my current non-working profile

dahlbyk:
Fixes #442 with two changes:

1. Hides the warning if posh-git is managed by Chocolatey, since we can largely control use of `profile.example.ps1` with the installer
2. Replaces the "copy its contents into your $PROFILE" suggestion with a better recommendation

#338 notwithstanding, one useful thing `profile.example.ps1` does is call `Start-SshAgent`, so I've also added `Add-PoshGitToProfile -StartSshAgent` to add the same.

### DIFF
--- a/chocolatey/tools/chocolateyInstall.ps1
+++ b/chocolatey/tools/chocolateyInstall.ps1
@@ -27,7 +27,7 @@
             if ($line -like '*PoshGitPrompt*') { continue; }
 
             if($line -like '. *posh-git*profile.example.ps1*') {
-                $line = ". '$currentVersionPath\profile.example.ps1'"
+                $line = ". '$currentVersionPath\profile.example.ps1' choco"
             }
             if($line -like 'Import-Module *\src\posh-git.psd1*') {
                 $line = "Import-Module '$currentVersionPath\src\posh-git.psd1'"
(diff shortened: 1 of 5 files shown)


dahlbyk:
> Do you think it is worth adding a pester test for the new parameter to Utils.Tests.ps1?

I do, and have.

Issue #345, opened 2016-12-31
Bust up Write-GitStatus into smaller functions to make it easier for individuals to override parts of the status display


rkeithhill:
I see a number of PRs come across to change the Git status summary display for things that probably don't apply to that many folks.  And folks can today write their own prompt function to take the information from `Get-GitStatus` to display the Git status summary, plus any additional information in anyway they would like.  However, in many cases folks just want to tweak one area of the Git status summary info.  With this in mind, we should break up the bits and pieces of `Write-GitStatus` into reusable functions.  

Perhaps we have `Write-GitStatus` call down to `Write-BranchStatus`, `Write-GitIndexStatus`, and `Write-GitWorkingDirStatus`. 

In one PR scenario the OP wants to display any tags associated with HEAD.  In this case, the OP could create his own version of Write-BranchStatus to display tags. Then in his prompt function he could call `Write-GitIndexStatus` and `Write-GitWorkingDirStatus` to create the rest of the status summary text. This would make overriding an individual status summary section much easier.

You could make the functions even finer grained as well. Perhaps break down `Write-BranchStatus` into `Write-BranchName` and `Write-BranchSyncStatus`?

dahlbyk:
This is also related to #304 and https://github.com/Jaykul/PowerLine/issues/1. More modularity and customizability in how the prompt displays would be **awesome**. And if we do it right, it seems like an opportunity to remove a bunch of clutter from `$GitPromptSettings`, e.g. default might be:

```powershell
function Write-GitStatus ($GitStatus) {
  Write-Prompt ' [' -ForegroundColor Yellow # No need for Before settings
  Write-GitBranchStatus $GitStatus
  Write-GitFileStatus $GitStatus
  Write-GitLocalStatus $GitStatus
  Write-Prompt ']' -ForegroundColor Yellow # No need for After settings
}
```

And to customize:

```powershell
function Write-GitStatus ($GitStatus) {
  Write-Prompt ' {'
  Write-GitBranchStatus $GitStatus -AheadBehindDisplay Compact -GoneText ''
  # No need for EnableFileStatus; just don't call Write-GitFileStatus
  Write-GitLocalStatus -StagedText ''
  Write-GitStashStatus -Before ' (' -After '(' -ForegroundColor Red
  Write-Prompt '}'
}
```

rkeithhill:
[...]

Issue #340, opened 2016-12-31
Export function to manipulate GitPromptSettings


dahlbyk:
Spawned from https://github.com/dahlbyk/posh-git/pull/256#discussion_r94244778, similar to #111.

Pros:
- Discoverability (module definition, etc)
- Tab completion
- Parameter Validation (e.g. `BranchBehindAndAheadDisplay` "enum")

Cons:
- More code to maintain
- Migration path from `$GitPromptSettings`? Or does function just manipulate the global?

rkeithhill:
Pertinent to this discussion (and the discussion on module install script) - https://github.com/PoshCode/PowerShellPracticeAndStyle/issues/79

rkeithhill:
You get tab completion/Intellisense with the $GitPromptSettings variable so that really isn't a "pro" for a function.

rkeithhill:
For 1.0.0 we're going in the direction of a more structured, strongly typed GitPromptSettings object.  If you disagree or think this needs to be reopened, please reopen it.

PR #447, opened 2017-03-01
Use ANSI color codes if SupportsVirtualTerminal

### LINKED ISSUE #282: posh-git prompt makes PowerShell transcripts less readable

dahlbyk:
See https://github.com/dahlbyk/posh-git/pull/304 for detail and much conversation. For now I have just rebased on `develop`, resolved all the conflicts, and updated the new default `prompt`.

Closes #282 and Jaykul/PowerLine#1.

### DIFF
--- a/src/AnsiUtils.ps1
+++ b/src/AnsiUtils.ps1
@@ -1,0 +1,36 @@
+# Color codes from https://msdn.microsoft.com/en-us/library/windows/desktop/mt638032(v=vs.85).aspx
+$ConsoleColorToAnsi = @(
+    30 # Black
+    34 # DarkBlue
+    32 # DarkGreen
+    36 # DarkCyan
+    31 # DarkRed
+    35 # DarkMagenta
+    33 # DarkYellow
+    37 # Gray
+    90 # DarkGray
+    94 # Blue
+    92 # Green
+    96 # Cyan
+    91 # Red
+    95 # Magenta
+    93 # Yellow
+    97 # White
+)
+$AnsiDefaultColor = 39
+$AnsiEscape = [char]27 + "["
+
+function Get-VirtualTerminalSequence ($color, [int]$offset = 0) {
+    if (($color -is [ConsoleColor]) -and ($color -ge 0) -and ($color -le 15)) {
+        return "${AnsiEscape}$($ConsoleColorToAnsi[$color] + $offset)m"
+    }
+    return "${AnsiEscape}$($AnsiDefaultColor + $offset)m"
+}
+
+function Get-ForegroundVirtualTerminalSequence($Color) {
+    return Get-VirtualTerminalSequence $Color
+}
+
+function Get-BackgroundVirtualTerminalSequence($Color) {
+    return Get-VirtualTerminalSequence $Color 10
+}
--- a/src/ConsoleMode.ps1
+++ b/src/ConsoleMode.ps1
@@ -1,0 +1,121 @@
+# Hack! https://gist.github.com/lzybkr/f2059cb2ee8d0c13c65ab933b75e998c
+
+Add-Type @"
+using System;
+using System.Runtime.InteropServices;
+
+public class NativeConsoleMethods
+{
+    [DllImport("kernel32.dll", CharSet = CharSet.Unicode, SetLastError = true)]
+    public static extern IntPtr GetStdHandle(int handleId);
+
+    [DllImport("kernel32.dll", CharSet = CharSet.Unicode, SetLastError = true)]
+    public static extern bool GetConsoleMode(IntPtr hConsoleOutput, out uint dwMode);
+
+    [DllImport("kernel32.dll", CharSet = CharSet.Unicode, SetLastError = true)]
+    public static extern bool SetConsoleMode(IntPtr hConsoleOutput, uint dwMode);
+
+    public static uint GetConsoleMode(bool input = false)
+    {
+        var handle = GetStdHandle(input ? -10 : -11);
+        uint mode;
+        if (GetConsoleMode(handle, out mode))
+        {
+            return mode;
+        }
+        return 0xffffffff;
+    }
+
+    public static uint SetConsoleMode(bool input, uint mode)
+    {
+        var handle = GetStdHandle(input ? -10 : -11);
+        if (SetConsoleMode(handle, mode))
+        {
+            return GetConsoleMode(input);
+        }
+        return 0xffffffff;
+    }
+}
+"@
+
+[Flags()]
+enum ConsoleModeInputFlags
+{
+    ENABLE_PROCESSED_INPUT             = 0x0001
+    ENABLE_LINE_INPUT                  = 0x0002
+    ENABLE_ECHO_INPUT                  = 0x0004
+    ENABLE_WINDOW_INPUT                = 0x0008
+    ENABLE_MOUSE_INPUT                 = 0x0010
+    ENABLE_INSERT_MODE                 = 0x0020
+    ENABLE_QUICK_EDIT_MODE             = 0x0040
+    ENABLE_EXTENDED_FLAGS              = 0x0080
+    ENABLE_AUTO_POSITION               = 0x0100
+    ENABLE_VIRTUAL_TERMINAL_PROCESSING = 0x0200
+}
+
+[Flags()]
+enum ConsoleModeOutputFlags
+{
+    ENABLE_PROCESSED_OUTPUT            = 0x0001
+    ENABLE_WRAP_AT_EOL_OUTPUT          = 0x0002
+    ENABLE_VIRTUAL_TERMINAL_PROCESSING = 0x0004
+}
+
+function Get-ConsoleMode
+{
+    [CmdletBinding()]
+    param(
+        [switch]
+        $StandardInput
+    )
+
+    $mode = [NativeConsoleMethods]::GetConsoleMode($StandardInput)
+    if ($StandardInput)
+    {
+        [ConsoleModeInputFlags]$mode
+    }
+    else
+    {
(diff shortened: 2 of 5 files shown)


dahlbyk:
> Collect Git prompt in ArrayList instead of strings

Not sure if this is a good optimization or not. Thoughts, @lzybkr?

-----------------------------

More generally, I wonder if this is Close Enough™ for now to merge, giving basic support and a fixed point from which to iterate on how exactly ANSI colors work. In addition to the actual ANSI implementation (dedicated module?), colors play a key role in prompt customization (cc #340 #345).

src/ConsoleMode.ps1
```diff
@@ -0,0 +1,103 @@
+# Hack! https://gist.github.com/lzybkr/f2059cb2ee8d0c13c65ab933b75e998c
+
+if ($IsWindows -eq $false) {
```
rkeithhill: For non-cross-platform PowerShell Core, we should test $PSVersionTable.PSVersion.Major -ge 6 -and $IsWindows.  As that variable is only defined in PowerShell Core.  We can assume a major version of -lt 6 is Windows only.

src/GitPrompt.ps1
```diff
@@ -97,6 +97,8 @@ $global:GitPromptSettings = [pscustomobject]@{
 
     EnableWindowTitle                           = 'posh~git ~ '
 
+    AnsiConsole                                 = $Host.UI.SupportsVirtualTerminal -or ($Env:ConEmuANSI -eq "ON")
```
rkeithhill: I believe the SupportsVirtualTerminal property was added in PS 5.1.  I have started a prototype for ANSI stuff based on v5 classes. In it, I have a PromptRenderer class that has a HostSupportsAnsi method.  That method returns $false for version less than 5.1.  It returns false for Windows version less than 10.0.10586.0 (which is the build that support conhost ANSI seqs).  Finally, it returns the value of $Host.UI.SupportsVirtualTerminal.

src/GitPrompt.ps1
```diff
@@ -128,11 +130,18 @@ if (Get-Module NuGet) {
 }
 
 function Write-Prompt($Object, $ForegroundColor, $BackgroundColor = -1) {
+    if ($GitPromptSettings.AnsiConsole) {
```
rkeithhill: So in the PromptRenderer class I was working on, the idea was that it maintained an internal StringBuilder (for ANSI hosts) and it uses Write-Host for non-ANSI and the StringBuilder for ANSI.  Currently it takes an array of TextSpan objs (text, fg, bg in a class) but I was thinking it might be better to have a BeginWrite/Write/EndWrite set of methods that each individually take a TextSpan object.

src/GitPrompt.ps1
```diff
@@ -128,11 +130,18 @@ if (Get-Module NuGet) {
...
+        $e = [char]27 + "["
+        $f = Get-ForegroundVirtualTerminalSequence $ForegroundColor
+        $b = Get-BackgroundVirtualTerminalSequence $BackgroundColor
+        return "${f}${b}${Object}${e}0m"
```
rkeithhill: We need to be able to handle terms that support 24-bit color like the upcoming Windows 10 CU drop.  See my Twitter post on that - https://twitter.com/r_keith_hill/statuses/835556870697930753  This was something I was encapsulating in a Color class in my prototype.

rkeithhill:
Give me a day or so and I can commit my (very) prototype branch that uses v5 classes.  It is missing stuff you have like the setting the console mode and conversion from console colors to ANSI seq indices.  Maybe we can meld the two?

src/ConsoleMode.ps1
dahlbyk: Thought is that `$IsWindows` will be `$null` for PS `-lt 6`, and `$null -ne $false`. Net result is that we'll only create `NativeConsoleMethods` if we're not explicitly on not-Windows.

Or are you thinking that we shouldn't create `NativeConsoleMethods` for older versions of PS?

src/GitPrompt.ps1
dahlbyk: I'm not sure there's a practical difference, is there? If `SupportsVirtualTerminal` isn't set, it's falsey.

src/GitPrompt.ps1
dahlbyk: `$ForegroundColor` and `$BackgroundColor` are explicitly untyped, so in theory there's no reason `Get-ForegroundVirtualTerminalSequence` and `Get-BackgroundVirtualTerminalSequence` couldn't emit valid escape sequences given a variety of input types (`"#rrggbb"`, [`System.Drawing.Color`](https://msdn.microsoft.com/en-us/library/system.drawing.color.aspx), etc).

(Not that a `Color` class couldn't have equivalent support.)

src/ConsoleMode.ps1
rkeithhill: I guess I find the logic a bit cryptic.   Maybe this is worth a helper function IsWindows to encapsulate the logic.  RE `NativeConsoleMethods` no need to create these if they are not used and we won't use them except on Windows w/ANSI support, right?

src/GitPrompt.ps1
rkeithhill: You're probably right.  I carry some baggage having written scripts back in the early PS days that were affected by users setting `Set-StrictMode -Version Latest` in their global session.  Fortunately that does not impact modules.  That and I still cling to statically typed notions and referring to a property that may or may not exist just feels wrong.  ;-)

dahlbyk: > That and I still cling to statically typed notions and referring to a property that may or may not exist just feels wrong. ;-)

Understood. I'll try to help you get over it. 😀

src/ConsoleMode.ps1
dahlbyk: I wonder if the native console methods have an impact on ConEmu and other console hosts? Do they even exist on Windows pre-10?

lzybkr:
I'm not sure about `ArrayList` w/ `-join` versus `+=` on a string - my intuition says it won't matter - the strings are short anyway.

In general, use of `StringBuilder` would be a better optimization because you'll allocate fewer objects.

dahlbyk:
A quick benchmark confirms that it really doesn't matter for small strings, but for larger strings `StringBuilder` is an order of magnitude faster than `ArrayList`, which is maybe 25% faster than string concatenation.

We may end up wanting a `StringBuilder`, but I'm content to leave it for now.

> Give me a day or so and I can commit my (very) prototype branch that uses v5 classes. It is missing stuff you have like the setting the console mode and conversion from console colors to ANSI seq indices. Maybe we can meld the two?

I guess my question is if this branch is a reasonable enough starting point for people to 1) test it out, if they want, and 2) serve as a base for iterating on how we handle ANSI (your `PromptRenderer` or otherwise). Thoughts?

src/Utils.ps1
```diff
@@ -283,6 +283,43 @@ function Test-PoshGitImportedInScript {
...
+
+function Get-ForegroundVirtualTerminalSequence($Color) {
+    return Get-VirtualTerminalSequence $Color
+}
+
+function Get-BackgroundVirtualTerminalSequence($Color) {
+    return Get-VirtualTerminalSequence $Color 10
+}
+
 function dbg($Message, [Diagnostics.Stopwatch]$Stopwatch) {
```
rkeithhill: Eventually we might to put these into an Ansi.ps1 or AnsiUtils.ps1 file but that can wait.

rkeithhill:
> I guess my question is if this branch is a reasonable enough starting point 

Good point.  Go ahead and merge it. I have an Ubuntu VM I work on everyday for embedded Linux development work. I could take it for a spin there.

dahlbyk:
Just rewrote the world:

- ANSI stuff moved to separate file
- Squashed a few fixup-y commits
- Added a spike using `StringBuilder` just for fun... `Write-Prompt` semantics are now even more odd

src/AnsiUtils.ps1
```diff
@@ -0,0 +1,36 @@
+# Color codes from https://msdn.microsoft.com/en-us/library/windows/desktop/mt638032(v=vs.85).aspx
```
rkeithhill: Hehe, that is exactly the filename I'm using in my classes prototype branch.  Did some work on that last night but not quite ready yet for evaluation.

src/ConsoleMode.ps1
rkeithhill: What was the initial impetus for this?  PowerShell sets the console mode to enable ANSI seqs. Or was there a version (5.0) that didn't do this?

lzybkr: The api exists on all versions of Windows - but the flag we need is not recognized universally - just Win10 Anniversary and later I think.  The api call will fail (I think harmlessly) if it doesn't recognize the flag.

The workaround should not be needed in V6 - I fixed it in PowerShell: https://github.com/PowerShell/PowerShell/pull/2991

PR #449, opened 2017-03-02
During install create the profile dir if it does not already exist

tolgabalci:
While installing posh-git using install.ps1 on Windows 7 Enterprise SP1 laptops for users running Powershell 4.0 or 5.0 had some users receive this error:

    PS C:\tools\posh-git> .\install.ps1
    Add-Content : Could not find a part of the path
    'C:\Users\*username*\Documents\WindowsPowerShell\Microsoft.PowerShell_profile.ps1'.
    At C:\tools\posh-git\src\Utils.ps1:180 char:9
    +         Add-Content -LiteralPath $profilePath -Value $profileContent -Encoding U ...

Nothing had previously been added to the user's Profile scripts AND the folder "WindowsPowerShell" did not exists under the user's documents.  This caused Add-Content to fail.

The pull request is to create the Powershell profile directory for the current user if it does not already exists.

Tested both with and without -WhatIf and it works good.

### DIFF
--- a/src/Utils.ps1
+++ b/src/Utils.ps1
@@ -176,6 +176,14 @@ function Add-PoshGitToProfile {
         $profileContent = "`nImport-Module '$ModuleBasePath\posh-git.psd1'"
     }
 
+    # Make sure the PowerShell profile directory exists
+    $profileDir = Split-Path $profilePath -Parent
+    if (!(Test-Path -LiteralPath $profileDir)) { 
+        if ($PSCmdlet.ShouldProcess($profileDir, "Create current user PowerShell profile directory")) {
+            New-Item $profileDir -ItemType Directory -Verbose:$VerbosePreference > $null
+        }
+    }
+
     if ($PSCmdlet.ShouldProcess($profilePath, "Add 'Import-Module posh-git' to profile")) {
         Add-Content -LiteralPath $profilePath -Value $profileContent -Encoding UTF8
     }


src/Utils.ps1
```diff
@@ -176,6 +176,14 @@ function Add-PoshGitToProfile {
         $profileContent = "`nImport-Module '$ModuleBasePath\posh-git.psd1'"
     }
 
+    # Make sure the profile path exists
+    $profileDir = Split-Path -Parent $profilePath
+    if (-not (Test-Path -LiteralPath $profileDir)) {
```
rkeithhill: In the code base we tend not to use `-not` in favor of `!`.  We're not 100% there yet but I don't want to introduce more script that needs to be changed.

src/Utils.ps1
```diff
@@ -176,6 +176,14 @@ function Add-PoshGitToProfile {
         $profileContent = "`nImport-Module '$ModuleBasePath\posh-git.psd1'"
     }
 
+    # Make sure the profile path exists
+    $profileDir = Split-Path -Parent $profilePath
+    if (-not (Test-Path -LiteralPath $profileDir)) { 
+        if ($PSCmdlet.ShouldProcess($profileDir, "Create current user Powershell profile directory")) {
+            New-Item $profileDir -Type Directory
```
rkeithhill: Two things.  Better to use the actual parameter name `ItemType` than the alias `Type`.  Also, this command will output the created directory.  Better to just redirect the output to $null.  Then add support for -verbose e.g.:
```powershell
New-Item $profileDir -ItemType Directory -Verbose:$VerbosePreference > $null
```

src/Utils.ps1
```diff
@@ -176,6 +176,14 @@ function Add-PoshGitToProfile {
         $profileContent = "`nImport-Module '$ModuleBasePath\posh-git.psd1'"
     }
 
+    # Make sure the profile path exists
+    $profileDir = Split-Path -Parent $profilePath
+    if (-not (Test-Path -LiteralPath $profileDir)) { 
+        if ($PSCmdlet.ShouldProcess($profileDir, "Create current user Powershell profile directory")) {
```
rkeithhill: PowerShell is properly spelled with a capital `S` - `PowerShell`.

tolgabalci:
Great callouts!  Thanks.

Amended the suggested changes and retested. (Both with $VerbosePreference = "SilentlyContinue" and "Continue").  Tests looked good.

src/Utils.ps1
```diff
@@ -176,6 +176,14 @@ function Add-PoshGitToProfile {
         $profileContent = "`nImport-Module '$ModuleBasePath\posh-git.psd1'"
     }
 
+    # Make sure the profile path exists
+    $profileDir = Split-Path -Parent $profilePath
```
rkeithhill: Sorry missed one other minor issue (which I myself forget from time-to-time).  Can you switch the order on params here and use `-LiteralPath` e.g.` $profileDir = Split-Path -LiteralPath $profilePath -Parent`.  Thanks.

rkeithhill:
You should also be able to call `Add-PoshGitToProfile -Verbose` and see the verbose output from New-Item.  BTW thanks for the PR!  See the one other issue I found - it's a quick tweak.

src/Utils.ps1
tolgabalci: No problem, I moved the -Parent parameter to the end.

I also looked at using -LiteralPath with Split-Path, but it does not have an explicit -LiteralPath parameter, however based on the below test it looks like it already implicitly treats the path literally.

```
C:\test[1]> Get-Item -LiteralPath c:\test[1]\File[1].txt
    Directory: C:\test[1]
-a----         3/2/2017   3:18 AM              0 File[1].txt

C:\test[1]> Get-Item -Path c:\test[1]\File[1].txt

C:\test[1]> Split-Path -Path c:\test[1]\File[1].txt -Parent
c:\test[1]
```

I created a directory and file with names that both need to be interpreted literally.

- Get-Item with -LiteralPath works fine

- Get-Item with -Path fails to retrieve the file

- Split-Path with -Path however still correctly parses the path to retrieve the parent

So, I think Split-Path is good with just -Path.  Looking through the other instances of Split-Path in the code, I figured I would keep it consistent and not explicitly state -Path, but let me know if you want me to change it.  It's no problem.

dahlbyk:
LGTM, but care to add a [test](https://github.com/tolgabalci/posh-git/blob/c2a92dda7d11097558f62681de0a1a3989854ba3/test/Utils.Tests.ps1#L5) for this scenario?

rkeithhill:
@tolgabalci OK for the ParentSet parameter set, there is no -LiteralPath parameter. Looks good. I'm going to merge this and I'll add a test later tonight.  If you're really inspired to add the test, you can always open a separate PR.

dahlbyk:
It just dawned on me that, there's a much simpler fix for this: `New-Item -Force` will automatically create directories as necessary. This is what [`install.ps1` used to do](https://github.com/dahlbyk/posh-git/blob/b1c9ad4f7b504e3480daf3fce4b3b3a60702c983/install.ps1#L14).

rkeithhill:
Do you think it is worth changing this?  I'm not sure that it is.

dahlbyk:
Just a little less path-splitting logic to maintain. And thinking through edge cases, it's technically possible (though unlikely with defaults) that more than `$PROFILE`'s immediate parent could be missing. Of course, you could work around this with `-Force` when creating the parent. ¯\_(ツ)_/¯

tolgabalci:
@rkeithhill Thank you for adding the test and showing me how to do it!

@dahlbyk New-Item will actually create the full parent directory tree without -Force.  I do not think -Force is really intended for use when creating directories.  I was trying it out, it does have the interesting side effect of suppressing errors when one of the directories in the directory tree being created is a file and it fails to create the path.

@rkeithhill @dahlbyk, thank you so much for being so responsive!

dahlbyk:
> New-Item will actually create the full parent directory tree without -Force when used with -ItemType Directory. I do not think -Force is really intended for use when creating directories. I was trying it out, it does have the interesting side effect of suppressing errors when one of the directories in the directory tree being created is a file and it fails to create the path.

Interesting… I suppose the nested directory variations are all things we could test for, if you are interested in exploring publicly. 😁

PR #451, opened 2017-03-02
Verbose install parameter

tolgabalci:
When troubleshooting install issues, it would be nice to be able to call:
`.\install.ps1 -Verbose`

Alternative for seeing verbose output from install.ps1 is:
```
$VerbosePreference = "Continue"
.\install.ps1
```

Another option would be to add SupportsShouldProcess binding, but then Import-Module also displays verbose output and it is pretty chatty.  I am assuming that is why WhatIf was also added separately.

I wasn't sure though the order you would like the Verbose parameter declared in and then later used when calling Add-PoshGitToProfile.  Let me know if you would like for me to change the orders.

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -1,6 +1,6 @@
-param([switch]$WhatIf = $false, [switch]$Force = $false)
+param([switch]$WhatIf = $false, [switch]$Force = $false, [switch]$Verbose = $false)
 
 $installDir = Split-Path $MyInvocation.MyCommand.Path -Parent
 
 Import-Module $installDir\src\posh-git.psd1
-Add-PoshGitToProfile -WhatIf:$WhatIf -Force:$Force
+Add-PoshGitToProfile -WhatIf:$WhatIf -Force:$Force -Verbose:$Verbose


dahlbyk:
Thanks for the contribution!

> Another option would be to add SupportsShouldProcess binding, but then Import-Module also displays verbose output and it is pretty chatty. I am assuming that is why WhatIf was also added separately.

Honestly, `SupportsShouldProcess` isn't used because I didn't know about it at the time that I added `-WhatIf` [_six years ago_](https://github.com/dahlbyk/posh-git/commit/e1c65510c3c0035ae25f189dc29b6c6d505eca5b).

That said, it doesn't look like `Import-Module` actually `SupportsShouldProcess`. It does support `-Verbose`, and I was going to suggest passing your new `$Verbose` flag through to it - yeah it's chatty, but that's kind of the point of `-Verbose`, no?

> I wasn't sure though the order you would like the Verbose parameter declared in and then later used when calling Add-PoshGitToProfile. Let me know if you would like for me to change the orders.

I don't care about the order, but I do wonder if `install.ps1` should just use `[CmdletBinding()]` so we don't have to explicitly define or pass along `-WhatIf` or `-Verbose`. Thoughts?

PR #452, opened 2017-03-03
Add test for update to Add-PoshGitToProfile

rkeithhill:
This test checks that the command creates the profile parent dir (or any in the path) if they don't exist.  The command was modified to create any dir in the path by the addition of the -Force parameter on New-Item.

Follow-up on #449

### DIFF
--- a/src/Utils.ps1
+++ b/src/Utils.ps1
@@ -178,9 +178,9 @@ function Add-PoshGitToProfile {
 
     # Make sure the PowerShell profile directory exists
     $profileDir = Split-Path $profilePath -Parent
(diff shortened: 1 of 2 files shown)


dahlbyk:
> Simplify missing profile parent test

For your consideration... since we already have cleanup infrastructure (for the record `Remove-Item -Recurse` works fine on file paths), the test seems cleaner without the `try...finally`.

rkeithhill:
Nice!  I got hung up $profilePath being a generated filename (not a dir) but yeah, deleting the file and usurping that name for the dirname does make this cleaner.  Thanks!

PR #455, opened 2017-03-04
ANSI 256-color and 24-bit color support

dahlbyk:
This PR adds support for colors beyond `[ConsoleColor]`:

* `[byte]` color values are processed as as [Xterm 256 colors](http://conemu.github.io/en/AnsiEscapeCodes.html)
* [`[System.Drawing.Color]`](https://msdn.microsoft.com/en-us/library/system.drawing.color.aspx) color values are processed as 24-bit colors
* `[string]` values are [parsed as HTML color values](https://msdn.microsoft.com/en-us/library/system.drawing.colortranslator.fromhtml.aspx) and processed as 24-bit colors

Includes https://github.com/dahlbyk/posh-git/pull/454 for adjusting `Write-Prompt` semantics, updated for 1.0 changes.

24-bit support currently depends on System.Drawing. Not sure what Core options are there, but I've tried to make the assembly/type loading resilient.

### Usage

```
$GitPromptSettings.DefaultForegroundColor = [ConsoleColor]::Blue
$GitPromptSettings.DefaultForegroundColor = 'Aquamarine'
$GitPromptSettings.DefaultForegroundColor = [Drawing.Color]::Aqua
$GitPromptSettings.DefaultForegroundColor = '#bbb'
$GitPromptSettings.DefaultForegroundColor = [byte]0xf0
$GitPromptSettings.DefaultForegroundColor = $null
```

I have tried testing with ConEmu, but 256-color mode doesn't seem to be working? Or is that a limitation of PS v5?
conemu64_2017-03-04_11-28-00

### DIFF
--- a/src/AnsiUtils.ps1
+++ b/src/AnsiUtils.ps1
@@ -25,15 +25,17 @@ $ColorTranslatorType = 'System.Drawing.ColorTranslator' -as [Type]
 $ColorType = 'System.Drawing.Color' -as [Type]
 
 function Get-VirtualTerminalSequence ($color, [int]$offset = 0) {
-    if (($color -is [ConsoleColor]) -and ($color -ge 0) -and ($color -le 15)) {
-        return "${AnsiEscape}$($ConsoleColorToAnsi[$color] + $offset)m"
-    }
     if ($color -is [byte]) {
         return "${AnsiEscape}$(38 + $offset);5;${color}m"
     }
-    if ($ColorTranslatorType -and ($color -is [String])) {
+    if ($color -is [String]) {
         try {
-            $color = $ColorTranslatorType::FromHtml($color)
+            if ($ColorTranslatorType) {
+                $color = $ColorTranslatorType::FromHtml($color)
+            }
+            else {
+                $color = [ConsoleColor]$color
+            }
         }
         catch {
             Write-Debug $_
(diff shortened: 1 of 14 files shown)


dahlbyk:
Interesting... so a question maybe for @lzybkr: is it expected that the `$Host.UI.RawUI` color properties are mutated by `"<ANSI color string>" | Out-Host` but not `"<ANSI color string>" | Write-Host"`? Since #447, using ANSI in the prompt means `$Host.UI.RawUI` colors can't be adjusted because they're always reset by `ESC[0m` in the `prompt` result. Are `RawUI.ForegroundColor`/`BackgroundColor` not meant to be manipulated? How else might one change, let's say, the default colors from the (very peculiar) defaults of `DarkYellow`/`DarkMagenta`?

image

src/AnsiUtils.ps1
```diff
@@ -20,10 +20,28 @@ $ConsoleColorToAnsi = @(
 $AnsiDefaultColor = 39
 $AnsiEscape = [char]27 + "["
 
+[Reflection.Assembly]::LoadWithPartialName('System.Drawing') > $null
+$ColorTranslatorType = ([System.Management.Automation.PSTypeName]'System.Drawing.ColorTranslator').Type
```
lzybkr: This should be
```PowerShell
$ColorTranslatorType = 'System.Drawing.ColorTranslator' -as [Type]
```

lzybkr:
Regarding `ESC[0m` - I feel like there is a bug in conhost in not honoring the current foreground/background colors - but it depends on how you set those colors.

If you set fore/background colors via escape sequences, then I believe `ESC[0m` is working as expected - not necessarily the way I think it should, but it's compatible with *nix systems.

If you set fore/background colors via a specific Win32 api, I think `ESC[0m` should reset to those colors, not the initial colors from startup.

Ideally ANSI would have push/pop semantics, but that doesn't exist unfortunately.

dahlbyk:
@lzybkr so from posh-git's standpoint, then, would you agree that it is most correct for us to continue using `ESC[0m` to terminate our colorized strings?

I guess what confuses me most is the distinction between `Write-Host` (doesn't modify `RawUI`) and `Out-Host` (does). Apparently the `prompt` string is using the latter?

dahlbyk:
Merged so I can more easily bring `master`'s `Write-Prompt` changes into `develop`.

PR #461, opened 2017-03-15
Closes #416 - Add code coverage to Coveralls.io

JanJoris:
The API key used in appveyor.yml is still mine. It needs to be replaced before going to master. Create a new integration for posh-git at Coveralls.io and encrypt the project token using the secure variable page on [Appveyor](https://ci.appveyor.com/tools/encrypt).

PR #462, opened 2017-03-16
Fix Coveralls build?

dahlbyk:
Wondering if #461 is failing because it's building from @JanJoris fork? Seems unlikely, but it's unclear why else it would apparently not be decrypting my key.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,6 +1,10 @@
 os:
   - "WMF 5"
 
+environment:
+  CA_KEY:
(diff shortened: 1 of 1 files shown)


dahlbyk:
As Appveyor's tweeter so helpfully [pointed out](https://twitter.com/appveyor/status/842206544074747904) (and the [docs are quite clear](https://www.appveyor.com/docs/build-configuration/#secure-variables)), secure variables aren't available in PR builds.

PR #468, opened 2017-04-17
Fix 'Write-Prompt' to be able use Black as foreground color.

vladimir-poleh:
Current implementation replaces 'Black' color with DefaultForegroundColor. This breaks some themes from the oh-my-posh package.

### DIFF
--- a/src/GitPrompt.ps1
+++ b/src/GitPrompt.ps1
@@ -132,7 +132,7 @@ if (Get-Module NuGet) {
 
 function Write-Prompt($Object, $ForegroundColor = $null, $BackgroundColor = $null) {
     $s = $global:GitPromptSettings
-    if ($s -and !$ForegroundColor) {
+    if ($s -and $ForegroundColor -eq $null) {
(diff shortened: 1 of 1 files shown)


src/GitPrompt.ps1
```diff
@@ -132,7 +132,7 @@ if (Get-Module NuGet) {
 
 function Write-Prompt($Object, $ForegroundColor = $null, $BackgroundColor = $null) {
     $s = $global:GitPromptSettings
-    if ($s -and !$ForegroundColor) {
+    if ($s -and $ForegroundColor -eq $null) {
```
rkeithhill: Good catch!  To eliminate a PSScriptAnalyzer warning in VSCode, would you mind changing that ever so slightly to:
```posh
if ($s -and ($null -eq $ForegroundColor)) { 
```

vladimir-poleh: Thanks for the info about PSScriptAnalyzer.

dahlbyk:
Thanks!

PR #479, opened 2017-07-10
Specify CommandType as Application

mikesigs:
Related to #478. This is a better fix. By specifying the CommandType as Application we can continue looking for `git` (exe or cmd) but avoiding any potential aliases the user may have created for "git".


rkeithhill:
You beat me to it.  :-)

PR #484, opened 2017-09-14
squash ssh agent warnings if `-Quiet`

refack:
Maps better to stdout/stderr, so cases where PS is run as CLI tool
Refs: https://github.com/nodejs/node-gyp/issues/1195#issuecomment-329574067

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -15,7 +15,7 @@ init:
 install:
   - ps: |
       Install-PackageProvider -Name NuGet -MinimumVersion 2.8.5.201 -Force | Out-Null
-      Install-Module Pester -MinimumVersion 3.4.0 -Scope CurrentUser -Force | Out-Null
+      Install-Module Pester -MinimumVersion 3.4.0 -MaximumVersion 3.99.99 -Scope CurrentUser -Force | Out-Null
       "Git version: $(git.exe --version)"
       "PSVersion:   $($PSVersionTable.PSVersion), build: $($PSVersionTable.BuildVersion), clr version: $($PSVersionTable.ClrVersion)"
       "Host name:   $($Host.Name)"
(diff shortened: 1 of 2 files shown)


dahlbyk:
Hrm... most of these warnings aren't meant to be fatal (e.g. the posh-git default profile tries to start `ssh-agent`, but it's not a huge deal if it can't).

I do feel like `-NoProfile`, as you've proposed in https://github.com/nodejs/node-gyp/pull/1292, is the more correct solution to addressing these warnings. This module really isn't meant to be loaded non-interactively.

refack:
> I do feel like -NoProfile, as you've proposed in nodejs/node-gyp#1292, is the more correct solution to addressing these warnings. This module really isn't meant to be loaded non-interactively.

Gray area, agreed.
Only issue is that `node-gyp` has a long-ish release cycle (first need to land PR, then cut a new version, then get `npm` to adopt it, then get people to update `npm`), and I assume `posh-git` is more agile.

P.S. Looks like a nice tool, I'll give it a try.

dahlbyk:
> Only issue is that `node-gyp` has a long-ish release cycle (first need to land PR, then cut a new version, then get `npm` to adopt it, then get people to update `npm`), and I assume `posh-git` is more agile.

It is, but we also don't have an auto-update mechanism, so who knows how long until folks think to upgade?

That said, I'm quite confident this is only going to be hit by users that don't care about `ssh-agent` who are hitting `Start-SshAgent -Quiet` in the [default profile](). What if we only changed that function to suppress the two warnings altogether if `-Quiet`?

refack:
> What if we only changed that function to suppress the two warnings altogether if -Quiet?

Sure.
(I did the editing in the GitHub GUI, let me clone and force push)

refack:
Done

dahlbyk:
Close, but you updated `Add-SshAgent`, not `Start-SshAgent`!

dahlbyk:
Actually, `Start-SshAgent` calls `Add-SshAgent` so we should probably keep the `$Quiet` guard in both. Sorry for churn on this. 😞

dahlbyk:
@refack I've wasted enough of your time... made the rest of the changes, if you care to review?

refack:
> @refack I've wasted enough of your time... made the rest of the changes, if you care to review?

No time wasted... Cooperation is always a win 🏆

PR #487, opened 2017-09-28
Get-AuthenticodeSignature not on PS Core

rkeithhill:
Check if command available before trying to execute it.


rkeithhill:
To clarify, Get-AuthenticodeSignature is not on Linux (and probably not on MacOS) but it is available on PS Core when it's running on Windows.

dahlbyk:
`rebase`d onto latest `develop` to fix build.

PR #499, opened 2017-10-29
WIP: Reorgs settings using PS v5 classes

rkeithhill:
This is based on the existing `develop` branch approach to rendering ANSI.  The other approaches I tried aren't there yet. Add-Type is too slow on PS Core.  Using C# source code is a pain to get built and working cross-platform.

This approach uses v5 classes so we would have to drop support for v3/v4.  This seems reasonable to me as the primary points of this branch are to support cross-platform PowerShell Core and ANSI.

Before this change, looking at the `$GitPromptSettings` results in:

image

After this change:

image

### DIFF
--- a/src/AnsiUtils.ps1
+++ b/src/AnsiUtils.ps1
@@ -24,10 +24,31 @@ $AnsiEscape = [char]27 + "["
 $ColorTranslatorType = 'System.Drawing.ColorTranslator' -as [Type]
 $ColorType = 'System.Drawing.Color' -as [Type]
 
+function EscapseAnsiString([string]$AnsiString) {
+    if ($PSVersionTable.PSVersion.Major -ge 6) {
+        $res = $AnsiString -replace "$([char]0x1B)", '`e'
+    }
+    else {
+        $res = $AnsiString -replace "$([char]0x1B)", '$([char]0x1B)'
+    }
+
+    $res
+}
+
+function Test-VirtualTerminalSequece([psobject]$Object) {
+    if ($global:GitPromptSettings.AnsiConsole -and ($Object -is [string])) {
+        return $Object.Contains($AnsiEscape)
+    }
+    else {
+        return $false
+    }
+}
+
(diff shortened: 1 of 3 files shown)


rkeithhill:
BTW I want to add support for something like this `$GitPromptSettings.BranchColor.Foreground = 0xFF6600`.  Internally on systems that support ANSI, I'd try to convert that to an ANSI seq like `${e}[38;2;255;102;0m`.

src/GitPrompt.ps1
```diff
@@ -1,117 +1,221 @@
...
+            }
 
-    EnableWindowTitle                           = 'posh~git ~ '
+            $fg = $this.ForegroundColor
+            if ($fg -and !(Test-VirtualTerminalSequece $fg)) {
+                $fg = Get-ForegroundVirtualTerminalSequence $fg
+            }
 
-    AnsiConsole                                 = $Host.UI.SupportsVirtualTerminal -or ($Env:ConEmuANSI -eq "ON")
+            $str = "${fg}${bg}${txt}${e}0m"
```
rkeithhill: Update to not append ansi terminator if both fg and bg are $null (or empty).

src/AnsiUtils.ps1
```diff
@@ -24,10 +24,31 @@ $AnsiEscape = [char]27 + "["
 $ColorTranslatorType = 'System.Drawing.ColorTranslator' -as [Type]
 $ColorType = 'System.Drawing.Color' -as [Type]
 
+function EscapseAnsiString([string]$AnsiString) {
+    if ($PSVersionTable.PSVersion.Major -ge 6) {
+        $res = $AnsiString -replace "$([char]0x1B)", '`e'
```
rkeithhill: Pick either `[char]0x1b` or `[char]27` and stay consistent.

rkeithhill:
I think I will close this in favor of #513 which rounds out the desired functionality for `1.0.0` - better organized `$GitPromptSettings` and better factored `Write-GitStatus` functions.

PR #502, opened 2017-11-10
Directory names that contain [brackets] cause GitPrompt to fail

duncansmart:
Using `Get-Item -LiteralPath` means that "[brackets]" don't get interpreted as wildcards. 

Repro: 
- rename the directory in you have a git repo using square brackets e.g. `C:\code\foo[bar]`
- Open a posh-git enabled Powershell prompt in the directory
- The following error appears:  ``` C:\code\foo[bar] [Error: Cannot retrieve the dynamic parameters for the cmdlet. The specified wildcard character pattern is not valid: foo[bar]]```

### DIFF
--- a/src/GitUtils.ps1
+++ b/src/GitUtils.ps1
@@ -24,7 +24,7 @@ function Get-GitDirectory {
         $Env:GIT_DIR -replace '\\|/', [System.IO.Path]::DirectorySeparatorChar
     }
     else {
-        $currentDir = Get-Item $pathInfo -Force
+        $currentDir = Get-Item -LiteralPath $pathInfo -Force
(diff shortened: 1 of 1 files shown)


src/GitUtils.ps1
```diff
@@ -24,7 +24,7 @@ function Get-GitDirectory {
         $Env:GIT_DIR -replace '\\|/', [System.IO.Path]::DirectorySeparatorChar
     }
     else {
-        $currentDir = Get-Item $pathInfo -Force
+        $currentDir = Get-Item -LiteralPath $pathInfo -Force
```
rkeithhill: Good catch.  Thanks!

PR #504, opened 2017-12-08
Add `Add-PoshGitToProfile -AllUsers` support

dahlbyk:
Toward including an option to install posh-git with Git for Windows (https://github.com/git-for-windows/git/issues/1384), this adds an option to add to profile for all users.

Note that, as expected, `-AllUsers` from a non-elevated shell fails:

```
Add-Content : Access to the path
'C:\Windows\System32\WindowsPowerShell\v1.0\Microsoft.PowerShell_profile.ps1' is denied.
```
or
```
Add-Content : Access to the path
'C:\Windows\System32\WindowsPowerShell\v1.0\profile.ps1' is denied.
```

Running PowerShell as Administrator allows the global install.

### DIFF
--- a/src/Utils.ps1
+++ b/src/Utils.ps1
@@ -125,6 +125,10 @@ function Add-PoshGitToProfile {
         $TestParams
     )
 
+    if ($AllUsers -and !(Test-Administrator)) {
+        throw 'Installing posh-git for all users requires an elevated host.'
+    }
+
     $underTest = $false
 
     $profileName = $(if ($AllUsers) { 'AllUsers' } else { 'CurrentUser' }) `
--- a/src/Utils.ps1
+++ b/src/Utils.ps1
@@ -126,7 +126,7 @@ function Add-PoshGitToProfile {
     )
 
     if ($AllUsers -and !(Test-Administrator)) {
-        throw 'Installing posh-git for all users requires an elevated host.'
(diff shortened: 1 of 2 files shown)


rkeithhill:
BTW we do have this code to determine if the process is elevated or not. 

https://github.com/dahlbyk/posh-git/blob/65658bf6992282460726d6f4a552bae92073b1db/src/GitPrompt.ps1#L115-L124

We could check if the process is not elevated and the user specified `-AllHosts` and give a better error message.  One that tells them they need to be running as admin in order to update AllHosts profiles.

dahlbyk:
> We could check if the process is not elevated and the user specified -AllHosts and give a better error message.

Done.

> Installing posh-git for all users requires Administrator access.

src/Utils.ps1
```diff
@@ -104,9 +125,18 @@ function Add-PoshGitToProfile {
         $TestParams
     )
 
+    if ($AllUsers -and !(Test-Administrator)) {
+        throw 'Installing posh-git for all users requires an elevated host.'
```
rkeithhill: What do you think about rewording that as `'Adding posh-git to an AllUsers profile requires an elevated host.'`?

rkeithhill: Other than that, this PR looks good to me.

PR #506, opened 2017-12-12
Fix  duplicated branch completion for git checkout issue 505

### LINKED ISSUE #505: git checkout <Tab> tab-completion shows duplicated branch entries for branches that exist also on the remote
### System Details

- posh-git version/path: 0.7.1  C:\Program Files\WindowsPowerShell\Modules\posh-git\0.7.1
- PowerShell version: 5.1.16299.98
- git version 2.15.0.windows.1
- OS: Microsoft Windows NT 10.0.16299.0
- default settings, no customisation of `posh-git`

### Issue Description

Given a Git repo with branch `B1` being locally and on the remote upstream and branch `B2` being only locally and trying to tab complete as follows:
````powershell
git checkout B<Tab>
````

#### Expected Result
Tab completion should show 2 branches to tab complete

`B1 B2`

#### Actual Result

Branches that exist locally and on the remote are shown twice:

`B1 B1 B2`

### PR DESCRIPTION
**Pull request by bergmeister**
Fixes #505 by outputting only unique branches.
Instead of the proposed `Sort-Object -Unique`, I used `Select-Object -Unique` because:
- the 3 arrays for local branches, remote branches and tags are already sorted by themselves
- in order to not break existing behaviour and only fix the bug
- most user are less interested in tags and remote branches when using `git checkout`, therefore keeping the order as it was before makes sense
- better performance

### DIFF
--- a/src/GitTabExpansion.ps1
+++ b/src/GitTabExpansion.ps1
@@ -359,9 +359,11 @@ function GitTabExpansionInternal($lastBlock, $GitStatus = $null) {
 
         # Handles git checkout <ref>
         "^(?:checkout).* (?<ref>\S*)$" {
-            gitBranches $matches['ref'] $true
-            gitRemoteUniqueBranches $matches['ref']
-            gitTags $matches['ref']
+            $script:gitBranches = @(gitBranches $matches['ref'] $true)
+            $script:gitRemoteUniqueBranches = @(gitRemoteUniqueBranches $matches['ref'])
+            $script:gitTags = @(gitTags $matches['ref'])
+            # return only unique branches (to eliminate duplicates where the branch exists locally and on the remote)
+            $script:gitBranches + $script:gitRemoteUniqueBranches + $script:gitTags | Sort-Object -Unique
         }
 
         # Handles git worktree add <path> <ref>
--- a/src/GitTabExpansion.ps1
+++ b/src/GitTabExpansion.ps1
@@ -362,7 +362,7 @@ function GitTabExpansionInternal($lastBlock, $GitStatus = $null) {
             $script:gitBranches = @(gitBranches $matches['ref'] $true)
             $script:gitRemoteUniqueBranches = @(gitRemoteUniqueBranches $matches['ref'])
             $script:gitTags = @(gitTags $matches['ref'])
-            # return only unique branches (to eliminate duplicates where the branch exists locally and on the remote)
+            # Return only unique branches (to eliminate duplicates where the branch exists locally and on the remote)
             $script:gitBranches + $script:gitRemoteUniqueBranches + $script:gitTags | Select-Object -Unique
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/GitTabExpansion.ps1 [hunk 1]
```diff
@@ -359,9 +359,11 @@ function GitTabExpansionInternal($lastBlock, $GitStatus = $null) {
 
         # Handles git checkout <ref>
         "^(?:checkout).* (?<ref>\S*)$" {
-            gitBranches $matches['ref'] $true
-            gitRemoteUniqueBranches $matches['ref']
-            gitTags $matches['ref']
+            $script:gitBranches = @(gitBranches $matches['ref'] $true)
```
lzybkr: You aren't trying to cache the branches, so you probably shouldn't save them in a script scope variable.
I think using the pipeline like this might be a bit more natural and avoids creating multiple arrays:
````powershell
& {
    gitBranches $matches['ref'] $true
    gitRemoteUniqueBranches $matches['ref']
    gitTags $matches['ref']
} | Select-Object -Unique
````
You could use a function instead of the script block like I used here.

Comment (bergmeister) [reply src/GitTabExpansion.ps1]: Thanks for the suggestion, you're right that this is the more efficient way of doing it.

File: src/GitTabExpansion.ps1 [hunk 2]
```diff
@@ -359,11 +359,12 @@ function GitTabExpansionInternal($lastBlock, $GitStatus = $null) {
...
         "^(?:checkout).* (?<ref>\S*)$" {
-            $script:gitBranches = @(gitBranches $matches['ref'] $true)
-            $script:gitRemoteUniqueBranches = @(gitRemoteUniqueBranches $matches['ref'])
-            $script:gitTags = @(gitTags $matches['ref'])
-            # Return only unique branches (to eliminate duplicates where the branch exists locally and on the remote)
-            $script:gitBranches + $script:gitRemoteUniqueBranches + $script:gitTags | Select-Object -Unique
+            & {
+                @(gitBranches $matches['ref'] $true)
+                @(gitRemoteUniqueBranches $matches['ref'])
+                @(gitTags $matches['ref'])
```
dahlbyk: Shouldn't need the surrounding `@(...)` on these.

bergmeister: Well spotted, this was a left over from the previous version when I was using arrays.

dahlbyk:
I just noticed this is targeting `develop`; I've opened #507 against `master`.

Still wish GitHub supported retargeting PRs...

bergmeister:
Hmm, isn't that what `gitflow` would be for (except that it is from the command line)

dahlbyk:
Despite the alignment of branch names, we don't use Git Flow. `master` is still targeting v0.x, `develop` is code with breaking changes working toward v1.0.

PR #510, opened 2017-12-12
Fix PSPossibleIncorrectComparisonWithNull PSScriptAnalyzer warnings

bergmeister:
See [here](https://github.com/PowerShell/PSScriptAnalyzer/blob/development/RuleDocumentation/PossibleIncorrectComparisonWithNull.md) for some background about the rule


dahlbyk:
Thanks!

PR #511, opened 2017-12-12
Fix PSUseBOMForUnicodeEncodedFile PSScriptAnalyzer warning

bergmeister:
Fix [UseBOMForUnicodeEncodedFile](https://github.com/PowerShell/PSScriptAnalyzer/blob/development/RuleDocumentation/UseBOMForUnicodeEncodedFile.md)


bergmeister:
Ok, that's fine if that's the case but that rather suggests that the rule itself shoud be re-discussed and posssibly changed.

rkeithhill:
Yup, see - https://github.com/PowerShell/PSScriptAnalyzer/issues/810

PR #518, opened 2018-01-02
WIP - initial stab at a simple build script

rkeithhill:
We can add more steps to this to help automate preparation for publication.  Or if you don't want to go this route that's OK too.  I spent about an hour putting this together to see what's possible.  Most of that time was wasted on the worthless (buggy) `Update-ModuleManifest` command.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,7 @@
 VERSION
 *.nupkg
 obj
+
+# Ignore out dirs named after module version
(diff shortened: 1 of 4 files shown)


rkeithhill:
There are some more full featured module build systems coming into existence.  If we decide to go in this direction, I would use one of those modules at this point.  Probably https://github.com/psake/PowerShellBuild

PR #520, opened 2018-01-10
Default Prompt Colors

### LINKED ISSUE #469: Add custom path rendering in prompt
Along the same lines as with the work done for issue #386 it would be nice to more finely tune the path portion of the prompt. With older versions of posh-git I would typically hand edit my Profile to customize this but with the more extensible model through use of variables I think extending it this way will be cleaner and make it easier to setup a new custom environment up in a way that is less likely to break with updates.

## Example custom path

Before:

```
C:\root\sub1\sub2\sub3\workingdir [develop ≡ +0 ~1 -0 !]> 
```

After: 

```PowerShell
$GitPromptSettings.DefaultPromptPathTemplate = '$([System
.IO.Path]::GetFileName($currentPath))'
```

```
workingdir [develop ≡ +0 ~1 -0 !]> 
```

## What about non file-system paths?

Perhaps it would be more convenient to have a separate `DefaultPromptFilePathTemplate` and `DefaultPromptPathTemplate` just reduce the need for logical checks on the type of path for cases where the path is not a file system path. 

## What about home directory abbreviation?

Perhaps `DefaultPromptAbbreviateHomeDirectory` can be re-implemented in terms of this template.

## Template vs. Function

Maybe it's more reasonable to provide a way way of supplying or overriding a function that provides the path string as a result instead of trying to embed this stuff into a expanded string template. I'm not sure what the common conventions are in the PowerShell world for providing for this type of extensibility model but I have some ideas of what may work.

I'll be happy to take a stab at contributing the change if it's something others may be interested in.

### LINKED ISSUE #474: Provide DefaultPromptPrefixColor and DefaultPromptSuffixColor options
### System Details

- posh-git version/path: 0.7.1  C:\tools\poshgit\dahlbyk-posh-git-a4faccd\src
- PowerShell version: 5.1.15063.296
- git version 2.13.0.windows.1
- OS: Microsoft Windows NT 10.0.15063.0

### Issue Description

I would like to be able to easily customize the default prompt prefix and suffix colors independently from each other and the default foreground color. Perhaps this could be provided with DefaultPromptPrefixColor and DefaultPromptSuffixColor properties in $GitPromptSettings?

In prior versions I overrode the prompt function to accomplish this, however, it would be very nice to be able to change these colors without needing to override the prompt function via the enhanced $GitPromptSettings properties.

dahlbyk:
Much easier to close #474 with the changes from #513.

- `DefaultPromptPrefix`, `DefaultPromptSuffix` and `DefaultPromptDebugSuffix` are now `PoshGitTextSpan`s that understand colors
- <del>`DefaultPromptPathColor`</del> and `DefaultPromptTimingColor` are supported, too

**Update:** Also closes #469 by:

- Exporting `Get-PromptPath` with our default (`$GitPromptSettings`-based) path behavior
- Supporting `DefaultPromptPath` (also a `PoshGitTextSpan`) to control colors and the string to expand when `prompt` executes (`'$(Get-PromptPath)'` by default)

To customize how the path is displayed, the user would do something like...

```powershell
function Get-FancyPromptPath {
  (Get-PromptPath).ToUpper()
}
$GitPromptSettings.DefaultPromptPath.Text = '$(Get-FancyPromptPath)'
```

### DIFF
--- a/src/Utils.ps1
+++ b/src/Utils.ps1
@@ -282,6 +282,27 @@ function Get-PathStringComparison {
     }
 }
 
+function Get-PromptPath {
+    $settings = $global:GitPromptSettings
+    $abbrevHomeDir = $settings -and $settings.DefaultPromptAbbreviateHomeDirectory
+
+    # A UNC path has no drive so it's better to use the ProviderPath e.g. "\\server\share".
+    # However for any path with a drive defined, it's better to use the Path property.
+    # In this case, ProviderPath is "\LocalMachine\My"" whereas Path is "Cert:\LocalMachine\My".
+    # The latter is more desirable.
+    $pathInfo = $ExecutionContext.SessionState.Path.CurrentLocation
+    $currentPath = if ($pathInfo.Drive) { $pathInfo.Path } else { $pathInfo.ProviderPath }
+
+    $stringComparison = Get-PathStringComparison
+
+    # Abbreviate path by replacing beginning of path with ~ *iff* the path is in the user's home dir
+    if ($abbrevHomeDir -and $currentPath -and $currentPath.StartsWith($Home, $stringComparison)) {
+        $currentPath = "~" + $currentPath.SubString($Home.Length)
+    }
+
+    return $currentPath
+}
+
 function Get-PSModulePath {
     $modulePaths = $Env:PSModulePath -split ';'
     $modulePaths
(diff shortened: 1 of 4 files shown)


dahlbyk:
I just pushed an additional commit to pivot from `DefaultPromptPathColor` to `DefaultPromptPath` to close #469. Updating this PR description to reflect the change.

image

src/Utils.ps1
```diff
@@ -282,6 +282,27 @@ function Get-PathStringComparison {
     }
 }
 
+function Get-PromptPath {
```
rkeithhill: It's tempting to add parameters to this function Iike `-AbbreviateHomeDir` and `-PathTruncateLimit` which would eliminate the need for the equivalent $GitPromptSettings but the "discoverability" of this functionality would be very poor.  Probably best to leave those in $GitPromptSettings.

dahlbyk: That's exactly how I started, but moved away from it in favor of keeping the default `DefaultPromptPath` text simple (`'$(Get-PromptPath)').

Issue #302, opened 2016-08-21
PowerShell Core on Linux support


theaquamarine:
Issues I've found running posh-git on OS X:
- [ ] Path separators are a known issue in PowerShell on Linux: only forward slashes are supported (https://github.com/PowerShell/PowerShell/blob/master/docs/KNOWNISSUES.md#directly). profile.example.ps1 uses `.\posh-git` rather than `Join-Path . posh-git`. Failing that, swapping to forward slash would work as they are supported everywhere.
- [x] Administrator check in GitPrompt.ps1:98 doesn't work as `[Security.Principal.WindowsIdentity]::GetCurrent()` is unsupported. 
- [x] GitUtils.ps1:226 doesn't work for setting env variables, as [EnvironmentVariableTarget] is unavailable. We use Process, which is the default, so could skip this.
- [x] The prompt shows up twice. This isn't actually our fault, there's an issue with using `Write-Host` in the prompt on OS X/Linux. Tracked upstream at https://github.com/PowerShell/PowerShell/issues/1897
- [ ] Actual ssh-agent support could probably be better. ssh-agent always runs OS X 10.5+ and uses Keychain as storage so I'm not sure how much posh-git actually needs to manage agent/keys on OS X.

jonathankarsh:
Can confirm similar issues exist on Ubuntu 14.04.

andschwa:
Path separators actually shouldn't be an issue. PowerShell on Linux translates backslashes to forward slashes to be "slash agnostic"; what this prevents is a backslash as a literal character in a path name due to the inability to escape it. But your use of `\` as a `/` should be just fine.

Anything using `Security.Principal.WindowsIdentity` is just (expectedly) not going to work on Linux.

You honestly may just want to wait for `EnvironmentVariableTarget`. It's been [added back to CoreFX](https://github.com/dotnet/corefx/commits/9348997920c1ba782b45adb5e8ee9754149ab1e2/src/System.Runtime.Extensions/src/System/EnvironmentVariableTarget.cs). However, if you note the date, that was July 5th, and the stable release that PowerShell is using was [published June 27th](https://www.nuget.org/packages/System.Runtime.Extensions/). So when .NET Core drops their next big release and PowerShell picks it up, this problem just goes away. (And then `GetCurrent` can turn into `[Environment]::UserName` probably.)

Double prompting is a strange bug; I believe @ealexjordan is working on it.

dahlbyk:
Thanks, @andschwa!

[...]

PR #523, opened 2018-01-12
Add Travis config for Linux

theaquamarine:
.travis.yml as discussed in #302. Linux only at the moment, Mac OS builds have a very long queue so taking longer to test.

Pester is set to use the number of failed tests as the exit code, so build is reported as failed due to 41 tests failing (a large chunk are references to `git.exe` or the C: drive).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,20 @@
+language: generic
+
+matrix:
+  include:
+    - os: linux
+      dist: trusty
+      sudo: false
+      addons:
+        apt:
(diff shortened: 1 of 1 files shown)


dahlbyk:
Once again wishing GitHub allowed retargeting PRs. @theaquamarine can you reopen this rebased on `develop`and  targeting `develop`?

As long as we keep `.travis.yml` out of the `v0.x` history we should be able to avoid running these builds where they can never pass.

PR #524, opened 2018-01-13
Add Travis config for Linux

theaquamarine:
For #302, rebased against develop as requested in #523.

Reports as failed due to 40 tests failing, again largely references to `git.exe` and `C:`.

PR #525, opened 2018-01-13
Travis for Mac OS

theaquamarine:
Builds on the Linux config in #524, from #302.

Mac OS builds report the same errors as Linux (which makes sense, since from a PowerShell perspective they're about identical), so adding it as a target may not be desirable, given the very long queue lengths Travis normally has- a couple of hours at the moment. Travis won't report back to GitHub until both builds are done, but results for Linux and Mac OS can be seen separately on the actual build page.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,9 @@
 language: generic
 
+branches:
+  only:
+  - develop
+
 stages:
 - linux
 - osx
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,8 @@ matrix:
       before_install:
         - pwsh -c 'Set-PSRepository -Name PSGallery -InstallationPolicy Trusted'
         - pwsh -c 'Install-Module Pester -Scope CurrentUser'
-    - os: osx
+    - if: branch = develop
+      os: osx
       before_install:
         - brew update
         - brew tap caskroom/cask
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,20 @@
+language: generic
+
+matrix:
+  include:
+    - os: linux
+      dist: trusty
+      sudo: false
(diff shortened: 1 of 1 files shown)


dahlbyk:
Is it possible to only trigger Mac builds on `develop`, as a sanity check, but not on PRs?

theaquamarine:
Ok, that appears to be working. It's [going to build on Mac OS for this pull against develop](https://travis-ci.org/dahlbyk/posh-git/builds/328463756), but [not on my Travis branch](https://travis-ci.org/theaquamarine/posh-git/builds/328463746).

dahlbyk:
This latest Travis build seems o be stalled: https://travis-ci.org/dahlbyk/posh-git/builds/328463756. Linux failed fast, but it appears to be waiting for the Mac build to start?

dahlbyk:
Check out https://stackoverflow.com/a/47734517/54249

theaquamarine:
Yeah, that's the delay in the job queue for a Mac OS VM. It'll list the build as started, not returning success or failure until the Mac OS build is at least attempted. Eventually the Mac OS job will run and Travis will report success/failure, but it can take a while- [this one](https://travis-ci.org/theaquamarine/posh-git/builds/328307849) waited over three hours to get to the head of the queue, the last commit in this PR took nearly two, and I've had several wait longer than that. [The status page](https://www.traviscistatus.com/) suggests their Mac OS builds are pretty much always backlogged.

We could try narrowing the condition more, but I can't think of anything that'd make Travis return faster other than dropping Mac OS entirely- which is probably reasonable since Linux is going to be pretty close and builds in a matter of seconds.

theaquamarine:
1 hour 50 minutes for that to finish.

PR #526, opened 2018-01-13
Separate out SSH functionality

theaquamarine:
Start towards #338: Move the SSH stuff into separate files and break dependencies. Remaining SSH stuff outside of SshUtils.ps1:

- SshAgent bit of Add-PoshGitToProfile
- function exports


dahlbyk:
Thanks!

PR #528, opened 2018-01-13
Add Visual Studio pssproj & solution

theaquamarine:
Sets it up so poshtools is ready to go with tests loaded.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 VERSION
 *.nupkg
 obj
+.vs/
--- a/posh-git.pssproj
+++ b/posh-git.pssproj
@@ -1,0 +1,62 @@
+﻿<Project ToolsVersion="4.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
+    <SchemaVersion>2.0</SchemaVersion>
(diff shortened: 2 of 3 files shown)


rkeithhill:
So, why do we need Visual Studio and PoshTools (Windows only)? There isn't a build per-se for this module yet.  We just a need to run Pester tests. This seems a bit sledgehammer-ish.  :-)

theaquamarine:
Mostly just because I find it useful, so others might. I figure anything that might make contributing for someone new is a plus, though I suppose the niche of people who use VS, have poshtools and don't know how to use Pester any other way is going to be fairly small.

rkeithhill:
Yeah, given that the push here is to support PS Core which is cross-platform, it would be nice to avoid using "Windows-only" tools like Visual Studio.   Also, the intersection of that Venn diagram you mention, is likely very small.  :-)  

If we really want to do this, I think it would be better to put the two VS specific files in a subdir like `.vs`.

dahlbyk:
I'm ambivalent on this. On one hand, including these files may reduce barriers to contribution for folks familiar with the VS tools, without interfering with those who won't use them. On the other hand, the `.pssproj` seems somewhat likely to get out-of-date if it's not actively used by most contributors, leading to frustration for would-be contributors if things don't Just Work™️.

What if we just ignored `*.pssproj` and `*.sln` along with `.vs/`?

rkeithhill:
The suggested solution is to create another PR to add *.pssproj, *.sln and .vs/ to the .gitignore file.

PR #532, opened 2018-01-15
Fixes two bugs - one ANSI and one PS Core

rkeithhill:
The ANSI bug is that if both fg and bg are default ($null) we were still emitting the terminating seq `e[0m.

The PS Core bug is that on PS Core there is no byte value for Get-Content -Encoding.   The replacement is Get-Content -AsByteStream.

Also. added ToEscapedString() method to [PoshGitCellColor] and [PoshGitTextSpan].  Was invaluable for working on the Pester tests and  I think it will be very handy for debugging.  If someone's GitPrompSettings isn't displaying properly we can ask them to execute: `$GitPromptSettings.DelimText.ToEscapedString()`

This will output:
``` `e[93m |`e[0m```

### DIFF
--- a/src/PoshGitTypes.ps1
+++ b/src/PoshGitTypes.ps1
@@ -131,57 +131,7 @@ class PoshGitTextSpan {
         $this.CustomAnsi = $null
     }
 
-    [string] ToEscapedString() {
-        if ($global:GitPromptSettings.AnsiConsole) {
-            $ansiTerm = EscapseAnsiString ([char]27 + "[0m")
-
-            if ($this.CustomAnsi) {
-                $escAnsi = EscapseAnsiString $this.CustomAnsi
-                $str = "${escAnsi}$($this.Text)$ansiTerm"
-            }
-            else {
-                $color = [PoshGitCellColor]::new($this.ForegroundColor, $this.BackgroundColor)
-                $escAnsi = $color.ToEscapedString()
-                if ($escAnsi) {
-                    $str = "${escAnsi}$($this.Text)$ansiTerm"
-                }
-                else {
-                    $str = $this.Text
-                }
-            }
-        }
-        else {
-            $str = $this.Text
-        }
-
-        return $str
-    }
-
-    [string] ToString() {
-        if ($global:GitPromptSettings.AnsiConsole) {
-            if ($this.CustomAnsi) {
-                $e = [char]27 + "["
-                $ansi = $this.CustomAnsi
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/PoshGitTypes.ps1 [hunk 1]
```diff
@@ -101,6 +131,32 @@ class PoshGitTextSpan {
         $this.CustomAnsi = $null
     }
 
+    [string] ToEscapedString() {
```
dahlbyk: Isn't this whole method equivalent to `EscapeAnsiString $this.RenderAnsi()`?

src/PoshGitTypes.ps1
```diff
@@ -44,11 +44,41 @@ class PoshGitCellColor {
         return $str
     }
 
+    [string] ToEscapedString() {
```
dahlbyk: Does it make sense for `PoshGitCellColor` to have a `RenderAnsi()`, such that this function is essentially `EscapeAnsiString $this.RenderAnsi()`? That function could then be used in `PoshGitTextSpan`.

dahlbyk: Also, would `ToAnsiString()` be a better name than `RenderAnsi()`?

rkeithhill: That's what I get for coding and watching the playoffs.  :-)  Renamed `RenderAnsi()` to `ToAnsiString()`.   Can we do the `[PoshGitCellColor]` tweaks - ToAnsiString() and ToEscapedString() simplification in another PR?

src/PoshGitTypes.ps1
rkeithhill: Yup.  Done.

src/PoshGitTypes.ps1
```diff
@@ -144,7 +150,47 @@ class PoshGitTextSpan {
...
+            else {
+                $str = $txt
+            }
+        }
+
+        return $str
+    }
+
+    [string] ToEscapedString() {
+        if ($global:GitPromptSettings.AnsiConsole) {
```
dahlbyk: Not a blocker, but should `ToEscapedString()` care if the console supports ANSI? Properly encoding/escaping a string doesn't require it…

src/PoshGitTypes.ps1
dahlbyk: Yeah, that works. While you're at it, fix the typo in `EscapseAnsiString`. 😏

rkeithhill: Doh! Yeah, will do.

Issue #527, opened 2018-01-13
Test on PSCore on Windows


theaquamarine:
We've got CI running on every other platform now. [SHiPS's appveyor.yml](https://github.com/PowerShell/SHiPS/blob/development/appveyor.yml) appears to be doing it using environment variables to separate the two editions, just installing PSCore and running using pwsh should be plenty.

dahlbyk:
Should we try to do this with GitHub Actions?

PR #533, opened 2018-01-15
Fix tests on WinPS, PS Core Windows and Linux

### LINKED ISSUE #531: Fix Pester tests on Linux
FYI, I'm currently working on this.  I'll be updating to Pester 4 syntax in the process.

rkeithhill:
Fixes #531

Still have a bit of clean-up work to do but wanted to see how the updated tests performed.  Well, after PR #532 is merged since that PRs fixes several Pester test failures.

### DIFF
--- a/test/DefaultPrompt.Tests.ps1
+++ b/test/DefaultPrompt.Tests.ps1
@@ -1,47 +1,43 @@
 . $PSScriptRoot\Shared.ps1
 
-Describe 'Default Prompt Tests' {
+Describe 'Default Prompt Tests - NO ANSI' {
     BeforeAll {
-        $GitPromptSettings.AnsiConsole = $false
         $prompt = Get-Item Function:\prompt
         $OFS = ''
     }
     BeforeEach {
         # Ensure these settings start out set to the default values
-        $GitPromptSettings.DefaultPromptPrefix = ''
-        $GitPromptSettings.DefaultPromptSuffix = '$(''>'' * ($nestedPromptLevel + 1)) '
-        $GitPromptSettings.DefaultPromptDebugSuffix = ' [DBG]$(''>'' * ($nestedPromptLevel + 1)) '
-        $GitPromptSettings.DefaultPromptAbbreviateHomeDirectory = $false
-        $GitPromptSettings.DefaultPromptEnableTiming = $false
+        $global:GitPromptSettings = & $module.NewBoundScriptBlock({[GitPromptSettings]::new()})
+        $GitPromptSettings.AnsiConsole = $false
     }
 
     Context 'Prompt with no Git summary' {
         It 'Returns the expected prompt string' {
-            Set-Location $env:windir -ErrorAction Stop
+            Set-Location $env:HOME -ErrorAction Stop
             $res = [string](&$prompt 6>&1)
-            $res | Should BeExactly "$env:windir> "
+            $res | Should BeExactly "${env:HOME}> "
         }
         It 'Returns the expected prompt string with changed DefaultPromptPrefix' {
             Set-Location $Home -ErrorAction Stop
-            $GitPromptSettings.DefaultPromptPrefix = 'PS '
+            $GitPromptSettings.DefaultPromptPrefix.Text = 'PS '
             $res = [string](&$prompt 6>&1)
-            $res | Should BeExactly "PS $Home> "
+            $res | Should BeExactly "PS ${Home}> "
         }
         It 'Returns the expected prompt string with expanded DefaultPromptPrefix' {
             Set-Location $Home -ErrorAction Stop
-            $GitPromptSettings.DefaultPromptPrefix = '[$(hostname)] '
+            $GitPromptSettings.DefaultPromptPrefix.Text = '[$(hostname)] '
             $res = [string](&$prompt 6>&1)
             $res | Should BeExactly "[$(hostname)] $Home> "
         }
         It 'Returns the expected prompt string with changed DefaultPromptSuffix' {
             Set-Location $Home -ErrorAction Stop
-            $GitPromptSettings.DefaultPromptSuffix = '`n> '
+            $GitPromptSettings.DefaultPromptSuffix.Text = '`n> '
             $res = [string](&$prompt 6>&1)
             $res | Should BeExactly "$Home`n> "
         }
         It 'Returns the expected prompt string with expanded DefaultPromptSuffix' {
             Set-Location $Home -ErrorAction Stop
-            $GitPromptSettings.DefaultPromptSuffix = ' - $(6*7)> '
+            $GitPromptSettings.DefaultPromptSuffix.Text = ' - $(6*7)> '
             $res = [string](&$prompt 6>&1)
             $res | Should BeExactly "$Home - 42> "
         }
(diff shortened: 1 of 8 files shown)


dahlbyk:
Updated to include merged #532.

rkeithhill:
Do we want to run `InstallChocolatey.Tests.ps1` on Linux?  I'm thinking no.

rkeithhill:
WIndows and Linux pass.  Just waiting for the macOS build.

rkeithhill:
So there are failures on macOS but I'm not going to be able to debug these - don't have have mac.  But both Windows and Linux are passing.  I want to wait for the next macOS build to finish to see if the latest commit made any difference.

test/Shared.ps1
```diff
@@ -1,20 +1,44 @@
...
 $moduleManifestPath = "$modulePath\posh-git.psd1"
 
+$csi = [char]0x1b + "["
+if (($PSVersionTable.PSVersion.Major -le 5) -or $IsWindows) {
+    # On Windows, we can access the git binary via git.exe
+    $global:gitbin = Get-Command -Name git -CommandType Application
+}
+else {
+    # On Linux/macOS, we can access the git binary via its path /usr/bin/git
+    $global:gitbin = (Get-Command -Name git -CommandType Application).Path
```
theaquamarine: I've got two versions of git installed on Mac OS, and iirc that's normal with Homebrew, so test using this break at the moment. `(Get-Command blah)[0]` fixes them.

dahlbyk:
> Try to exclude osx via stages filter

This PR _shouldn't_ be running on `osx`. Hoping this will turn the build green, then a separate PR can temporarily disable the OSX branch filter to troubleshoot Mac-specific issues.

theaquamarine:
I think Travis' thinking is this is a PR going in to develop, so gets treated as being on develop, rather than on the source branch? Kind of makes sense, I guess. With the fix above, it's 94 passes and 2 failures on Mac OS- Test Worktree, and Test bare repository in the Get-GitDirectory tests. Something to do with Mac OS's temp folders.

test/Shared.ps1
dahlbyk: Added `-TotalCount 1`

theaquamarine:
The remaining Mac OS failures are due to the fact Mac OS's temp folder has a symlink in its path- `/var` is linked to `/private/var`. `git` resolves the symlink to `/private/var/`, but `[System.IO.Path]::GetTempPath()` returns the path starting with `/var`. I don't think PowerShell has a way to resolve the path, so it might just be easiest to special-case Mac OS and add `/private` to the start of the temp path?

rkeithhill:
@theaquamarine Do you want to take a whack at fixing the temp dir issue?  Maybe we should have a function in Shared.ps1 called `Get-PlatformTempPath()` where this can be done right once and then use it within Shared.ps1 (NewGitTempRepo) and Get-GitDirectory.Tests.ps1.

rkeithhill:
Wow, the wait on the macOS build is epic.  :-(

rkeithhill:
Argh, this is like waiting for water to boil.  Strike that - much worse.  :-)

rkeithhill:
@theaquamarine Nevermind. I took a stab at it.  Really want to get this PR in so my other (soon to be submitted PRs) won't have test failures.

test/Utils.Tests.ps1
```diff
@@ -1,6 +1,8 @@
 . $PSScriptRoot\Shared.ps1
 . $modulePath\Utils.ps1
 
+$expectedEncoding = if ($PSVersionTable.PSVersion.Major -le 5) { "utf8" } else { "ascii" }
```
dahlbyk: For PS Core we should default BOM-less files to `UTF8NoBOM` instead of ASCII, or maybe `Default` adequately covers both?

test/DefaultPrompt.Tests.ps1
```diff
@@ -74,23 +70,127 @@ Describe 'Default Prompt Tests' {
...
 
             $res = [string](&$prompt 6>&1)
-            Assert-MockCalled git -ModuleName posh-git
+            Assert-MockCalled git -ModuleName posh-git -Scope It
             $res | Should BeExactly "$PSScriptRoot [master +1 ~0 -0 | +0 ~1 -1 !]> "
         }
     }
 }
+
+Describe 'Default Prompt Tests - ANSI' {
```
dahlbyk: 💯

test/DefaultPrompt.Tests.ps1
```diff
@@ -53,8 +49,8 @@ Describe 'Default Prompt Tests' {
         }
         It 'Returns the expected prompt string with prefix, suffix and abbrev home set' {
             Set-Location $Home -ErrorAction Stop
-            $GitPromptSettings.DefaultPromptPrefix = '[$(hostname)] '
-            $GitPromptSettings.DefaultPromptSuffix = ' - $(6*7)> '
+            $GitPromptSettings.DefaultPromptPrefix.Text = '[$(hostname)] '
+            $GitPromptSettings.DefaultPromptSuffix.Text = ' - $(6*7)> '
```
dahlbyk: If we're assuming no colors, is there a downside to setting the property directly (meaning a new span is created from the string constructor) instead of setting `Text`?

rkeithhill: No, just a minor perf hit no one would likely notice.

test/Utils.Tests.ps1
rkeithhill: There is no `Default` value for the `-Encoding` parameter on PS Core.  But obviously there is a default value that is used when `-Encoding` isn't specified.  I'm pretty sure that is utf8NoBom which is probably what we should be using on PS Core Linux/macOS.   I think UTF8 is still a good default for Windows.

sdwheeler: Default encoding is UTF-8 without a BOM
PowerShell Core changes the default encoding to conform with the broader ecosystems.

See https://github.com/PowerShell/PowerShell-Docs/blob/staging/reference/docs-conceptual/whats-new/What-s-New-in-PowerShell-Core-60.md#default-encoding-is-utf-8-without-a-bom

rkeithhill: Looks like PS Core on Windows uses `utf8NoBom` as well. 

The tricky part here that where we use this we are typically adding to an existing profile.  I would consider it a bit rude to change the file encoding of someone's profile script.  I think we are fairly safe to assume that PS Core profiles will be utf8NoBom given the default.  Not so much for Windows PowerShell though.  And our function to detect file encoding is pretty dumb.  It can't tell the diff between uft8NoBom and ASCII because it only looks for a BOM (or lack thereof).

sdwheeler: This might work for encoding detection: https://gist.github.com/sdwheeler/777b9ec25e914bdeb0774471a149951c#file-get-fileencoding-ps1

I think Lee Holmes wrote this. I found it on poscode.org a few years back. It does a decent job with BOM-less files.

rkeithhill: It is probably better than what we've got but not by much.  It appears to be just looking for a BOM as well.

rkeithhill:
Ugh, now the Linux build is hung as well.

test/Utils.Tests.ps1
dahlbyk: Re: no `Default`, I guess i misread https://github.com/PowerShell/PowerShell/blob/4bc52d2358222084738157a08907fac32d13bd3a/src/System.Management.Automation/utils/EncodingUtils.cs#L37

If there's no BOM, you have to make an assumption; trying to detect encoding is notoriously difficult, and we'll beyond our needs here. Pre-Core the PS assumption is ANSI; in Core the assumption is UTF8, on all platforms. I don't see much reason not to align with that. If it causes a problem for folks, they should get on board with UTF8.

The Holmes code is basically equivalent to what we have here: match by BOM, otherwise default to something (UTF7, in that case).

rkeithhill: > in Core the assumption is UTF8

Well, UTF8 with no BOM just to be clear/pedantic.  :-)

dahlbyk:
https://www.traviscistatus.com/incidents/c1s4dlyxd4lf

rkeithhill:
Doh, looks like I broke Travis.  :-)

rkeithhill:
Should we just go ahead and merge this?  The wait is getting ridiculous.

rkeithhill:
According the incident status page, they're going to cancel all queued builds to try to get it running again.

theaquamarine:
Looks like we're good https://travis-ci.org/dahlbyk/posh-git/builds/329647338

dahlbyk:
Merged!

> Fix tests on PS Core Windows

We don't actually have CI running tests on PS Core Windows yet, correct?

theaquamarine:
We do not. #527

PR #536, opened 2018-01-18
Lookup HtmlColors before ConsoleColors

rkeithhill:
Turns out ConsoleColors are a lot easier to specify than HtmlColors (that happen to match ConsoleColor names).

Add `PSEdition_Core` tag so that posh-git shows up if someone searches for that tag - https://www.powershellgallery.com/items?q=tags%3Apsedition_core&x=0&y=0

### DIFF
--- a/src/AnsiUtils.ps1
+++ b/src/AnsiUtils.ps1
@@ -58,11 +58,11 @@ function Get-VirtualTerminalSequence ($color, [int]$offset = 0) {
 
     if ($color -is [String]) {
         try {
-            if ($null -ne ($color -as [System.ConsoleColor])) {
-                $color = [System.ConsoleColor]$color
-            }
-            elseif ($ColorTranslatorType) {
+            if ($ColorTranslatorType) {
                 $color = $ColorTranslatorType::FromHtml($color)
+            }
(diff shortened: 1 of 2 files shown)


src/AnsiUtils.ps1
```diff
@@ -58,12 +58,12 @@ function Get-VirtualTerminalSequence ($color, [int]$offset = 0) {
...
         try {
-            if ($null -ne ($color -as [System.ConsoleColor])) {
-                $color = [System.ConsoleColor]$color
-            }
-            elseif ($ColorTranslatorType) {
+            if ($ColorTranslatorType) {
                 $color = $ColorTranslatorType::FromHtml($color)
             }
+            elseif ($null -ne ($color -as [System.ConsoleColor])) {
+                $color = [System.ConsoleColor]$color
```
lzybkr: This is the anti-pattern of converting twice - you should either use `-as` or the cast, but not both.

rkeithhill: OK, caught me being lazy but at least we fixed the bug where `[System.ConsoleColor]::Black` (value of 0) was not registering as a ConsoleColor.  :-)

PR #538, opened 2018-01-18
Added Setting to change Admin text in window title.

### LINKED ISSUE #537: Feature request:   Add a setting for the AdminText prefix in window title
I dislike the "Administor: " prefix on the powershell window title because it makes it harder to determine the subject of a window when it is minimized.  I usually have 3-6 powershell windows open, so this is important for my workflow.

I already have a commit with a fix for this.  It's six-line change to add the setting as well as move the $adminHeader calculation to the bottom of Write-GitStatus where the other window title logic is located.

nebosite:
Also moved the logic for this into write-gitstatus, so all of the title logic is together

Closes #537

### DIFF
--- a/src/GitPrompt.ps1
+++ b/src/GitPrompt.ps1
@@ -99,6 +99,7 @@ $global:GitPromptSettings = [pscustomobject]@{
     DescribeStyle                               = ''
 
     EnableWindowTitle                           = 'posh~git ~ '
+    AdminTitlePrefixText                        = 'Administrator: '
 
     DefaultPromptPrefix                         = ''
     DefaultPromptSuffix                         = '$(''>'' * ($nestedPromptLevel + 1)) '
@@ -113,8 +114,6 @@ $global:GitPromptSettings = [pscustomobject]@{
 }
 
 $isAdminProcess = Test-Administrator
-
-$adminHeader = if ($isAdminProcess) { 'Administrator: ' } else { '' }
 
 $WindowTitleSupported = $true
 if (Get-Module NuGet) {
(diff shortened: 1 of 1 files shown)


src/GitPrompt.ps1
```diff
@@ -112,10 +113,6 @@ $global:GitPromptSettings = [pscustomobject]@{
     TruncatedBranchSuffix                       = '...'
 }
 
-$isAdminProcess = Test-Administrator
```
dahlbyk: This value can't change in the context of a host instance, so let's keep this cached here.

src/GitPrompt.ps1
```diff
@@ -112,9 +113,7 @@ $global:GitPromptSettings = [pscustomobject]@{
...
-
-$adminHeader = if ($isAdminProcess) { 'Administrator: ' } else { '' }
+$isAdminProcess = Test-Administrator
```
dahlbyk: Trailing space! Feel free to fix and squash into a single commit.

rkeithhill:
@nebosite Which version of PowerShell are you using?  You can check with `$PSVersionTable.PSVersion`.  If you are on v5 or higher, would you be interested in moving to posh-git 1.0.0 (currently in beta) to get this feature?  

I'm just wondering if the fix that is on the v1 branch is worth porting to v0.7x.  If you could move to v1, then I recommend we skip applying this fix to v0.7x (in which case we can close this issue as fixed by PR #541) .  Part of the point of v1 is to provide richer prompt customization.

rkeithhill:
If we merge this into the 0.x branch we should not merge that to 1.x branch.  We have a different mechanism in 1.x that accomplishes the same goal.

dahlbyk:
> If we merge this into the 0.x branch we should not merge that to 1.x branch. We have a different mechanism in 1.x that accomplishes the same goal.

This setting will disappear as part of resolving the merge conflict.

PR #541, opened 2018-01-25
Improved WindowTitle customization experience

rkeithhill:
v1 fix for https://github.com/dahlbyk/posh-git/issues/537

Needs more Pester tests.  Also, I think we might want to keep `EnableWindowTitle` as a simple bool.  If you disable the WindowTitle by setting it to $null, you lose the original definition.  

Also you made a comment about evaluating the WindowTitle scriptblock during prompt eval. It does seem a bit unusual that we change the WindowTitle in the Write-GitStatus function.  Is the proposal to move the setting of the host WindowTitle to the built-in prompt?

BTW there was some code we used to determine WindowTitleSupported that I've thought was circumspect for a while. This gave me a chance to take a better (I hope) approach.

### DIFF
--- a/src/GitPrompt.ps1
+++ b/src/GitPrompt.ps1
@@ -1,7 +1,7 @@
 ﻿# Inspired by Mark Embling
 # http://www.markembling.info/view/my-ideal-powershell-prompt-with-git-integration
 
-$global:GitPromptSettings = [GitPromptSettings]::new()
+$global:GitPromptSettings = [PoshGitPromptSettings]::new()
 
 # Override some of the normal colors if the background color is set to the default DarkMagenta.
 $s = $global:GitPromptSettings
@@ -13,15 +13,19 @@ if ($Host.UI.RawUI.BackgroundColor -eq [ConsoleColor]::DarkMagenta) {
     $s.WorkingColor.ForegroundColor             = 'Red'
 }
 
-$isAdminProcess = Test-Administrator
-$adminHeader = if ($isAdminProcess) { 'Administrator: ' } else { '' }
+$IsAdmin = Test-Administrator
 
-$WindowTitleSupported = $true
-# TODO: Hmm, this is a curious way to detemine window title supported
-# Could do $host.Name -eq "Package Manager Host" but that is kinda specific
-# Could attempt to change it and catch any exception and then set this to $false
-if (Get-Module NuGet) {
-    $WindowTitleSupported = $false
+# Probe $Host.UI.RawUI.WindowTitle to see if it can be set without errors
+$WindowTitleSupported = $false
+try {
+    $global:PreviousWindowTitle = $Host.UI.RawUI.WindowTitle
+    $newTitle = "$origTitle "
+    $Host.UI.RawUI.WindowTitle = $newTitle
+    $WindowTitleSupported = $Host.UI.RawUI.WindowTitle -eq $newTitle
+    $Host.UI.RawUI.WindowTitle = $global:PreviousWindowTitle
+}
+catch {
+    Write-Debug "Probing for WindowTitleSupported errored: $_"
 }
 
 <#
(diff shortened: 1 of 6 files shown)


src/GitPrompt.ps1
```diff
@@ -13,15 +13,19 @@ if ($Host.UI.RawUI.BackgroundColor -eq [ConsoleColor]::DarkMagenta) {
...
-# TODO: Hmm, this is a curious way to detemine window title supported
-# Could do $host.Name -eq "Package Manager Host" but that is kinda specific
-# Could attempt to change it and catch any exception and then set this to $false
-if (Get-Module NuGet) {
-    $WindowTitleSupported = $false
+# Probe $Host.UI.RawUI.WindowTitle to see if it can be set without errors
+$WindowTitleSupported = $false
+try {
+    $global:PreviousWindowTitle = $Host.UI.RawUI.WindowTitle
+    $newTitle = "$origTitle "
```
dahlbyk: Should this be `"$PreviousWindowTitle "`?

src/PoshGitTypes.ps1
```diff
@@ -244,8 +244,9 @@ class GitPromptSettings {
     [Nullable[bool]]$EnableFileStatusFromCache        = $null
     [string[]]$RepositoriesInWhichToDisableFileStatus = @()
 
-    [string]$DescribeStyle       = ''
-    [psobject]$EnableWindowTitle = 'posh~git ~ '
+    [string]$DescribeStyle   = ''
+    [bool]$EnableWindowTitle = $true
```
dahlbyk: > Also, I think we might want to keep `EnableWindowTitle` as a simple bool. If you disable the WindowTitle by setting it to $null, you lose the original definition.

I'm not too worried about this, since resetting is as simple as opening a new shell. Fewer settings seems better than many.

src/GitPrompt.ps1
rkeithhill: Doh!  Yes it should be. Thanks.  I'll fix it tonight.

src/PoshGitTypes.ps1
rkeithhill: Well, we have reduced the number of "top-level" settings a lot.  :-)  But I don't feel that strongly about this.

dahlbyk:
> BTW, in the $GitStatus is null case as well as the case where the user's WindowTitle scriptblock/string generates an error, we fall back to the $global:PreviousWindowTitle.

Is that what we want, though? It seems more useful if this feature can be used to customize the window title at all times (e.g. showing `$PID`), not just in a posh-git context. Falling back to `$global:PreviousWindowTitle` should only be necessary if the scriptblock/string comes back errored or empty.

rkeithhill:
> Is that what we want, though?

I'm definitely open to making changes here.  It always seemed a bit odd to me that the title swapped around a lot between Git and not Git dirs.  It would complicate the default scriptblock a bit but I think that is OK.  I could see it being something like this:
```powershell
{
    param($GitStatus, [bool]$IsAdmin) 
    "$(if ($IsAdmin) {'Administrator: '})$(if ($GitStatus) {"posh~git ~ $($GitStatus.RepoName) [$($GitStatus.Branch)] "})PowerShell $(if ([IntPtr]::Size -eq 8) {'x64'} else {'x86'}) $($PSVersionTable.PSVersion) ($PID)"
} 
```
That gives title bar text like this outside a git repo (as admin):
```
Administrator: PowerShell x64 5.1.16299.98 (33624)
```
and inside a Git repo (not as admin):
```
posh~git ~ vscode-powershell [master] PowerShell x64 5.1.16299.98 (33624)
```
When we display the posh-git info, perhaps we need another separator before `PowerShell`:
```
posh~git ~ vscode-powershell [master] ~ PowerShell x64 5.1.16299.98 (33624)
```
Thoughts?

rkeithhill:
I removed the x64/x86 bit for now.

rkeithhill:
Thanks.  I think that is what folks will expect.  I can't think of many any real user scenarios where someone would want to set their title bar to `0` or `""` or `$false`.  And if they did, they can set  `$host.UI.RawUI.WindowTitle` directly.

dahlbyk:
Nice work! I'll see about back-porting for 0.7.2.

PR #547, opened 2018-02-05
Ensure origin exists for tests

dahlbyk:
Continuing on from #545 and #546, `develop` branch tests are now failing because `origin/master` doesn't exist. So let's ensure `origin` exists, just for good measure, then `origin/master` and `origin/HEAD`. Integration tests are such a pain... Wish we could adequately test this outside of `develop`!


rkeithhill:
Doh!  Thanks for fixing that.

PR #548, opened 2018-02-05
Cheat to ensure origin refs exist

dahlbyk:
#547 doesn't quite work due to how Travis clones:

```
git clone --depth=50 --branch=develop https://github.com/dahlbyk/posh-git.git dahlbyk/posh-git
```

This results in the following config setting:

```
remote.origin.fetch=+refs/heads/develop:refs/remotes/origin/develop
```

As opposed to the usual:

```
remote.origin.fetch=+refs/heads/*:refs/remotes/origin/*
```

In practice, this means `git fetch origin master` doesn't match a reference, so it's not fetched.

Rather than fix that, it seems easier to just create the refs the tests expect.

### DIFF
--- a/test/TabExpansion.Tests.ps1
+++ b/test/TabExpansion.Tests.ps1
@@ -11,7 +11,7 @@ Describe 'TabExpansion Tests' {
             # Ensure an origin remote exists
             &$gitbin remote add origin . 2>$null
             # Ensure origin/master exists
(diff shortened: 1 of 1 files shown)


dahlbyk:
@rkeithhill `develop` is finally green!

rkeithhill:
Yay!  Thanks for the fix on my attempt/PR on this.

PR #552, opened 2018-03-02
Fix bug w/setting fg/bg color to 0x0 & add regression test

rkeithhill:


rkeithhill:
@dahlbyk If you don't have time to review this one -real short that's OK.  I'll give it a day and then merge it.

PR #554, opened 2018-03-07
Add Get-GitRemotes function

hugo-vrijswijk:
I've added a function that returns an object with all git remotes and the url for each remote. 

The output of the function is as follows (with one remote): 

```
Name   Url
----   ---
origin git@github.com:hugo-vrijswijk/posh-git.git
```

The idea behind this function is that [I am working on a feature for an oh-my-posh theme that shows an icon for certain remotes](https://github.com/hugo-vrijswijk/oh-my-posh/tree/remote-icon), and there is no simple way to get the remote from posh-git. Result would be like this for GitHub:
image

Instead of having oh-my-posh query git, I thought it would be more sensible to extract the function to posh-git, so it could be reused should anyone see use in it.

I thought about adding it to `Get-GitStatus` but decided against it as it's not part of the default `git status` anyway

### DIFF
--- a/src/GitUtils.ps1
+++ b/src/GitUtils.ps1
@@ -405,3 +405,35 @@ function Update-AllBranches($Upstream = 'master', [switch]$Quiet) {
 
     git checkout -q $head
 }
+
+<#
+.SYNOPSIS
+    Gets a Git remotes object
+.DESCRIPTION
+    Gets a git remotes object
+    The object contains the name and URL of all remotes
+.EXAMPLE
+    PS C:\> $s = Get-GitRemotes
+    Returns: 
+        Name   Url
+        ----   ---
+        origin git@github.com:dahlbyk/posh-git.git
+.INPUTS
+    None
+.OUTPUTS
+    System.Management.Automation.PSMethod
+#>
+function Get-GitRemotes {
+    $remoteNames = git remote
+    $remotes = New-Object System.Collections.ArrayList
+    foreach ($remoteName in $remoteNames) {
+        $remoteUrl = git remote get-url $remoteName
+
+        $remote = New-Object System.Object
+        $remote | Add-Member -MemberType NoteProperty -Name "Name" -Value $remoteName
+        $remote | Add-Member -MemberType NoteProperty -Name "Url" -Value $remoteUrl
+
+        $remotes.Add($remote) | Out-Null
+    }
+    $remotes
+}
--- a/src/GitUtils.ps1
+++ b/src/GitUtils.ps1
@@ -414,7 +414,7 @@ function Update-AllBranches($Upstream = 'master', [switch]$Quiet) {
     The object contains the name and URL of all remotes
 .EXAMPLE
     PS C:\> $s = Get-GitRemotes
-    Returns: 
+    Returns:
         Name   Url
         ----   ---
         origin git@github.com:dahlbyk/posh-git.git
@@ -424,10 +424,10 @@ function Update-AllBranches($Upstream = 'master', [switch]$Quiet) {
     System.Management.Automation.PSMethod
 #>
 function Get-GitRemotes {
-    $remoteNames = git remote
+    $remoteNames = Invoke-Utf8ConsoleCommand { git remote }
     $remotes = New-Object System.Collections.ArrayList
     foreach ($remoteName in $remoteNames) {
-        $remoteUrl = git remote get-url $remoteName
+        $remoteUrl = Invoke-Utf8ConsoleCommand { git remote get-url $remoteName }
 
         $remote = New-Object System.Object
         $remote | Add-Member -MemberType NoteProperty -Name "Name" -Value $remoteName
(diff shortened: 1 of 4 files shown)


rkeithhill:
The primary intent of this project is to support the use of git.exe by providing tab completion for the native executable and to provide Git status summary info integrated into the prompt.  In general, we are not trying to present "PowerShell oriented" versions of git commands except where we need them for support of the prompt and tab-completion.  

Now, if the proposal were to display extra info in the posh-git prompt function, then the addition of this function makes more sense.  Note that we don't have to enable display of "remote info" by default.  It could be opt-in.  I don't think we could rely on PowerLine fonts so I'm not sure what we would display - perhaps just text? GH for GitHub, BB for BitBucket, ??  Ultimately this is @dahlbyk 's call.

hugo-vrijswijk:
Hmm, I see your point about just translating git to powershell commands. 

I am not sure about displaying remote info by default. There definitely should be options, IMO. Especially because not all users have the correct PowerLine fonts installed and it would just look buggy. It could be off by default, and when enabled a toggle for PowerLine icons or text.

How would you feel about having the option for it in the git prompt, but also opening a function for the symbol so themes like oh-my-posh can utilize it?

rkeithhill:
I'm open to the idea but in theory you should be able to do this all from oh-my-posh. Just insert the proper glyph before the branch symbol based on your `Get-GitRemote` function.  Let's see what @dahlbyk thinks?

src/GitUtils.ps1
```diff
@@ -405,3 +405,35 @@ function Update-AllBranches($Upstream = 'master', [switch]$Quiet) {
...
+    Returns:
+        Name   Url
+        ----   ---
+        origin git@github.com:dahlbyk/posh-git.git
+.INPUTS
+    None
+.OUTPUTS
+    System.Management.Automation.PSObject
+#>
+function Get-GitRemotes {
```
rkeithhill: PowerShell usually uses singular nouns e.g Stop-Process not Stop-Processes.  I would recommend changing this to `Get-GitRemote`.

src/GitUtils.ps1
```diff
@@ -405,3 +405,35 @@ function Update-AllBranches($Upstream = 'master', [switch]$Quiet) {
...
+    $remoteNames = Invoke-Utf8ConsoleCommand { git remote }
+    $remotes = New-Object System.Collections.ArrayList
```
rkeithhill: No need to create an ArrayList here - just output each remote in the foreach loop.

src/GitUtils.ps1
```diff
@@ -405,3 +405,35 @@ function Update-AllBranches($Upstream = 'master', [switch]$Quiet) {
...
+    foreach ($remoteName in $remoteNames) {
+        $remoteUrl = Invoke-Utf8ConsoleCommand { git remote get-url $remoteName }
+
+        $remote = New-Object System.Object
+        $remote | Add-Member -MemberType NoteProperty -Name "Name" -Value $remoteName
+        $remote | Add-Member -MemberType NoteProperty -Name "Url" -Value $remoteUrl
+
+        $remotes.Add($remote) | Out-Null
```
rkeithhill: Just change this to `$remote`  which is the same as `Write-Output $remote`.

src/GitUtils.ps1
```diff
@@ -405,3 +405,35 @@ function Update-AllBranches($Upstream = 'master', [switch]$Quiet) {
...
+    }
+    $remotes
```
rkeithhill: Don't need this.

src/GitUtils.ps1
bergmeister: I totally agree on this as well that singular nouns should be used.

src/GitUtils.ps1
bergmeister: Also, experience has shown that the `$null = ...` pattern is better than piping to `Out-Null`.

dahlbyk:
> The primary intent of this project is to support the use of git.exe by providing tab completion for the native executable and to provide Git status summary info integrated into the prompt. In general, we are not trying to present "PowerShell oriented" versions of git commands except where we need them for support of the prompt and tab-completion.

@rkeithhill is correct - we're trying to stay focused here, as there are efforts elsewhere to provide a more native PowerShell interface for Git.

> Now, if the proposal were to display extra info in the posh-git prompt function, then the addition of this function makes more sense. Note that we don't have to enable display of "remote info" by default. It could be opt-in. I don't think we could rely on PowerLine fonts so I'm not sure what we would display - perhaps just text? GH for GitHub, BB for BitBucket, ?? Ultimately this is @dahlbyk 's call.

I'm not strictly opposed to encoding the remote host, though I don't personally care. But we wouldn't want to use `Get-GitRemotes` when we're only interested in a single remote.

That said, we do appreciate the propsal. If you (or anyone) does think of a clever way to encode remote details in our default prompt, I'd be curious to see it.

Issue #556, opened 2018-03-22
Prompt will not display untracked files if status.showUntrackedFiles is set to "no"


TheSned:
### System Details

- posh-git version/path: 0.7.2.0 pre0 D:\Documents\WindowsPowerShell\Modules\posh-git
- PowerShell version: 5.1.16299.251
- git version 2.16.1.windows.4
- OS: Microsoft Windows NT 10.0.16299.0

### Issue Description

When the status.showUntrackedFiles setting is set to "no", the posh-git prompt will not display untracked files either. It seems like this should either be a setting or should append -uall to the git status command to always include untracked files.

### Steps to Reproduce

git init StatusBug
cd StatusBug
git config --local status.showUntrackedFiles no
"Hello, world!" > untracked.txt
Write-VcsStatus

PR #557, opened 2018-03-22
Update git status to always show untracked files.

TheSned:
Fixes dahlbyk/posh-git#556.

This assumes we want to *always* show untracked files rather than providing an option to include untracked files or not. If this is not the preferred way to fix this issue, let me know what is desired and I can probably implement it.

### DIFF
--- a/src/GitPrompt.ps1
+++ b/src/GitPrompt.ps1
@@ -87,6 +87,9 @@ $global:GitPromptSettings = [pscustomobject]@{
 
     ShowStatusWhenZero                          = $true
 
+    # Valid values are "all", "no", and "normal"
+    UntrackedFilesMode                          = $null
+
     AutoRefreshIndex                            = $true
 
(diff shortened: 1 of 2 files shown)


dahlbyk:
I do think it would be a good idea to make this configurable. Let's add `$GitPromptSettings.UntrackedFilesMode` to allow providing a `-u` value, defaulting to `$null` to preserve existing behavior. If a value is provided (`all`, `normal` or `no`), it should be added to the `git status` call. Does that seem reasonable?

Thanks for the proposal, and sorry for the absurd delay in reviewing this!

TheSned:
Yes, that sounds reasonable. I'll update it when I get a chance.

dahlbyk:
Great - thanks!

rkeithhill:
Do we have a way to tag this commit for merging to `develop`?

PR #563, opened 2018-04-17
Export Expand-GitCommand

### LINKED ISSUE #562: [Request] Export GitTabExpansion
I've written a tiny module to configure aliases for git subcommands to omit typing `git`.

https://gitlab.com/igrep/pwsh-git-aliases/blob/master/GitAliases.psm1

Then, I want to configure `TabExpansion` for the generated aliases by reusing the `GitTabExpansion` function.
So can you export `GitTabExpansion` to make it reusable?

P.S. Thank you for an excellent module! I use it every day!

dahlbyk:
Close #562

### DIFF
--- a/src/GitTabExpansion.ps1
+++ b/src/GitTabExpansion.ps1
@@ -224,8 +224,8 @@ function script:expandParamValues($cmd, $param, $filter) {
         ForEach-Object { -join ("--", $param, "=", $_) }
 }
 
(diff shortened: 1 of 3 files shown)


rkeithhill:
This is another that needs to be merged to `develop`, right?

dahlbyk:
> This is another that needs to be merged to `develop`, right?

Correct. It's been a while since `master` has been merged into `develop`.

PR #564, opened 2018-04-17
Implement $GitPromptSettings.UntrackedFilesMode in the develop branch.

TheSned:


dahlbyk:
Looks good! If you are bored (or anyone else), it would be neat to have tests for this, too.

PR #567, opened 2018-04-20
Shorten window title text

### LINKED ISSUE #565: Consider shorter WindowTitle for when we get a tabbed console in Windows 10 1809
With a [tabbed console](https://blogs.msdn.microsoft.com/commandline/2018/04/13/tabbed-console-starts-here/) coming:

image

We should rethink our WindowTitle in 1.0.0 a bit:

image

I like that we put the PowerShell version/arch/pid info to the right.  Folks should be able to hover over the tab and get a tooltip with that info.  However the `posh~git ~` will take up valuable, visible tab space.  We should think about nuking that prefix and go with just:

image

Another thought - with a tabbed interface like this it might be handy to display the current working dir in the title but only when *not* in a repo e.g.:

image

Finally, I've been thinking about tweaking the `Get-PromptPath` function so that when `DefaultPromptAbbreviateHomeDirectory` is $true and you are in your home dir, you get the full path e.g. `C:\Users\Keith`.  It is only when you go down a level that you get the `~` e.g. `~\GitHub\dahlbyk\posh-git`.  That would turn this somewhat odd looking prompt:
```
~
04-18 21:17:58 33ms 9>
```
into
```
C:\Users\Keith
04-18 21:17:58 33ms 9>
```
but still give me a shortened path in this case:
```
~\GitHub\dahlbyk\posh-git\src
04-18 21:17:58 33ms 9>
```

rkeithhill:
Modify Get-PromptPath to return full dir when path is equal to $Home.
It will use ~ when path is a subdir *under* $HOME.

This does not address other path shortening techniques.
We could do those in this PR or another one.

Fixes #565 

Also I think this supersedes PR #538 for the 1.x branch.

### DIFF
--- a/src/PoshGitTypes.ps1
+++ b/src/PoshGitTypes.ps1
@@ -271,7 +271,7 @@ class PoshGitPromptSettings {
     [string[]]$RepositoriesInWhichToDisableFileStatus = @()
 
     [string]$DescribeStyle = ''
-    [psobject]$WindowTitle = {param($GitStatus, [bool]$IsAdmin) "$(if ($IsAdmin) {'Admin: '})$(if ($GitStatus) {"$($GitStatus.RepoName) [$($GitStatus.Branch)]"} else {Get-PromptPath}) - PowerShell $($PSVersionTable.PSVersion) $([IntPtr]::Size * 8)-bit ($PID)"}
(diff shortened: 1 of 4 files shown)


rkeithhill:
Working on the test failures.

dahlbyk:
Please rebase on latest and squash your test fixes.

rkeithhill:
Hopefully this will merge with the branch name change.  :-0

rkeithhill:
Oh yeah, was that because of some conemu issue?

rkeithhill:
Also, can we merge this then?  :-)

dahlbyk:
> Oh yeah, was that because of some conemu issue?

Started that way, yeah: https://github.com/dahlbyk/posh-git/commit/1ed6d4d3c26059c5b0abbd33708dde2be70ad8f2

GSGBen:
Is there a way to disable this? I'd prefer ~ when in the root.

PR #542, opened 2018-01-30
Add 1.0.0-beta2 section to CHANGELOG.md

rkeithhill:
Clean up various markdown linter warnings.

Let's use this PR for on-going tweaks to the CHANGELOG.md. Although I'm not anticipating making any more changes myself.  Although I need to wrap up the README changes and tweak the Wiki on customizing your prompt function.

Should we shoot for a Feb 14th release of beta 2? I think we could easily pull that up **if** we wanted to.

PR #521, opened 2018-01-10
Update CHANGELOG for 1.0.0-beta1 release

rkeithhill:
I'm presuming beta1 but I think that is the appropriate moniker.  I think we're further along than an alpha.

I guess the README.md update will come later.  Question is, should I update the `master` readme since that is the default readme folks will see?  Or should we rely on folks to switch to the develop branch.  For sure, the README on develop should be updated.

Issue #467, opened 2017-04-15
Module no-longer loads after #376 (manifest moved  to src/ directory).


bgshacklett:
### System Details

- posh-git version/path:
   - 0.6.1.20160330
   - C:\Users\bria0265\Documents\WindowsPowerShell\Modules\Posh-Git\posh-git.psm1
- PowerShell version: 5.1.14393.693
- Git version: git version 2.9.0.windows.1
- Operating system name and version: 10.0.14393 N/A Build 14393

### Issue Description

My method of installation is to add this repository as a submodule under my PowerShell profile's Modules directory and source the example profile. This has worked great for a significant period of time. Unfortunately, after updating to 0.7.1, I found that the  module was no-longer available for import. I see that #376 moved the module manifest to the src/ directory which appears to have caused the problem. Has there been any discussion about this behavior? While I can certainly pull this module out of the PowerShell gallery, being able to just perform a recursive clone of my profile made things very simple and quick to get new workspaces set up.

Thank you for your time.

rkeithhill:
The issue is separating the artifacts that ship as part of the module from the artifacts that don't ship.  The files under `src` are what get's published - for now.  Eventually, if we have generated files i.e. signed scripts, module file catalog or platyPS generated help files, then the module publishing will be done from another, output dir after the module is `built`.  

The preferred mechanism for using the module is to install it from the PowerShell Gallery via Install-Module.  If you are developing for posh-git, then import the module by path.  This is what I put in my profile:
```posh
Import-Module ~\GitHub\dahlbyk\Posh-Git\src\posh-git.psd1
```

dahlbyk:
It's worth noting that I ran into this [here](https://github.com/dahlbyk/posh-git/issues/391#issuecomment-276732276), and at the time it seemed easiest to just work around it. But then I hit it again [here](https://github.com/dahlbyk/posh-git/pull/400#issuecomment-280065698) while publishing to the PowerShell Gallery.

As @rkeithhill notes, you should still be able to `Import-Module ~\Documents\WindowsPowerShell\Modules\Posh-Git\src\posh-git.psd1` directly, but that rather defeats the point of living in `Modules`.

[...]

PR #577, opened 2018-05-14
Release v1.0.0-beta2

dahlbyk:
Following on from #542, I've tagged v1.0.0-beta2 and released to PowerShell Gallery. I'm trying to remember if there was a reason I didn't publish v1.0.0-beta1 to the Chocolatey Gallery as well...? Anyway, I'm holding off on that until I convince myself it's not an issue. Maybe I was thinking we shouldn't publish anything 1.0 to Chocolatey until we know how 1.0 is going to work with Chocolatey? Seems reasonable...

@rkeithhill since you tend to kick off the release process with a PR to update the changelog, feel free to name those as release branches and we can do this all at once (e.g. beta1 was tagged in #521).

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,8 @@ You can also tab complete remote names and branch names e.g.: `git pull or<tab> 
 
 #### Releases
 
+- v1.0.0-beta2
+  ( [README][v1b2-readme] • [CHANGELOG][v1b2-change] )
 - v1.0.0-beta1
   ( [README][v1b1-readme] • [CHANGELOG][v1b1-change] )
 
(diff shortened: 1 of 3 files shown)


rkeithhill:
> feel free to name those as release branches and we can do this all at once

Not sure I follow completely.  Do you mean I should name the next CHANGELOG "prep-for-release" branch: `release-v1.0.0-beta3` (assuming the next milestone is beta3)?

dahlbyk:
> Do you mean I should name the next CHANGELOG "prep-for-release" branch: release-v1.0.0-beta3 (assuming the next milestone is beta3)?

Yes!

And I do think we'll want a beta3. My goals for that release are:

1. Figure out v1 Chocolatey strategy
2. Drop ssh-agent stuff into separate module.
3. Maybe move posh-git code into `src/posh-git/`? PowerShell still really doesn't like modules living in folders that don't match their name; I currently rename `src/` to `posh-git/` every time I publish to PS Gallery. A build process (#518) could also be the answer to that, but I do think it's important that a fresh clone works as expected (#467).

Issue #473, opened 2017-06-01
Avoiding creating a C:\tools directory


jnm2:
Is it possible to move `C:\tools\poshgit` to `%appdata%\poshgit`?

I'm using the Chocolatey package.

dahlbyk:
In the short term, you can change that path with `$env:ChocolateyToolsLocation`. (I would uninstall first.) See [`Get-ToolsLocation`](https://github.com/chocolatey/choco/blob/master/src/chocolatey.resources/helpers/functions/Get-ToolsLocation.ps1) for details.

Beyond that, there's some discussion over in https://github.com/dahlbyk/posh-git/issues/328 about the future of distributing posh-git with Chocolatey. That may be a while yet, though, so I would not be opposed to switching from `Get-ToolsLocation` to installing posh-git elsewhere. Rather than `%appdata%`, I'm inclined to follow the suggestion from [this comment](https://chocolatey.org/packages/poshgit#comment-2816165073) of installing to a location that's discoverable by `Get-Module -ListAvailable` (i.e. a `posh-git` folder inside the first path from `$Env:PSModulePath`).

If you, or anyone, does decide to tackle this, keep in mind that the uninstaller should continue to correctly purge posh-git from `Get-ToolsLocation` as well.

carlomorelli:
Hi,
about the Chocolatey deliverable that installs in C:\tools.

I would ask to make an example out of InvokeBuild package: https://chocolatey.org/pack.... This is a build tool that I use at work, in the form of Powershell module. It installs correctly in the conventional directories where Powershell modules are discoverable.

These directories can be:
```
C:\Users\<user>\Documents\WindowsPowerShell
C:\Program Files\WindowsPowerShell
C:\Windows\System32\WindowsPowerShell\v1.0 (this is reserved to older scripts I believe)
```

Please check out this document from MSDN: https://msdn.microsoft.com/en-us/library/dd878350(v=vs.85).aspx#Anchor_1

To check if the module is correctly installed in the system, it should appear in `Get-Module -ListAvailable` command output.

The use case for this is that installing a module in the correct directories is important if you use 3rd party software that depends on these modules, as it wouldn't be discoverable by them otherwise.

For example I use the console emulator Cmder, which supposedly integrates with posh-git when emulating a Powershell terminal in a Git directory. 
The integration works if you install posh-git with PowershellGet (Install-Module posh-git) but gives constant warnings ("the module cannot be located") if you install posh-git via Chocolatey.

[...]

PR #578, opened 2018-05-14
Install module to 'Program Files' and remove modifying $PROFILE

pauby:
This is the Chocolatey package install / uninstall files I use which copies the embedded module to the `Program Files\WindowsPowerShell\Modules` folder. 

The module itself is embedded into the package for easier distribution and I do this using an automatic update script. As you're not using that then simply add the code you want to distribute into a folder called `posh-git` within the tools folder. Inside the `posh-git` folder should be the code itself. The installer script takes care of creating a version folder if needed for PS 5+.

See my other modules for PSScriptAnalyzer, PSCodeHealth, Plaster, PlatyPS, PSStringTemplate and EditorServicesCommandSuite [here](https://github.com/pauby/ChocoPackages/tree/master/automatic) for more info.

### DIFF
--- a/chocolatey/tools/chocolateyInstall.ps1
+++ b/chocolatey/tools/chocolateyInstall.ps1
@@ -1,49 +1,35 @@
-﻿try {
-    $poshgitPath = join-path (Get-ToolsLocation) 'poshgit'
+﻿$ErrorActionPreference = 'Stop'
 
-    try {
-      if (test-path($poshgitPath)) {
-        Write-Host "Attempting to remove existing `'$poshgitPath`'."
-        remove-item $poshgitPath -recurse -force
-      }
-    } catch {
-      Write-Host "Could not remove `'$poshgitPath`'"
-    }
+$toolsDir = "$(Split-Path -parent $MyInvocation.MyCommand.Definition)"
+$moduleName = 'posh-git'  # this may be different from the package name and different case
 
-    $version = "v$Env:chocolateyPackageVersion"
-    if ($version -eq 'v') { $version = 'master' }
-    $poshGitInstall = if ($env:poshGit ) { $env:poshGit } else { "https://github.com/dahlbyk/posh-git/zipball/$version" }
-    $zip = Install-ChocolateyZipPackage 'poshgit' $poshGitInstall $poshgitPath
-    $currentVersionPath = Get-ChildItem "$poshgitPath\*posh-git*\" | Sort-Object -Property LastWriteTime | Select-Object -Last 1
+# module may already be installed outside of Chocolatey
+Remove-Module -Name $moduleName -Force -ErrorAction SilentlyContinue
 
-    if(Test-Path $PROFILE) {
-        $oldProfile = @(Get-Content $PROFILE)
+$sourcePath = Join-Path -Path $toolsDir -ChildPath "$modulename\*"
+$destPath = Join-Path -Path $env:ProgramFiles -ChildPath "WindowsPowerShell\Modules\$moduleName"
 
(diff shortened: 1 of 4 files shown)


dahlbyk:
I'll take a closer look at your new code later, but for now:

- 👍 for new location
- 👍 for including only relevant files
- It seems like we should try to clean up the legacy `$poshgitPath` on install to the new path, as we currently do on uninstall.
- Similarly, if we can detect that this package has previously modified someone's `$PROFILE` with a reference to `$poshgitPath` (and other legacy cruft), we should continue to clean up after ourselves.
- As I mentioned in https://github.com/dahlbyk/posh-git/issues/473#issuecomment-388987308, thoughts on how we could let the user opt out of modifying profile? Or how we could prompt if they wish us to add to profile?
  - We could also reasonably let them specify which profile to modify, assuming someone gets around to fixing https://github.com/chocolatey/choco/issues/1154.

chocolatey/tools/chocolateyInstall.ps1
```diff
@@ -1,49 +1,35 @@
...
-    $installer = Join-Path $currentVersionPath 'install.ps1'
-    & $installer
-} catch {
-  try {
-    if($oldProfile){ Set-Content -path $PROFILE -value $oldProfile -Force -Encoding $oldProfileEncoding }
-  }
-  catch {}
-  throw
+if ($PSVersionTable.PSVersion.Major -lt 4) {
+    $modulePaths = [Environment]::GetEnvironmentVariable('PSModulePath', 'Machine') -split ';'
```
rkeithhill: Can we assume no one will try to use Chocolatey to install this on Linux or macOS?  If not, you should use `[System.IO.Path]::PathSeparator` instead of hard-coding in `;`.

chocolatey/tools/chocolateyInstall.ps1
```diff
@@ -1,49 +1,35 @@
...
-        Set-Content -path $profile -value $newProfile -Force -Encoding $oldProfileEncoding
-    }
+$toolsDir = "$(Split-Path -parent $MyInvocation.MyCommand.Definition)"
+$moduleName = 'posh-git'  # this may be different from the package name and different case
+
+# module may already be installed outside of Chocolatey
+Remove-Module -Name $moduleName -Force -ErrorAction SilentlyContinue
+
+$sourcePath = Join-Path -Path $toolsDir -ChildPath "$modulename\*"
+$destPath = Join-Path -Path $env:ProgramFiles -ChildPath "WindowsPowerShell\Modules\$moduleName"
```
rkeithhill: I thought there was Chocolatey support for PowerShell Core?  If so, we need to detect if we are on PS Core and if so, we should probably also install to `"PowerShell\Modules\$moduleName"` if the `$env:ProgramFiles\PowerShell` dir exists.

rkeithhill:
RE:
> for including only relevant files

Don't mean to sound like a broken record but ... with a build script - or at least a publish script - we could copy the relevant module files for distribution into an out dir.  Which I have proposed would be `<repo-root>/<version-number>`.  :-)

chocolatey/tools/chocolateyInstall.ps1
pauby: I agree. I have been wondering what to do about whether you run Chocolatey under Core or not and where you should put the modules. However this is purely for the v0 branch so PS Core wasn't a consideration.

chocolatey/tools/chocolateyInstall.ps1
pauby: For the time being we can. As per the next comment this is only for v0 so we don't need to worry about PS Core.

Issue #549, opened 2018-02-10
Add support for vsts-cli Git integration


flcdrg:
The [vsts-cli](https://github.com/Microsoft/vsts-cli) tool can integrate with git, providing aliases such as  `git pr` and `git repo`.

It would be good to add support for this to posh-git.

(I'm working on a PR for this)

rkeithhill:
This support will be available in 1.0.0-beta3.

PR #581, opened 2018-05-19
Initial support for vsts-cli

flcdrg:
- Add all `pr` commands and short commands (equivalent to `vsts code pr`)
- Add test for filtering params

Implements #549

Not implemented in this pull request (future work):
- 'subgroups' (work-items set-vote policies)
- other `vsts code` operations (eg. `repo`)

### DIFF
--- a/src/GitTabExpansion.ps1
+++ b/src/GitTabExpansion.ps1
@@ -8,6 +8,7 @@ $Global:GitTabSettings = New-Object PSObject -Property @{
 $subcommands = @{
     bisect = "start bad good skip reset visualize replay log run"
     notes = 'add append copy edit get-ref list merge prune remove show'
+    pr = 'create update show list complete abandon reactivate reviewers work-items set-vote policies'
     reflog = "show delete expire"
     remote = "
         add rename remove set-head set-branches
@@ -56,6 +57,8 @@ $script:someCommands = @('add','am','annotate','archive','bisect','blame','branc
 $script:gitCommandsWithLongParams = $longGitParams.Keys -join '|'
 $script:gitCommandsWithShortParams = $shortGitParams.Keys -join '|'
 $script:gitCommandsWithParamValues = $gitParamValues.Keys -join '|'
+$script:vstsCommandsWithShortParams = $shortVstsParams.Keys -join '|'
+$script:vstsCommandsWithLongParams = $longVstsParams.Keys -join '|'
 
 try {
     if ($null -ne (git help -a 2>&1 | Select-String flow)) {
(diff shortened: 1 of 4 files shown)


dahlbyk:
Neat! It's not obvious why builds are failing; do they pass on your machine?

flcdrg:
They did. 

I liked the fact that you had Pester there (I should make more use of it elsewhere!)

flcdrg:
Hmm.. tests definitely failing on those builds though. I'll see if I can reproduce the problem locally. Is there any tricks to testing posh-git?

flcdrg:
For some reason, on the CI builds, the $result is empty for my new tests, but not having any success replicating that locally

flcdrg:
Totally fine with your changes 😃 

I tried to fit in with the existing structure and make the smallest change possible to get the subcommands working as I could.

Making a more general solution totally makes sense. I'd guess the design for that could start to become evident once there's one or two more extensions that want to be included.

One of the areas I skipped for this PR was the vsts 'subgroup' commands, which I think end up a bit like sub-sub-sub commands.

Another thing for the future is to add value completion for the boolean and 'enum' type parameters.

But baby steps - get something basic working and then I can build on that!

dahlbyk:
> Making a more general solution totally makes sense. I'd guess the design for that could start to become evident once there's one or two more extensions that want to be included.

It's an interesting design process, for sure. I'd almost argue that posh-git isn't even the right place to solve this problem for `vsts`. Maybe better to hook into a posh-vsts™ module's `Expand-VstsCommand` for relevant aliases.

Out of curiosity, is there any overlap between data available from `vsts` and opportunities to improve git-tfs expansion support? I don't use either.

> But baby steps - get something basic working and then I can build on that!

💯

I'm going to give this a day or two to see if @rkeithhill or anyone else weighs in.

src/GitTabExpansion.ps1
```diff
@@ -264,6 +276,23 @@ function GitTabExpansionInternal($lastBlock, $GitStatus = $null) {
 
     switch -regex ($lastBlock -replace "^$(Get-AliasPattern git) ","") {
 
+        # Handles git pr alias
```
rkeithhill: It's not a big perf hit but it strikes me that these "case" statements are less likely to be encountered than the basic `git <cmd>` completions.  As such, it seems like the new vsts-oriented "case" statements should go towards the bottom after the more common cases.

rkeithhill: Other than the above, this PR LGTM.

flcdrg:
> Out of curiosity, is there any overlap between data available from vsts and opportunities to improve git-tfs expansion support? I don't use either.

I'm not sure. git-tfs is a bridge like git-svn. I used it a few years back against an on-prem TFS server. It doesn't have the concept of pull requests, so I don't think there's much in common. It seems the focus of vsts-cli is git operations rather than TFVC.

dahlbyk:
Thanks!

rkeithhill:
Thanks!

PR #585, opened 2018-06-14
Removes ssh utils

JeremySkinner:
Removes ssh utils and associated tests.


dahlbyk:
I'll take a closer look later today, but an immediate observation: what if we moved the `posh-git` module into `src/posh-git/` and SSH module into `src/posh-ssh/`? I think this would fix #467.

dahlbyk:
> what if we moved the posh-git module into `src/posh-git/` and SSH module into `src/posh-ssh/`?

Nevermind. I really don't think we want to invest anything in making the SSH organization better here.

chocolatey/tools/chocolateyUninstall.ps1
```diff
@@ -13,7 +13,6 @@
         foreach($line in $oldProfile) {
             if ($line -like '*PoshGitPrompt*') { continue; }
             if ($line -like '*Load posh-git example profile*') { continue; }
-            if ($line -like '*Start-SshAgent*') { continue; }
```
dahlbyk: Hrm... I was going to say we should leave this, to clean up after old versions, but this uninstall script will only be used for 1.0+ which will have no responsibility for `Start-SshAgent`...so I think it's fine to go? Thoughts?

dahlbyk:
Still pondering this. I wonder if it's worth making any SSH fixes at all here in this repo, rather than just spawning a new repo?

From a preservation-of-history standpoint, perhaps the easiest path forward would be to actually make posh-ssh a fork of the posh-git repo at whatever point we consider the v0 fixes "done". Delete all the Git-only files, rename/prune the module, new readme/changelog, fix up some code and we're good to go. Thoughts?

chocolatey/tools/chocolateyUninstall.ps1
JeremySkinner: Yes good idea.

JeremySkinner:
I like that idea, preserving the history is good.

JeremySkinner:
@dahlbyk I've removed the commits that create posh-ssh and squashed the destructive ones. I think this should be good to go - let me know if you'd like anything changed.

JeremySkinner:
@dahlbyk Is there anything else you'd like me to do for this, or are you happy for me to merge it to master?

Issue #583, opened 2018-05-23
Windows 10 version 1803 broke my ssh-agent


holytshirt:
### System Details

- posh-git version/path: 0.7.3.1  C:\tools\poshgit\dahlbyk-posh-git-9bda399\src
- PowerShell version: 5.1.17134. 48
- Git version: git version 2.17.0.windows.1
- Operating system name and version: Win32NT   10.0.17134.0 Microsoft Windows NT 10.0.17134.0
- SSH version: OpenSSH_for_Windows_7.6p1, LibreSSL 2.6.4 

### Issue Description

When Windows 10 updates to 1803, SSH is now native on windows and SSH-AGENT is now a windows service.
Posh-Git only currently works with SSH installed as part of the Git tools.

The effect is Start-SshAgent -Quiet will ask you to enter a password every time you start a console session and git will ask you to enter your password as the ssh-agent's are conflicting.

If you need to get git working with the new SSH 
http://adamralph.com/2018/05/15/windows-10-version-1803-broke-my-git-ssh/

Fun times :)

rkeithhill:
Yup, see the latest on #575  Actually, I prefer the new ssh support as I can use VS now without having to start it from PowerShell.  We're also considering removing the ssh support in posh-git to another module.

rkeithhill:
On a system like Windows 10 1803, with OpenSSH installed or earlier version of Windows can install it from [here](https://github.com/PowerShell/openssh-portable/releases).  I'm thinking maybe we just remove the helpers.  

The helpers are pretty simply replaced on such systems:
```
Add-SshKey    -> ssh-add
Get-SshAgent   -> Get-Process ssh-agent | Foreach Id
Get-SshPath     -> ~\.ssh\<file>
Start-SshAgent  -> Start-Service ssh-agent
Stop-SshAgent  -> Stop-Service ssh-agent
```

myty:
I was getting this message and I went into look at the local services and the actual "OpenSSH Authentication Agent" service was disabled.  Maybe the 1803 update disabled it, but I'm not sure.  I set the service to automatic and now all is fine.

rkeithhill:
On my one test PC, every single, new Windows 10 Insiders build, disables the service so I have to configure/start it after each install.  :-(

JeremySkinner:
Opened #586 to address this.

JeremySkinner:
@dahlbyk should this now be under the 0.8 milestone?

sebid:
[...]

Issue #575, opened 2018-05-09
Get-SshAgent return always 0


nlhommet:
### System Details

- posh-git version/path:  `0.7.3` /  `C:\Program Files\WindowsPowerShell\Modules\posh-git`
- PowerShell version: `5.1.17134 revision 48`
- Git version: `version 2.17.0.windows.1`
- Operating system name and version: windows 10, version 1803

### Issue Description

Since last update of windows (1803), `Start-SshAgent` seems not working.

My profile.ps1
```
Import-Module "posh-git"
Start-SshAgent -Quiet
$ChocolateyProfile = "$env:ChocolateyInstall\helpers\chocolateyProfile.psm1"
if (Test-Path($ChocolateyProfile)) {
  Import-Module "$ChocolateyProfile"
}

Import-Module "Oh-My-Posh"
```

`Get-SshAgent` return always 0 and Git keeps prompting me for password.

rkeithhill:
From PowerShell, what path does this command return `Get-Command ssh-agent.exe`?

nlhommet:
```
CommandType     Name                                               Version    Source
-----------     ----                                               -------    ------
Application     ssh-agent.exe                                      7.6.0.0    C:\WINDOWS\System32\OpenSSH\ssh-agent.exe
````

**Comment by dahlbyk** [comment]
What happens if you run `ssh-agent` without the posh-git `Start-SshAgent`? The wrapper is really old, meant for the msys agent that ships with Git, not newer OpenSSH.

**Comment by nlhommet** [comment]
When I run `ssh-agent` nothing appear.
`Get-Service -Name 'ssh-agent'` return 
```
Status   Name               DisplayName
------   ----               -----------
Running  ssh-agent          OpenSSH Authentication Agent
```

dahlbyk:
That seems promising. The other thing `Start-SshAgent` does is `ssh-add`, which should add private key identities from `~/.ssh/`.

nlhommet:
`ssh-add -L` return my private key `ssh-rsa AAAAB...G1Q== ~/.ssh/id_rsa`

dahlbyk:
But Git is still prompting you for your password?

I honestly have no idea if Git and Win32-OpenSSH are expected to play nicely together, or how to fix. I'll drop a note over in [Git for Windows Gitter](https://gitter.im/git-for-windows/git) to see if someone can help.

nlhommet:
[...]

PR #586, opened 2018-06-14
Add support for win10 openssh

JeremySkinner:
@dahlbyk @rkeithhill  As discussed, adds support for the windows 10 native openssh agent (#583) 

- `Get-NativeSshAgent` determines if the Win32-OpenSSH client is installed and the agent service exists
- `Start-NativeSshAgent` will start the service if it's running (and the user is administrator)
- `Start-SshAgent` now calls into `Start-NativeSshAgent` and does no further processing if it returns true.
- `Stop-SshAgent` stops the agent service (if the user is admin)
- If we're using the native agent, then `Add-SshKey` checks to see if keys have already been added as each time you call ssh-add.exe, it'll prompt for the key password again (even if it's already been added)

I didn't want to change `Start-SshAgent` too much as master has already diverged from v0, which is why I added the new function.

A couple of questions that need some thought:
- Is it safe to check if the agent is empty by comparing the output of `ssh-add -L` to `"The agent has no identities."`? Is that string ever likely to be localized?
- If the service can't be started because the user isn't an admin, do we want to display a message or throw an exception?
- Do we want to provide a way to override this? For example if the user wants to force use of ssh.exe that's part of git for windows? At the moment the only way to override this would be editing PATH. 

Please take a look and let me know if you'd like me to change anything.

### DIFF
--- a/src/GitUtils.ps1
+++ b/src/GitUtils.ps1
@@ -416,19 +416,20 @@ function Get-SshAgent() {
 }
 
 function Get-NativeSshAgent {
-    # The ssh.exe binary version must include "OpenSSH"
-    # The windows ssh-agent service must exist
-    $service = Get-Service ssh-agent
-    $executableMatches = Get-Command ssh.exe | ForEach-Object FileVersionInfo | Where-Object ProductVersion -match OpenSSH
-    $valid = $service -and $executableMatches
-
-    if ($valid) {
-        return $service;
+    # $IsWindows is defined in PS Core. 
+    if (($PSVersionTable.PSVersion.Major -lt 6) -or $IsWindows) {
+        # The ssh.exe binary version must include "OpenSSH"
+        # The windows ssh-agent service must exist
+        $service = Get-Service ssh-agent -ErrorAction Ignore
+        $executableMatches = Get-Command ssh.exe | ForEach-Object FileVersionInfo | Where-Object ProductVersion -match OpenSSH
+        $valid = $service -and $executableMatches
+        if ($valid) {
+            return $service;
+        }
     }
-
-    return $false;
 }
-function Start-NativeSshAgent([switch]$Quiet) {
+
+function Start-NativeSshAgent([switch]$Quiet, [string]$StartupType = 'Manual') {
     $service = Get-NativeSshAgent
     
     if (!$service) {
@@ -436,13 +437,13 @@ function Start-NativeSshAgent([switch]$Quiet) {
     }
 
     # Enable the servivce if it's disabled and we're an admin
-    if ($service.StartupType -eq "Disabled") {
+    if ($service.StartType -eq "Disabled") {
         if (Test-Administrator) {
-            Set-Service $service -StartupType Automatic
+            # Must pipe rather than calling Set-Service $service or this won't work on PS <= 5.
+            $service | Set-Service -StartupType $StartupType
         }
         else {
-            # Todo: Do we want to display a message, throw an exception or something else if the service is disabled?
-            write-host The ssh-agent service is disabled. Please start the service and try again.
+            Write-Error "The ssh-agent service is disabled. Please start the service and try again."
             # Exit with true so Start-SshAgent doesn't try to do any other work.
             return $true 
         }
@@ -451,7 +452,7 @@ function Start-NativeSshAgent([switch]$Quiet) {
     # Start the service
     if ($service.Status -ne "Running") {
         if (!$Quiet) {
-            write-host "Starting ssh agent service."
+            Write-Host "Starting ssh agent service."
         }
         Start-Service $service
     }
@@ -499,11 +500,10 @@ function Find-Ssh($program = 'ssh-agent') {
 }
 
 # Loosely based on bash script from http://help.github.com/ssh-key-passphrases/
-function Start-SshAgent([switch]$Quiet) {
-
+function Start-SshAgent([switch]$Quiet, [string]$StartupType = 'Manual') {
     # If we're using the win10 native ssh client,
     # we can just interact with the service directly.
-    if (Start-NativeSshAgent -Quiet:$Quiet) {
+    if (Start-NativeSshAgent -Quiet:$Quiet -StartupType:$StartupType) {
         return
     }
 
@@ -613,7 +613,7 @@ function Add-SshKey([switch]$Quiet) {
             # Win10 ssh agent will prompt for key password even if the key has already been added
             # Check to see if any keys have been added. Only add keys if it's empty.
             if (Get-NativeSshAgent) {
-                & $sshAdd -L | Out-Null
+                (& $sshAdd -L) | Out-Null
                 if ($LASTEXITCODE -eq 0) {
                     # Keys have already been added
                     if (!$Quiet) {
@@ -643,7 +643,7 @@ function Stop-SshAgent() {
                 Stop-Service $nativeAgent
             }
             else {
-                write-host Access denied. Please open a prompt as administrator and try again.
+                Write-Error "Access denied. Please open a prompt as administrator and try again."
             }
         }
         return;
--- a/src/GitUtils.ps1
+++ b/src/GitUtils.ps1
@@ -517,11 +517,18 @@ function Find-Ssh($program = 'ssh-agent') {
 }
 
 # Loosely based on bash script from http://help.github.com/ssh-key-passphrases/
-function Start-SshAgent(
-    [switch]$Quiet,
+function Start-SshAgent {
+    param(
+       [Parameter(Position = 0)]
+       [ValidateSet("Automatic", "Boot", "Disabled", "Manual", "System")]
+       [string]
+       $StartupType = "Manual",
 
-    [ValidateSet("Automatic", "Boot", "Disabled", "Manual", "System")]
-    [string]$StartupType = 'Manual') {
+       [Parameter()]
+       [switch]
+       $Quiet
+    )
+
     # If we're using the win10 native ssh client,
     # we can just interact with the service directly.
     if (Start-NativeSshAgent -Quiet:$Quiet -StartupType:$StartupType) {
--- a/src/GitUtils.ps1
+++ b/src/GitUtils.ps1
@@ -439,8 +439,7 @@ function Start-NativeSshAgent([switch]$Quiet, [string]$StartupType = 'Manual') {
     # Enable the servivce if it's disabled and we're an admin
     if ($service.StartType -eq "Disabled") {
         if (Test-Administrator) {
-            # Must pipe rather than calling Set-Service $service or this won't work on PS <= 5.
-            $service | Set-Service -StartupType $StartupType
+            Set-Service "ssh-agent" -StartupType $StartupType
         }
         else {
             Write-Error "The ssh-agent service is disabled. Please start the service and try again."
@@ -454,11 +453,11 @@ function Start-NativeSshAgent([switch]$Quiet, [string]$StartupType = 'Manual') {
         if (!$Quiet) {
             Write-Host "Starting ssh agent service."
         }
-        Start-Service $service
+        Start-Service "ssh-agent"
     }
 
     # Make sure GIT_SSH is set to OpenSSH-Win32
-    setenv 'GIT_SSH' (Get-Command ssh.exe -ErrorAction Ignore | Select-Object -ExpandProperty Path)
+    setenv "GIT_SSH" (Get-Command ssh.exe -ErrorAction Ignore | Select-Object -ExpandProperty Path)
     
     if (!$Quiet) {
         Write-Host "Setting GIT_SSH set to $($env:GIT_SSH)."
--- a/src/GitUtils.ps1
+++ b/src/GitUtils.ps1
@@ -457,16 +457,21 @@ function Start-NativeSshAgent([switch]$Quiet, [string]$StartupType = 'Manual') {
     }
 
     # Make sure git is configured to use OpenSSH-Win32
-    $hasSshExplicitlySet = git config --global core.sshCommand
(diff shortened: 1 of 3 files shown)


src/GitUtils.ps1
```diff
@@ -415,6 +415,52 @@ function Get-SshAgent() {
...
+    }
+
+    # Enable the servivce if it's disabled and we're an admin
+    if ($service.StartupType -eq "Disabled") {
+        if (Test-Administrator) {
+            Set-Service $service -StartupType Automatic
+        }
+        else {
+            # Todo: Do we want to display a message, throw an exception or something else if the service is disabled?
+            write-host The ssh-agent service is disabled. Please start the service and try again.
```
dahlbyk: I think `Write-Error` is most appropriate, since the service failed to start? Even with `-Quiet`, a failure is a failure. I believe this also makes the following `return $true` unnecessary.

And general nitpick: casing should be `Write-Host`, and the message should be quoted.

src/GitUtils.ps1
```diff
@@ -415,6 +415,52 @@ function Get-SshAgent() {
     return 0
 }
 
+function Get-NativeSshAgent {
+    # The ssh.exe binary version must include "OpenSSH"
+    # The windows ssh-agent service must exist
+    $service = Get-Service ssh-agent
```
dahlbyk: `-ErrorAction Ignore`?

src/GitUtils.ps1
```diff
@@ -415,6 +415,52 @@ function Get-SshAgent() {
...
+    $executableMatches = Get-Command ssh.exe | ForEach-Object FileVersionInfo | Where-Object ProductVersion -match OpenSSH
+    $valid = $service -and $executableMatches
+
+    if ($valid) {
+        return $service;
+    }
+
+    return $false;
```
dahlbyk: Would it be more PowerShell-y to `return $null` here? Or to not `return` anything at all?

src/GitUtils.ps1
```diff
@@ -415,6 +415,52 @@ function Get-SshAgent() {
...
+    $service = Get-NativeSshAgent
+    
+    if (!$service) {
+        return $false;
+    }
+
+    # Enable the servivce if it's disabled and we're an admin
+    if ($service.StartupType -eq "Disabled") {
+        if (Test-Administrator) {
+            Set-Service $service -StartupType Automatic
```
dahlbyk: IMO `Manual` seems maybe more appropriate here if they're using a manual `Start-SshAgent` function.

src/GitUtils.ps1
```diff
@@ -569,6 +635,20 @@ function Add-SshKey([switch]$Quiet) {
...
 function Stop-SshAgent() {
+    $nativeAgent = Get-NativeSshAgent
+
+    if ($nativeAgent) {
+        if ($nativeAgent.Status -eq 'Running') {
+            if (Test-Administrator) {
+                Stop-Service $nativeAgent
+            }
+            else {
+                write-host Access denied. Please open a prompt as administrator and try again.
```
dahlbyk: `Write-Error` (with quoted message) here too, I think.

src/GitUtils.ps1
rkeithhill: Return nothing if the service is not installed.  That will successfully compare to $null.

src/GitUtils.ps1
rkeithhill: There are going to be times when you want to start VS after a reboot and before you've fired up a PowerShell console.  In those cases, using Visual Studio against a Git repo using an ssh URL will be no bueno because Windows did not automatically start ssh-agent.

dahlbyk:
It seems like a good idea to set `GIT_SSH` in `Start-NativeSshAgent`, to the path found with `Get-Comman ssh.exe`. #575 mentions `GIT_SSH_COMMAND`, too.

rkeithhill:
`GIT_SSH_COMMAND` is a bit more corner case i.e. only when you need to specify not just the ssh.exe path but some parameters as well.

src/GitUtils.ps1
rkeithhill: Should we firewall this function to be Windows only?  If I ran this on PS Core on Linux/macOS, it would choke on `Get-Service` and `FileVersionInfo`

src/GitUtils.ps1
rkeithhill: Also another command that probably should be Windows only.  BTW, I do this test with:
```
if (($PSVersionTable.PSVersion.Major -lt 6) -or $IsWindows) {
```
That seems to work with strict mode.

rkeithhill: `Manual` startup would indeed be less presumptuous but I just don't see folks using SSH keys w/passphrases on Windows as wanting the startup mode to be anything other than `Automatic`.   At the very least, a switch parameter for automatic startup would be nice.

src/GitUtils.ps1
rkeithhill: Parts of this (Stop-Service) are Windows specific and should be firewalled off on Linux/macOS.

src/GitUtils.ps1
```diff
@@ -557,6 +610,19 @@ function Add-SshKey([switch]$Quiet) {
         }
 
         if ($args.Count -eq 0) {
+            # Win10 ssh agent will prompt for key password even if the key has already been added
```
rkeithhill: How do we get here for Win10 ssh agent?  Won't the `return` on line 507 result in this never being executed?

dahlbyk: This is in `Add-SshKey`.

src/GitUtils.ps1
dahlbyk: As long as `Get-NativeSshAgent` returns falsey on non-Windows, no changes should be necessary here.

Let's go with Manual, plus an option if @JeremySkinner is up for it. Priority is to stop it being completely broken.

src/GitUtils.ps1
dahlbyk: Note this is targeting v0, so xplat is a nice-to-have. Easy enough to fix, regardless, if we intend posh-ssh to be xplat.

src/GitUtils.ps1
rkeithhill: D'oh!  Yeah, that was kind of obscured by GH's helpful collapsing of unchanged lines.  :-)

src/GitUtils.ps1
JeremySkinner: Fixed

JeremySkinner: The `return $true` is still necessary even when using `Write-Error`

src/GitUtils.ps1
JeremySkinner: Added `-ErrorAction Ignore` and added OS checks using PS Core's `$IsMacOs` and `$IsLinux` (undefined on PS 5). Let me know if you'd prefer a different mechanism for doing the OS check

src/GitUtils.ps1
JeremySkinner: Done.

src/GitUtils.ps1
JeremySkinner: - Added `StartupType` parameter to the function, which defaults to manual, which is now passed through from `Start-SshAgent`
- Version check is now handled in `Get-NativeSshAgent` so there shouldn't be a need for another one in `Start-SshAgent`

I also realised the `$service.StartupType` property is only defined in PS Core, so switched to `$service.StartType` instead as this is available in PS 4 and newer. I'm assuming we don't care about older PS versions as they won't run on win10 anyway.

src/GitUtils.ps1
JeremySkinner: - Switched to Write-Error
- `Get-NativeSshAgent` now does an OS check, so an explicit check in `Stop-Service` shouldn't be necessary

JeremySkinner:
I've made all the changes above and also changed `Start-NativeSshAgent` to set `GIT_SSH` (alternatively we could switch to setting the ssh path in `.gitconfig` if that's a better option). 

One issue I have run into is that if you run powershell as admin, then `ssh-add` does not work - it always returns the error `Error connecting to agent: No such file or directory`. Not sure what the best way to handle this.

JeremySkinner:
Made the following changes this afternoon:

- Added tests
- Stopped creation of the SSH_AUTH_SOCK/SSH_AGENT_PID environment variables if we're using native openssh

Re: the `GIT_SSH` environment variable: I'm currently setting it for the current process like we do for variables elsehwere, I'm now thinking it'd be better for them to be set at the user level as then it'll affect other git clients too. However setting the environment variable for the user is very slow, so now I'm leaning towards the config route instead.

rkeithhill:
> However setting the environment variable for the user is very slow, so now I'm leaning towards the config route instead.

👍

JeremySkinner:
I've made the change to set the `core.sshcommand` config setting rather than using `GIT_SSH` and updated the tests.

src/GitUtils.ps1
```diff
@@ -456,11 +456,17 @@ function Start-NativeSshAgent([switch]$Quiet, [string]$StartupType = 'Manual') {
...
 
-    # Make sure GIT_SSH is set to OpenSSH-Win32
-    setenv "GIT_SSH" (Get-Command ssh.exe -ErrorAction Ignore | Select-Object -ExpandProperty Path)
-    
-    if (!$Quiet) {
-        Write-Host "Setting GIT_SSH set to $($env:GIT_SSH)."
+    # Make sure git is configured to use OpenSSH-Win32
+    $hasSshExplicitlySet = git config --global core.sshCommand
+
+    if (!$hasSshExplicitlySet) {
```
dahlbyk: Should we warn folks if their existing `core.sshCommand` looks like it's wrong relative to the `ssh.exe` we found?

JeremySkinner: Good idea, I've modified this to output a warning if it's set to something other than what we expect.

JeremySkinner:
@dahlbyk @rkeithhill I don't think there's anything outstanding on this. Please give it a try and let me know if you run into any problems.

src/GitUtils.ps1
```diff
@@ -453,7 +517,13 @@ function Find-Ssh($program = 'ssh-agent') {
 }
 
 # Loosely based on bash script from http://help.github.com/ssh-key-passphrases/
-function Start-SshAgent([switch]$Quiet) {
+function Start-SshAgent([switch]$Quiet, [string]$StartupType = 'Manual') {
```
rkeithhill: Could we add `[ValidateSet("Automatic", "Boot", "Disabled", "Manual", "System")]` to the `$StartupType` parameter?  This will enable tab-completion on this paramter.

JeremySkinner:
@rkeithhill thanks, I've made that change.

src/GitUtils.ps1
```diff
@@ -517,7 +517,11 @@ function Find-Ssh($program = 'ssh-agent') {
 }
 
 # Loosely based on bash script from http://help.github.com/ssh-key-passphrases/
-function Start-SshAgent([switch]$Quiet, [string]$StartupType = 'Manual') {
+function Start-SshAgent(
+    [switch]$Quiet,
+
+    [ValidateSet("Automatic", "Boot", "Disabled", "Manual", "System")]
+    [string]$StartupType = 'Manual') {
```
rkeithhill: When using parameter attributes I would recommend using a param block e.g.:
```
 function Start-SshAgent {
    param(
        [Parameter(Position = 0)]
        [ValidateSet("Automatic", "Boot", "Disabled", "Manual", "System")]
        [string]
        StartupType = "Manual",

        [Parameter()]
        [switch]
        $Quiet
    )
    ...
```
Not required but it is more "canonical" PowerShell.  :-)

JeremySkinner: @rkeithhill Ah! Thanks, I didn't know that. That's the c# developer in me ;) I'll change that now.

JeremySkinner: Done!

src/GitUtils.ps1
```diff
@@ -517,11 +517,18 @@ function Find-Ssh($program = 'ssh-agent') {
...
+    param(
+       [Parameter(Position = 0)]
+       [ValidateSet("Automatic", "Boot", "Disabled", "Manual", "System")]
+       [string]
+       $StartupType = "Manual",
+
+       [Parameter()]
+       [switch]
+       $Quiet
+    )
```
rkeithhill: 👍

rkeithhill:
I think this is ready to merge.

dahlbyk:
Great work, thanks @JeremySkinner (and @rkeithhill for review)!

JeremySkinner:
Great, thanks both of you for the feedback. 

@dahlbyk how do you want to handle things next? I had a quick look through the outstanding ssh issues, and at this point I'd rather wait until we've done the separation to address those, unless there are any you're keen to get into a 0.x release sooner?

dahlbyk:
> I had a quick look through the outstanding ssh issues, and at this point I'd rather wait until we've done the separation to address those, unless there are any you're keen to get into a 0.x release sooner?

There are three open PRs with SSH changes: https://github.com/dahlbyk/posh-git/pulls?q=is%3Aopen+is%3Apr+milestone%3A%22SSH+Agents%22. Are any of those worth salvaging? I believe this PR makes #295 redundant.

Other than that, I think we're good. I've pushed `v0` latest to https://github.com/dahlbyk/posh-ssh as a starting point, and invited you as collaborators.

As for #585, can we just drop the constructive commits and keep the destructive ones that remove SSH stuff? I don't think we need a new PR.

JeremySkinner:
Yes #295 is now obsolete and can be closed. 

I want to implement #195 but will re-do it based on the ssh config parser I’m working on. 

I don’t understand the point of #316 and as the author never commented or added a description I’d suggest you close it. 

Yes I’ll keep the destructive commits in #585 and remove the rest - I’ll do that tomorrow.

PR #595, opened 2018-07-09
Add Get-PromptConnectionInfo as DefaultPromptPrefix

### LINKED ISSUE #591: Enhance prompt function to show username/hostname in SSH connection
As I've been using Windows 10 OpenSSH's sshd service and connecting to my Windows 10 system using ssh, I've been missing seeing the `[username@hostname]` info the PowerShell prompt on the remote system that I'm used to seeing when I ssh into a Linux box.

We could rectify that by adding the following to our prompt function (somewhere before the prompt suffix):
```powershell
"$(if (Test-Path Env:\SSH_CONNECTION) {" [$([System.Environment]::UserName)@$([System.Environment]::MachineName)]"})"
```
A few notes, you get the remote system's PowerShell prompt when you A) ssh into the remote system with OpenSSH sshd running and configured with this subsystem:
```
Subsystem	powershell	C:/Program Files/PowerShell/6-preview/pwsh.exe -sshs -NoLogo -NoProfile
```
And no, I'm not quite sure why we get the posh-git prompt when the above specifies `-NoProfile`.  Anyway, I've modified my posh-git prompt to this:
```
$global:GitPromptSettings.DefaultPromptBeforeSuffix.Text = '`n$([DateTime]::Now.ToString("MM-dd HH:mm:ss"))$(if (Test-Path Env:\SSH_CONNECTION) {" [$([System.Environment]::UserName)@$([System.Environment]::MachineName)]"})'
```
So that when I ssh into this system, I see this prompt:
```
C:\Users\hillr
06-27 09:51:19 [hillr@HILLR2] 1>
```
Note that when you use PS Remoting over SSH, PowerShell controls the prompt.  In that case, you currently get just the hostname in the prompt:
```
06-27 09:51:57 8> Enter-PSSession -HostName hillr2
hillr@KEYSIGHT@hillr2's password:
[hillr2]: PS C:\Users\hillr\Documents>
```
But that is changing - see https://github.com/PowerShell/PowerShell/issues/7156 &  https://github.com/PowerShell/PowerShell/pull/7191

So, should we consider covering this scenario with our prompt function or just add an example of how to customize the posh-git prompt if you want this feature?

dahlbyk:
Procrastination!

Closes #591 by roughly imitating https://github.com/PowerShell/PowerShell/pull/7191 (show `[user@host]` if user seems to not match, otherwise `[host]`). I don't use PS remoting, so I'll need some help verifying this works as expected.

Rather than place the logic inline, I figure exporting `Get-PromptConnectionInfo` affords flexibility to make this configurable (e.g. `$GitPromptSettings.DefaultPromptConnectionShowUserName = 'always|never|different'` or `$GitPromptSettings.DefaultPromptConnectionFormatWithUserName`).

### DIFF
--- a/src/PoshGitTypes.ps1
+++ b/src/PoshGitTypes.ps1
@@ -280,6 +280,7 @@ class PoshGitPromptSettings {
     [PoshGitTextSpan]$DefaultPromptSuffix       = '$(">" * ($nestedPromptLevel + 1)) '
 
     [bool]$DefaultPromptAbbreviateHomeDirectory = $true
+    [string]$DefaultPromptConnectionFormat      = '[{1}@{0}]: '
     [bool]$DefaultPromptWriteStatusFirst        = $false
     [bool]$DefaultPromptEnableTiming            = $false
     [PoshGitTextSpan]$DefaultPromptTimingFormat = ' {0}ms'
--- a/src/PoshGitTypes.ps1
+++ b/src/PoshGitTypes.ps1
@@ -273,7 +273,7 @@ class PoshGitPromptSettings {
     [string]$DescribeStyle = ''
     [psobject]$WindowTitle = {param($GitStatus, [bool]$IsAdmin) "$(if ($IsAdmin) {'Admin: '})$(if ($GitStatus) {"$($GitStatus.RepoName) [$($GitStatus.Branch)]"} else {Get-PromptPath}) ~ PowerShell $($PSVersionTable.PSVersion) $([IntPtr]::Size * 8)-bit ($PID)"}
 
-    [PoshGitTextSpan]$DefaultPromptPrefix       = ''
+    [PoshGitTextSpan]$DefaultPromptPrefix       = '$(Get-PromptConnectionInfo)'
     [PoshGitTextSpan]$DefaultPromptPath         = '$(Get-PromptPath)'
     [PoshGitTextSpan]$DefaultPromptBeforeSuffix = ''
     [PoshGitTextSpan]$DefaultPromptDebug        = [PoshGitTextSpan]::new(' [DBG]:', [ConsoleColor]::Magenta)
--- a/src/PoshGitTypes.ps1
+++ b/src/PoshGitTypes.ps1
@@ -273,14 +273,13 @@ class PoshGitPromptSettings {
     [string]$DescribeStyle = ''
     [psobject]$WindowTitle = {param($GitStatus, [bool]$IsAdmin) "$(if ($IsAdmin) {'Admin: '})$(if ($GitStatus) {"$($GitStatus.RepoName) [$($GitStatus.Branch)]"} else {Get-PromptPath}) ~ PowerShell $($PSVersionTable.PSVersion) $([IntPtr]::Size * 8)-bit ($PID)"}
 
-    [PoshGitTextSpan]$DefaultPromptPrefix       = '$(Get-PromptConnectionInfo)'
+    [PoshGitTextSpan]$DefaultPromptPrefix       = '$(Get-PromptConnectionInfo -Format "[{1}@{0}]: ")'
     [PoshGitTextSpan]$DefaultPromptPath         = '$(Get-PromptPath)'
     [PoshGitTextSpan]$DefaultPromptBeforeSuffix = ''
     [PoshGitTextSpan]$DefaultPromptDebug        = [PoshGitTextSpan]::new(' [DBG]:', [ConsoleColor]::Magenta)
     [PoshGitTextSpan]$DefaultPromptSuffix       = '$(">" * ($nestedPromptLevel + 1)) '
 
     [bool]$DefaultPromptAbbreviateHomeDirectory = $true
-    [string]$DefaultPromptConnectionFormat      = '[{1}@{0}]: '
     [bool]$DefaultPromptWriteStatusFirst        = $false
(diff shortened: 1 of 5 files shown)


src/Utils.ps1
```diff
@@ -301,6 +301,20 @@ function Get-PromptPath {
     return $currentPath
 }
 
+$sshType = 'System.Management.Automation.Runspaces.SSHConnectionInfo' -as [Type]
+$sshUserName = Get-Runspace | ForEach-Object ConnectionInfo |
+  Where-Object { $sshType -and $_ -is $sshType } | ForEach-Object UserName
+
+function Get-PromptConnectionInfo {
+  if (Test-Path Env:SSH_CONNECTION) {
+    if ($sshUserName -and $sshUserName -ne [System.Environment]::UserName) {
```
rkeithhill: They made a late change in that PowerShell PR due to user names on Linux being case-sensitive.  I think you want `-cne` here.  Also, I thought we were using soft-tabs of *4* spaces?  :-)  I'll test this tomorrow when I get into work and have access to a system with `sshd` set to use PowerShell.

rkeithhill:
Looking good for the same user:
image

But when I ssh into my Windows box as a different user, I do not get the username.  It appears that the Runspace's `ConnectionInfo` object is `$null`.  That leaves no good way to determine the orginal user.  :-(

dahlbyk:
> But when I ssh into my Windows box as a different user, I do not get the username. It appears that the Runspace's `ConnectionInfo` object is `$null`. That leaves no good way to determine the orginal user.

What's different here compared with https://github.com/PowerShell/PowerShell/pull/7191?

rkeithhill:
I think it is because that method in the PR is getting a `RemoteRunspace`.  When I run in my ssh session `Get-Runspace | Get-Member`  I get `System.Management.Automation.Runspaces.LocalRunspace`.  Maybe when we detect an SSH session, we just always display the username?  That would be pretty consistent with ssh'ing into sh/bash.

dahlbyk:
Taking a closer look at [`EvaluatePrompt`](https://github.com/PowerShell/PowerShell/blob/e177fcaef610f874aa31cd3d85f97162d2859e72/src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHost.cs#L2748-L2777), I'm fuzzy on why we need to do anything. It seems like `prompt` gets called and then [`GetRemotePrompt`](https://github.com/PowerShell/PowerShell/blob/ef104601727745f357bc95f32d4a43c6df2a54f1/src/System.Management.Automation/engine/hostifaces/HostUtilities.cs#L766-L779) wraps the result.

@rkeithhill does `$Host.Runspace` get you a `RemoteRunspace`?

rkeithhill:
That code gets used from a PS remoting session. It does not get used when you SSH into a Windows system where the default shell is set to PowerShell.

rkeithhill:
$Host.Runspace is a LocalRunspace as well.

dahlbyk:
> That code gets used from a PS remoting session. It does not get used when you SSH into a Windows system where the default shell is set to PowerShell.

I knew I was missing something. That's the ELI5 I needed.

--------------------------------

Added code to always include username (`[user@host]: `), plus a setting to change the connection format string.

src/PoshGitTypes.ps1
```diff
@@ -273,13 +273,14 @@ class PoshGitPromptSettings {
...
 
-    [PoshGitTextSpan]$DefaultPromptPrefix       = ''
+    [PoshGitTextSpan]$DefaultPromptPrefix       = '$(Get-PromptConnectionInfo)'
     [PoshGitTextSpan]$DefaultPromptPath         = '$(Get-PromptPath)'
     [PoshGitTextSpan]$DefaultPromptBeforeSuffix = ''
     [PoshGitTextSpan]$DefaultPromptDebug        = [PoshGitTextSpan]::new(' [DBG]:', [ConsoleColor]::Magenta)
     [PoshGitTextSpan]$DefaultPromptSuffix       = '$(">" * ($nestedPromptLevel + 1)) '
 
     [bool]$DefaultPromptAbbreviateHomeDirectory = $true
+    [string]$DefaultPromptConnectionFormat      = '[{1}@{0}]: '
```
rkeithhill: I wonder if instead of this setting that isn't perhaps obviously connected to `Get-PromptConnectionInfo`, we should just have that command take a format string as parameter.  We could make that parameter optional (default it to `'[{1}@{0}]: ' ` but I'd use it explicitly for `$DefaultPromptPrefix` so folks see they can tweak the format e.g.:
```
[PoshGitTextSpan]$DefaultPromptPrefix       = '$(Get-PromptConnectionInfo -Format "[{1}@{0}]: ")' 
```

Although at this point maybe we switch the format placeholders to `[{01}@{1}]: `?

rkeithhill: I think `-Format` is better and we should definitely provide a default.  The only reason I would specify the `-Format …` parameter in that setting is purely pedagogical.


Re `DefaultPromptAbbreviateHomeDirectory`, sounds good to me.  We should note that in the CHANGELOG as a potentially breaking change though.

dahlbyk: Eh, `DefaultPromptAbbreviateHomeDirectory` has been established since v0.7.0, doesn't seem worth breaking everyone now.

I'll replace the new setting with `-Format`.

PR #597, opened 2018-07-11
Do not update WindowTitle at all in prompt func if setting is $null

### LINKED ISSUE #594: WindowTitle set by user overwritten by module
### System Details

- posh-git version/path: 1.0.0 beta2 ~\Documents\WindowsPowerShell\Modules\posh-git\1.0.0
- PowerShell version: 5.1.17134.112
- git version 2.18.0.windows.1
- OS: Microsoft Windows NT 10.0.17134.0

### Issue Description

I have the following in my $profile:

    Import-Module posh-git
    $GitPromptSettings.WindowTitle = $null

If I start a script that changes `$host.UI.RawUI.WindowTitle` directly then this still gets overwritten with `$PreviousWindowTitle` that was initialised with the text from when the PowerShell window was started. I would like my script to be able to change the title and have the same effect regardless of whether posh-git is loaded.

I know there has been some discussion about this before in previous issues - but could it be considered that a setting could be added to `$GitPromptSettings` to disable any update of `$host.UI.RawUI.WindowTitle`?

Maybe this could also be added as a parameter when the module starts so that the WindowTitle never gets updated or changed by the module (and hence would not need to be reverted to $PreviousWindowTitle).

rkeithhill:
We used to restore the previousWindowTitle but this had the effect of
always updating the WindowTitle with a fixed string - whatever we
captured in $PreviousWindowTitle.

However, we still restore the original window title when the module is
removed. But we only do this if the module is configured to supply the
window title.

Rename variable to make its intent clearer.

Fixes #594

### DIFF
--- a/src/WindowTitle.ps1
+++ b/src/WindowTitle.ps1
@@ -10,11 +10,13 @@ function Test-WindowTitleIsWriteable {
             $Host.UI.RawUI.WindowTitle = $newTitle
             $script:HostSupportsSettingWindowTitle = ($Host.UI.RawUI.WindowTitle -eq $newTitle)
             $Host.UI.RawUI.WindowTitle = $OriginalWindowTitle
+            Write-Debug "HostSupportsSettingWindowTitle: $HostSupportsSettingWindowTitle"
+            Write-Debug "OriginalWindowTitle: $OriginalWindowTitle"
         }
         catch {
             $script:OriginalWindowTitle = $null
             $script:HostSupportsSettingWindowTitle = $false
-            Write-Debug "Probing for HostSupportsSettingWindowTitle errored: $_"
+            Write-Debug "HostSupportsSettingWindowTitle error: $_"
         }
     }
     return $HostSupportsSettingWindowTitle
@@ -27,6 +29,7 @@ function Reset-WindowTitle {
 
     # Revert to original WindowTitle, but only if posh-git is currently configured to set it
     if ($HostSupportsSettingWindowTitle -and $OriginalWindowTitle -and $settings.WindowTitle) {
+        Write-Debug "Resetting WindowTitle: '$OriginalWindowTitle'"
         $Host.UI.RawUI.WindowTitle = $OriginalWindowTitle
     }
 }
@@ -40,13 +43,14 @@ function Set-WindowTitle {
     if ($settings.WindowTitle -and (Test-WindowTitleIsWriteable)) {
         try {
             if ($settings.WindowTitle -is [scriptblock]) {
-                $windowTitleText = & $settings.WindowTitle $GitStatus $IsAdmin
+                # ensure results returned by scriptblock are flattened into a string
+                $windowTitleText = "$(& $settings.WindowTitle $GitStatus $IsAdmin)"
             }
             else {
                 $windowTitleText = $ExecutionContext.SessionState.InvokeCommand.ExpandString("$($settings.WindowTitle)")
             }
 
-            # Put $windowTitleText in a string to ensure results returned by scriptblock are flattened into a string
+            Write-Debug "Setting WindowTitle: $windowTitleText"
             $Host.UI.RawUI.WindowTitle = "$windowTitleText"
         }
         catch {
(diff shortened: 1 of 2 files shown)


src/posh-git.psm1
```diff
@@ -18,16 +18,17 @@ if (!$Env:HOME) { $Env:HOME = "$Env:USERPROFILE" }
 $IsAdmin = Test-Administrator
 
 # Probe $Host.UI.RawUI.WindowTitle to see if it can be set without errors
-$WindowTitleSupported = $false
+$HostSupportsSettingWindowTitle = $false
 try {
-    $global:PreviousWindowTitle = $Host.UI.RawUI.WindowTitle
-    $newTitle = "${global:PreviousWindowTitle} "
+    $global:PoshGitOrigWindowTitle = $Host.UI.RawUI.WindowTitle
```
dahlbyk: Does this actually need to be global? Could it be script-scoped alongside `$HostSupportsSettingWindowTitle`, with a corresponding `Reset-WindowTitle` to use in `OnRemove`?

src/posh-git.psm1
```diff
@@ -92,34 +89,22 @@ $GitPromptScriptBlock = {
...
 
-    # Update the host's WindowTitle is host supports it and user has not disabled $GitPromptSettings.WindowTitle
+    # Update the host's WindowTitle if host supports it and user has not disabled $GitPromptSettings.WindowTitle
     # This has to be *after* the call to Write-VcsStatus, which populates $global:GitStatus
-    if ($WindowTitleSupported) {
-        $windowTitle = $settings.WindowTitle
-        if (!$windowTitle) {
-            if ($global:PreviousWindowTitle) {
-                $Host.UI.RawUI.WindowTitle = $global:PreviousWindowTitle
+    if ($HostSupportsSettingWindowTitle -and $settings.WindowTitle) {
```
dahlbyk: While we're at it, should this whole block move into a `Set-WindowTitle` function?

src/posh-git.psm1
```diff
@@ -18,16 +18,17 @@ if (!$Env:HOME) { $Env:HOME = "$Env:USERPROFILE" }
 $IsAdmin = Test-Administrator
 
 # Probe $Host.UI.RawUI.WindowTitle to see if it can be set without errors
-$WindowTitleSupported = $false
+$HostSupportsSettingWindowTitle = $false
```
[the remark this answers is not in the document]
rkeithhill: But when `$settings.WindowTitle` isn't falsey, we would be doing this "writable" check during every prompt eval - effectively setting the title three times.  

I think it is better to do this check once during module load but I'm fine with creating a function in Utils.ps1 so we'd have:

```posh
$IsAdmin = Test-Administrator
$IsWindowTitleWritable = Test-WindowTitleIsWritable
```

src/posh-git.psm1
rkeithhill: Yeah, it could be script scoped.

src/posh-git.psm1
rkeithhill: Sure, but we'd need to export it from the module to make it callable from the prompt func.  How about `Reset-WindowTitle` - does it need to be exported or would we be the only caller in the OnRemove event handler.

dahlbyk:
> > While we're at it, should this whole block move into a `Set-WindowTitle` function?
>
> Sure, but we'd need to export it from the module to make it callable from the prompt func.

I was pretty sure this wasn't the case, and my experiment to find out may have gone a bit far. Anyway, the new commits here seem to work as expected and the module is much cleaner.

image

src/WindowTitle.ps1
```diff
@@ -0,0 +1,60 @@
+$HostSupportsSettingWindowTitle = $null
+$OriginalWindowTitle = $null
+
+function Test-WindowTitleIsWriteable {
+    if ($null -eq $HostSupportsSettingWindowTitle) {
```
rkeithhill: 👍

rkeithhill:
Looks good!  My daughter just got back from her choir trip to Europe so didn't have time to tackle this last night.  Thanks!!  Go ahead and merge whenever you're ready.

I have to say I'm a bit surprised the scriptblock we assign to the global func prompt can see that non-exported command.  I was assuming the scriptblock closed over the variables it used.  TIL that scriptblock seems to have access to module scope.  :-)

PR #606, opened 2018-08-18
Remove ".exe" suffix from PowerTab integration

bradrn:
This small change was necessary to get it working on my system.

### DIFF
--- a/src/GitTabExpansion.ps1
+++ b/src/GitTabExpansion.ps1
@@ -442,7 +442,7 @@ function GitTabExpansionInternal($lastBlock, $GitStatus = $null) {
 
 $PowerTab_RegisterTabExpansion = if (Get-Module -Name powertab) { Get-Command Register-TabExpansion -Module powertab -ErrorAction SilentlyContinue }
 if ($PowerTab_RegisterTabExpansion) {
-    & $PowerTab_RegisterTabExpansion "git.exe" -Type Command {
+    & $PowerTab_RegisterTabExpansion "git" -Type Command {
         param($Context, [ref]$TabExpansionHasOutput, [ref]$QuoteSpaces)  # 1:
 
         $line = $Context.Line


rkeithhill:
Where you getting an error or just no tab-completion?  I'm curious why removing the extension makes this work.  Are you on a non-Windows OS?

dahlbyk:
@bradrn what does `Get-Command git` show for you? I'm wondering if this is a difference between an Application in `PATH` vs an Alias.

Note the original implementation (b35085d4a32c86d85b89b2e88420f4cc7a72fdac via #63) used `git.cmd`, but was switched to `git.exe` in #66.

cc @jasonmarcher

rkeithhill:
I'm not sure if PowerTab works on PSCore on Linux/macOS but if it does, `git` there will have no extension at all.  Maybe that's what @bradrn is running into.  That should work on Windows as well.

bradrn:
@rkeithhill:
- (Re https://github.com/dahlbyk/posh-git/pull/606#issuecomment-414074160) I'm not getting an error, there's just no tab-completion. When I don't load PowerTab everything works fine: it's just when I load PowerTab that problems occur.
- (Re https://github.com/dahlbyk/posh-git/pull/606#issuecomment-414079103) I'm using Windows.

@dahlbyk: `Get-Command git` shows the following:
```
CommandType     Name                                               Version    Source
-----------     ----                                               -------    ------
Application     git.exe                                            2.10.1.1   C:\Program Files\Git\cmd\git.exe
```

dahlbyk:
Hrm... I was going to try this myself, but the PowerTab install failed. Will try to try again soon, if better guidance doesn't come along first.

bradrn:
Thank you @dahlbyk. I do know that I ran into [this problem](https://github.com/jasonmarcher/PowerTab/issues/101#issuecomment-413797109) when I tried to install PowerTab - I eventually fixed it by doing [this](https://github.com/jasonmarcher/PowerTab/issues/101#issuecomment-414025820).

rkeithhill:
I took a peek at the PowerTab `Register-TabExpansion` impl.  It looks like if `git.exe` was already registered, the new handler is not added.  PowerTab uses the already registered handler.  Also, all of the examples I see specify the extension: ssh.exe, netsh.exe, robocopy.exe.  I do see that `Register-TabExpansion` has a `-Force`  parameter.  What happens if you put it back to `git.exe` but add the `-Force` parameter?

bradrn:
@rkeithhill It still doesn't work if I add `-Force`.

(I'm not actually using PowerTab anymore, so this issue doesn't really affect me anymore. But it would still be nice to get this working...)

bradrn:
Thank you!

PR #607, opened 2018-08-18
Set `$Env:COLUMNS` on prompt if necessary

dahlbyk:
Git uses `COLUMNS`, which is not set by PowerShell, for formatting output (e.g. `git log --stat`). To account for resized windows, we'll set `COLUMNS` to `WindowSize.Width` on every prompt.

Setting `COLUMNS` is currently disabled if it's already set when posh-git loads, or by setting $`GitPromptSettings.SetEnvColumns = $false`.


dahlbyk:
> Setting COLUMNS is currently disabled if it's already set when posh-git loads

I'm tempted to just default this to `$true`. As-is, resizing a nested PowerShell console won't pick up changes in `WindowSize`.

rkeithhill:
Seems reasonable to me.

PR #614, opened 2018-08-31
Merge Posh-Hg into Posh-Git

ExE-Boss:
[**Posh‑Hg**](https://github.com/JeremySkinner/posh-hg) is a PowerShell module, which adds Mercurial integration using **Posh‑Git’s** interface.

Unfortunately, has abandoned @JeremySkinner **Posh‑Hg**, and it’s got compatibility issues with the **Posh‑Git** 1.0.0 beta, so I’m opening this PR to get this into **Posh‑Git** so that this repository can be made into a **Posh‑Vcs** monorepo.

### DIFF
--- a/posh-hg/HgPrompt.ps1
+++ b/posh-hg/HgPrompt.ps1
@@ -1,0 +1,122 @@
+# For backwards compatibility
+$global:HgPromptSettings = $global:PoshHgSettings
+
+function Write-Prompt($Object, $ForegroundColor, $BackgroundColor = -1) {
+    if ($BackgroundColor -lt 0) {
+        Write-Host $Object -NoNewLine -ForegroundColor $ForegroundColor
+    } else {
+        Write-Host $Object -NoNewLine -ForegroundColor $ForegroundColor -BackgroundColor $BackgroundColor
+    }
+}
+
+function Write-HgStatus($status = (get-hgStatus $global:PoshHgSettings.GetFileStatus $global:PoshHgSettings.GetBookmarkStatus)) {
+    if ($status) {
+        $s = $global:PoshHgSettings
+       
+        $branchFg = $s.BranchForegroundColor
+        $branchBg = $s.BranchBackgroundColor
+        
+        if($status.Behind) {
+          $branchFg = $s.Branch2ForegroundColor
+          $branchBg = $s.Branch2BackgroundColor
+        }
+
+        if ($status.MultipleHeads) {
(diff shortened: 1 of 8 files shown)


JeremySkinner:
(Disclaimer; I am the author of posh-hg as well as a contributor to posh-git). 

Personally I'd vote against this. This would significantly increase the complexity of the project and takes the focus away from the core intent of the project (git integration for powershell). If there is interest in posh-hg being resurrected, then it should be forked a new set of maintainers can continue its development (including compatibility with the latest posh-git), but I don't think combining them into a monorepo is a good idea. @dahlbyk what do you think?

dahlbyk:
Agree that a monorepo isn't the right approach, mostly because I don't want to maintain Mercurial code as someone who doesn't use Mercurial. (See also forking off https://github.com/dahlbyk/posh-sshell.)

That being said, my intent all along was to make [posh-vcs](https://github.com/dahlbyk/posh-vcs) a thing that posh-git and posh-hg could both depend on. There's certainly infrastructure here that's not specific to Git.

@ExE-Boss did you find making posh-hg work with posh-git _without_ a monorepo didn't work?

ExE-Boss:
> did you find making posh-hg work with posh-git without a monorepo didn't work?

Not entirely, but a monorepo helps make sure that the changes from one are propagated to the other (eg. the `Write-Prompt` changes), and there were a few slight issues which a monorepo helped solve.

dahlbyk:
> > did you find making posh-hg work with posh-git without a monorepo didn't work?
> 
> Not entirely, but a monorepo helps make sure that the changes from one are propagated to the other (eg. the Write-Prompt changes), and there were a few slight issues which a monorepo helped solve.

That's true. I really don't know much about the PowerShell module dependency story. @rkeithhill or others: is it a terrible idea to make posh-git and posh-hg depend on a common posh-vcs module? Is there anything resembling respect for SemVer so we could rev the core without breaking dependents?

rkeithhill:
PowerShellGet and PowerShell manifests handle module dependencies - see https://docs.microsoft.com/en-us/powershell/wmf/5.0/psget_moduledependency.  But I don't have any first-hand experience creating modules that depend on other modules.  So, I'm not sure about what (if any) issues we might run into if we do this.

dahlbyk:
I don't think it makes sense to disrupt v1.0 by introducing a dependency, so I've assigned this to a v2.0 milestone.

PR #615, opened 2018-08-31
Added note about -AllowPrerelease not being available

ajw1970:
Greg, this is my first attempt at a pull request. Please forgive me if I'm doing it wrong or in a way that isn't standard. 

-AllowPrerelease isn't available even with the updated version of PowerShellGet. I got things working by using Chocolately instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,6 +108,8 @@ posh-git is available on the [PowerShell Gallery][psgallery-v1] and can be insta
     # NOTE: If asked to trust packages from the PowerShell Gallery, answer yes to continue installation of posh-git
     # NOTE: If the AllowPrerelease parameter is not recognized, update your version of PowerShellGet to >= 1.6 e.g.
     #       Install-Module PowerShellGet -Scope CurrentUser -Force -AllowClobber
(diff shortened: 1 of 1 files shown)


rkeithhill:
PS> -AllowPrerelease isn't available even with the updated version of PowerShellGet.

This is not correct. Are you sure you are on the latest version of PowerShellGet?

```
PS> Get-Command Install-Module -Syntax

Install-Module [-Name] <string[]> [-MinimumVersion <string>] [-MaximumVersion <string>] [-RequiredVersion <string>] [-Repository <string[]>] [-Credential <pscredential>] [-Scope <string>] [-Proxy <uri>] [-ProxyCredential <pscredential>] [-AllowClobber] [-SkipPublisherCheck] [-Force] 
>>>>[-AllowPrerelease]<<<<< [-AcceptLicense] [-WhatIf] [-Confirm] [<CommonParameters>]
...

> Get-Module PowerShellGet

ModuleType Version    Name                                ExportedCommands
---------- -------    ----                                ----------------
Script     1.6.0      PowerShellGet                       {Find-Command, Find-DscResource, Find-Module, Find-RoleCap.
```

Issue #612, opened 2018-08-28
Write-VcsStatus behaving differently


codedog:
- posh-git version/path: 1.0.0.0 beta2x C:\source\posh-git\src
- PowerShell version: 5.1.17134.228
- git version 2.18.0.windows.1
- OS: Microsoft Windows NT 10.0.17134.0

### Issue Description

I have a custom multi-lines prompt that used to work with v0.6.1.20160330 of posh-git. The git status stays on the first line. The second line contains just the right angle bracket. Now I have installed it on a new machine, and git status is appearing in the second line.

Here's the content of my profile script:

```
if(!(Test-Path function:\TabExpansion)) { New-Item function:\Global:TabExpansion -value '' | Out-Null }

function Test-Administrator {
    $user = [Security.Principal.WindowsIdentity]::GetCurrent();
    (New-Object Security.Principal.WindowsPrincipal $user).IsInRole([Security.Principal.WindowsBuiltinRole]::Administrator)
}

function prompt {
    $realLASTEXITCODE = $LASTEXITCODE

    Write-Host

    if (Test-Administrator) {  # Use different username if elevated
        Write-Host "^" -NoNewline -ForegroundColor White
    }

    Write-Host "$ENV:USERNAME@" -NoNewline -ForegroundColor DarkYellow
    Write-Host "$ENV:COMPUTERNAME" -NoNewline -ForegroundColor Magenta

    if ($s -ne $null) {  # color for PSSessions
        Write-Host " (`$s: " -NoNewline -ForegroundColor DarkGray
        Write-Host "$($s.Name)" -NoNewline -ForegroundColor Yellow
        Write-Host ") " -NoNewline -ForegroundColor DarkGray
    }

    Write-Host ":" -NoNewline -ForegroundColor DarkGray
    Write-Host $($(Get-Location) -replace ($env:USERPROFILE).Replace('\','\\'), "~") -NoNewline -ForegroundColor Blue
    Write-Host ":" -NoNewline -ForegroundColor DarkGray
    Write-Host (Get-Date -Format "d-MMM HH:mm:ss") -NoNewline -ForegroundColor DarkMagenta
    Write-Host ":" -NoNewline -ForegroundColor DarkGray

    $global:LASTEXITCODE = $realLASTEXITCODE

    Write-VcsStatus

    Write-Host ""

    return "> "
}

function which($command) {
  get-command $command | Format-Table Path
}

Set-Alias which gcm

[...]

PR #616, opened 2018-09-04
Enhance PoshGitTextSpan support for custom VT seqs

rkeithhill:
@dahlbyk See what you think about this.  In the process of explaining for #612 how to use the `CustomAnsi` property on `PoshGitTextSpan` to customize built-in prompt, I realized my previous impl of this idea was too limiting.  For instance before the Path is display the user wanted this displayed in the prompt:

image

The ideal property to use for this info is the `$GitPromptSettings.DefaultPromptPrefix` but there are four different bits of info displayed with different colors.  That currently does NOT work.

My original idea with `CustomAnsi` was that folks could specify a custom sequence that would apply to the text.  That would allow the user to set other attributes like say underline or inverse as well as fg/bg colors.  But that sequence would apply to all of the text so you could have only a single span of text with those attributes.

Now I'm thinking perhaps we should just ditch the `CustomAnsi` property and allow folks to put as much custom seqs in the `Text` field as their heart desires.  For instance, with this change, I can achieve the above desired display with this one setting:
```
function Test-Administrator { $true }

$GitPromptSettings.DefaultPromptPrefix.Text =
    "`n`$(Get-PromptConnectionInfo -Format `"[{1}@{0}]: `")" +
    "`$(if (global:Test-Administrator) {`"$([char]27)[37m^`"})" +
    "$([char]27)[33m`${env:USERNAME}@$([char]27)[95m`${env:COMPUTERNAME}$([char]27)[90m:"
```

Thoughts?

### DIFF
--- a/src/AnsiUtils.ps1
+++ b/src/AnsiUtils.ps1
@@ -38,10 +38,10 @@ function EscapeAnsiString([string]$AnsiString) {
 function Test-VirtualTerminalSequece([psobject[]]$Object, [switch]$Force) {
     foreach ($obj in $Object) {
         if (($Force -or $global:GitPromptSettings.AnsiConsole) -and ($obj -is [string])) {
-            return $obj.Contains($AnsiEscape)
+            $obj.Contains($AnsiEscape)
         }
         else {
-            return $false
+            $false
         }
     }
 }
@@ -68,7 +68,7 @@ function Get-VirtualTerminalSequence ($color, [int]$offset = 0) {
             Write-Debug $_
         }
 
-        # Hard to get here but DarkYellow is not an HTML color but is a ConsoleColor
+        # Hard to get here but DarkYellow is not an HTML color but it is a ConsoleColor
         if (($color -isnot $ColorType) -and ($null -ne ($consoleColor = $color -as [System.ConsoleColor]))) {
             $color = $consoleColor
         }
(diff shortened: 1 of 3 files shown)


src/AnsiUtils.ps1
```diff
@@ -35,12 +35,14 @@ function EscapeAnsiString([string]$AnsiString) {
...
 }
 
-function Test-VirtualTerminalSequece([psobject]$Object) {
-    if ($global:GitPromptSettings.AnsiConsole -and ($Object -is [string])) {
-        return $Object.Contains($AnsiEscape)
-    }
-    else {
-        return $false
+function Test-VirtualTerminalSequece([psobject[]]$Object, [switch]$Force) {
+    foreach ($obj in $Object) {
```
dahlbyk: A little confused by this. Are you meaning to `return` based on the first object received?

rkeithhill: I thought I'd changed that. No, those should simply output a Boolean value.  I'll fix that.

rkeithhill:
I'll fix the tests tonight.  First, I wanted to see if you agreed in principle to this change.

rkeithhill:
Oh yeah baby, fixed the AppVeyor tests.  I submit a separate PR for that.

rkeithhill:
This is ready for a final review/merge.

rkeithhill:
One thing I wonder about this currently is that if the user embeds VT seqs in the .Text property without a terminating seq, we write that string "as-is" when AnsiConsole is set to $false.  That can mess up the screen colors.  We could A) strip out seqs, B) add a terminating seq if we detect seqs or C) leave it as it is.

PR #617, opened 2018-09-05
Fix Write-VcsStatus to not output anything when AnsiConsole -eq $false

rkeithhill:
Addresses potential v0.x to v1.x migration issue surfaced by #612


rkeithhill:
@dahlbyk So do you think we could merge this one?

PR #620, opened 2018-09-07
autocomplete `git log --stat`, `git log --patch` and `git diff --staged`

chillum:


rkeithhill:
RE `--staged` - this is an alias for `--cached` which is already supported.  Do we want to go down the path of supporting parameter aliases (or synonyms as the git docs call them)?  Maybe it would be better to add functionality via $GitTabSettings to allow folks to add parameters / parameter aliases to git commands?

The git-log parameter additions look good.

chillum:
Well, if:
1. Git supports this option
2. Your users request it’s support as they use it

— what’s the point of not supporting it? Saving 9 bytes? ORLY?

There are many ways to use Git, many are redundant. The purpose of tab completion is to help users, not learn them to use the ‘only true way of using Git’ as there are lots of them.

dahlbyk:
I'm with @chillum here - if the user types `git diff --st<tab>` I would expect `--staged` to complete.

Thanks for the contribution!

Issue #621, opened 2018-09-11
 tab for "git push --force-with-lease" did not work 


chucklu:
### System Details

- posh-git version/path:
SHA-1: 836ec8c80f85085891c82bada9f586ad107fab32
- PowerShell version:
```powershell
Name                           Value
----                           -----
PSVersion                      6.1.0-rc.1
PSEdition                      Core
GitCommitId                    6.1.0-rc.1
OS                             Microsoft Windows 10.0.16299
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```
- Git version:
git version 2.18.0.windows.1
- Operating system name and version:
OS                             Microsoft Windows 10.0.16299

### Issue Description

When  I input "git push --force-" then tab, I got "git push --force-with-lease=".
The last character "=" make the command failed, after I remove "=", it works.
I have tried to search the "--force-with-lease"  in this repository, I find this test case:
https://github.com/dahlbyk/posh-git/blob/b428c3ceb8dd54f447d545b37db37636724cbe55/test/GitParamTabExpansion.Tests.ps1#L13

some other commands end with "=" was found, I am not sure if they worked or not
>    $result -contains '--exec=' | Should Be $true
>    $result -contains '--receive-pack=' | Should Be $true
>    $result -contains '--recurse-submodules=' | Should Be $true
>    $result -contains '--repo=' | Should Be $true
>    $result = & $module GitTabExpansionInternal 'git push --recurse-submodules='

dahlbyk:
If you have time, please submit a PR to remove the optional `=` from test and code. There may be other instances of `=` on parameters where it's really optional, too.

There are some parameters (e.g. `git merge --strategy=ours`) where the value is required, so we can't just strip all of them. For the most part our parameter lists were copied from [`git-completion.bash`](https://github.com/git/git/blob/c05048d43925ab8edcb36663752c2b4541911231/contrib/completion/git-completion.bash), which apparently handles the optional `=` differently for some commands?

chucklu:
Hi @dahlbyk ,
      I have a pull request to fix this issue, please verify it.

PR #622, opened 2018-09-12
remove the character "=" which follows "git push --force-with-lease"

chucklu:
fix issue https://github.com/dahlbyk/posh-git/issues/621


dahlbyk:
Thanks!

PR #627, opened 2018-10-04
Fix AppVeyor tests - local repo need user email/name before committing

rkeithhill:


test/Shared.ps1
```diff
@@ -73,6 +76,8 @@ function NewGitTempRepo([switch]$MakeInitialCommit) {
     Set-Location $repoPath
 
     if ($MakeInitialCommit) {
+        git config user.email "spaceman.spiff@appveyor.com"
+        git config user.name "Spaceman Spiff"
```
dahlbyk: Should this not use `$gitbin`?

rkeithhill: Yup.  Sorry about that. Guess I was in a bit too much of hurry on this PR.  New PR coming.

Issue #592, opened 2018-06-29
1.0 beta prompt incorrect with conemu


spooky:
Because of the ssh-agent/windows update problems, I've updated to 1.0.0 beta posh-git but that seems to break the prompt integration with conemu (technically, cmder version 1.3.6 which uses conemu 180528).

The git part is placed at the end (after lambda sign in the second line) instead of after the direcotry in the first line (as it was with 0.7.3)

dahlbyk:
Interesting, I wouldn't expect the console host to have any impact, assuming it's correctly handling ANSI control sequences. Can you share a screenshot, or text that demonstrates what's happening? Are you customizing `$GitPromptSettings` at all?

dahlbyk:
Or are you using oh-my-posh?

spooky:
using just posh-git, no customizations on the $GitPromptSettings. I should be able to share a screenshot around monday.

rkeithhill:
Can you post the output you get from running `$PoshGitLastPrompt`?

dahlbyk:
Does `$GitPromptSettings.AnsiConsole` return `false`? Try setting `$GitPromptSettings.AnsiConsole = $true`.

dahlbyk:
FYI, we do check a ConEmu-specific environment variable for determining ANSI support: https://github.com/dahlbyk/posh-git/blob/dcb96a4dfce1c0f4fc1af1670e50b03f9bc1b080/src/PoshGitTypes.ps1

CollinChaffin:
I have to disagree.  This broke how many of us in the community use it in conjunction with conemu because you completely destroyed backward compatibility from .73 to v1.0.  Anything we had in our profiles and prompts broke, and as a DEV myself for the life of me, I'm still trying to figure out why you went in that direction.

For instance, in .73 there was an actual example including various functions (like setting console titlebar for git etc.).  In version 1.0, however, you've now removed the example.ps1 for the profile and literally more than QUADRUPLED your lines of code to over 174 lines of code+text for a function named "Add-PoshGitToProfile" that literally does exactly what this SINGLE LINE of code does, no more - no less:

````
[...]

PR #628, opened 2018-10-04
Add CommandNotFoundAcion hook to give info on posh-sshell

rkeithhill:
You can override this check by defining an env var: POSHGIT_NO_SSH_CHECK

Remove NoVersionWarn module param - no longer used.
Convert last module param to [bool] since [switch] doesn't really work.

FWIW I'm not sure this is worth doing but if we really want to notify users of
the removal of the ssh utils, this is the way to do it IMO.

Partially address #592

### DIFF
--- a/src/posh-git.psm1
+++ b/src/posh-git.psm1
@@ -1,4 +1,4 @@
-param([switch]$NoVersionWarn, [switch]$ForcePoshGitPrompt)
+param([bool]$ForcePoshGitPrompt = $false)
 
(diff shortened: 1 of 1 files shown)


rkeithhill:
Hmm, the more I think about this, the less I'm in favor of doing this.  Seems like we should just prominantly indicate in the v1 readme that all the SSH commands have moved to posh-sshell and you will need to install/import that module to continue using those commands.

rkeithhill:
I don't think this is worth the added complexity.

Issue #635, opened 2018-10-24
Setting variables   without importing posh-git


musm:
I am using posh-git and it auto loads when entering a directory that is using git.

I'd like to customize several module variables, without having to import the module in my profile since this drastically slows down starting powershell.

Is this possible?

```
# enabling the following darastically slows down starting powershell

# Import-Module posh-git

# $GitPromptSettings.BeforeStatus = 'on '
# $GitPromptSettings.AfterStatus = ''

# enable this but makes pwsh load times slow!
# $GitPromptSettings.EnableStashStatus = $true
```

Thanks.

rkeithhill:
When I measure `Import-Module posh-git` I'm seeing about a 1 second load time which is something we should look into.  Is this about the same slow down you're seeing?  Also, what version of posh-git are you on? You can get that info if you run `gmo posh-git` after the module has been imported.

Unfortunately, what you're asking for will not work atm.  Before we come up with yet another way to provide settings, I'd like to see if we can improve the module load time.

musm:
Yes it's about 1 second, which substantially slows down loading the already slow to start pwsh v6 shell. I'm running the latest version of posh git available from the gallery.

rkeithhill:
There are two "latest" available.  Which one you got depends on whether or not you used the `-AllowPrerelease` flag with Install-Module.

musm:
```
> gmo posh-git

ModuleType Version    Name                                ExportedCommands
---------- -------    ----                                ----------------
Script     1.0.0      posh-git                            {Add-PoshGitToProfile, Ad...
```

rkeithhill:
I have a PR pending that I think should help shave ~.5 secs off the load time on pwsh.  Hopefully I can get the PR merged for the next beta.

musm:
any progress?

rkeithhill:
That PR made it into the current beta.  It improved load time a little.  There is another PR (#698) that is attempting to re-write our PSM1 file at build-time with all the dot-source files embedded into it (rather than dot-sourcing them).  That is supposed to speed up load time a bit more.

musm:
Good to know. Yes powershell startup times are still a big problem for me.  Any improvements here would for sure be welcome. Thanks for your work.

musm:
any status update here?

rkeithhill:
[...]

PR #638, opened 2018-11-01
On Windows PowerShell, defer Add-Type until Set-ConsoleMode executed

### LINKED ISSUE #637: Skip the ConsoleMode hack for new Windows versions
### System Details

- posh-git version/path: 1.0.0 beta2 ~\Documents\PowerShell\Modules\posh-git\1.0.0
- PowerShell version: 6.1.0
- git version 2.19.1.windows.1
- OS: Microsoft Windows NT 10.0.17763.0

### Issue Description

Thank you for this wonderful project!
Sadly though I'm a bit dissatisfied with the startup speed (1000-1300ms) and tried improving the situation.

While doing so I realized that if I remove these 2 lines:
* https://github.com/dahlbyk/posh-git/blob/653101c3aae676eafa0f2ebe6b20f066a8e4013d/src/posh-git.psm1#L5
* https://github.com/dahlbyk/posh-git/blob/653101c3aae676eafa0f2ebe6b20f066a8e4013d/src/GitPrompt.ps1#L847

everything still works as it used to - including ANSI escape codes!

*But now startup time is reduced by 30-50% down to around 500-600ms!*

Could this be because in the newest Windows versions `HKCU:\Console\VirtualTerminalLevel` is `1` by default now? Because I personally don't have that registry key set at all.

If we could add some kind of check for this we could skip running the [expensive ConsoleMode hack](https://github.com/dahlbyk/posh-git/blob/master/src/ConsoleMode.ps1#L11-L47), no-op the `Set-ConsoleMode` function and thus significantly improve startup time.

I'm fairly inexperienced in proper PowerShell scripting and thus I don't have a proper idea how we can do this in a backwards-compatible way, but I'm happy to help wherever I can. 🙂

rkeithhill:
On PS Core, do not set the console mode at all.
PS Core, since the release of 6.0.0 has handled setting/resetting the
console mode, so no need for posh-git to do this. See:
https://github.com/PowerShell/PowerShell/pull/2991

This speeds up module import by ~.5 to .7 seconds on PSCore and by
~150-200 msecs on Windows PowerShell.

Fix #637 and maybe #635

### DIFF
--- a/src/ConsoleMode.ps1
+++ b/src/ConsoleMode.ps1
@@ -1,6 +1,10 @@
 # Hack! https://gist.github.com/lzybkr/f2059cb2ee8d0c13c65ab933b75e998c
 
-if ($IsWindows -eq $false) {
+# Always skip setting the console mode on non-Windows platforms.
+# On Windows, *if* we are running in PowerShell Core *and* the host is the "ConsoleHost",
+# rely on PS Core to manage the console mode.  This speeds up module import on standard
+# PowerShell Core on Windows.
+if (($IsWindows -eq $false) -or (($PSVersionTable.PSVersion.Major -ge 6) -and ($host.Name -eq 'ConsoleHost'))) {
     function Set-ConsoleMode {
         [Diagnostics.CodeAnalysis.SuppressMessageAttribute("PSUseShouldProcessForStateChangingFunctions", "")]
         param()
--- a/src/ConsoleMode.ps1
+++ b/src/ConsoleMode.ps1
@@ -1,14 +1,18 @@
 # Hack! https://gist.github.com/lzybkr/f2059cb2ee8d0c13c65ab933b75e998c
 
-if ($IsWindows -eq $false) {
+# We do not need to set the console mode in PowerShell Core on any platform.
+# On Windows, PowerShell Core added this support in this PR:
+# https://github.com/PowerShell/PowerShell/pull/2991
+if ($PSVersionTable.PSVersion.Major -ge 6) {
     function Set-ConsoleMode {
         [Diagnostics.CodeAnalysis.SuppressMessageAttribute("PSUseShouldProcessForStateChangingFunctions", "")]
         param()
     }
+
     return
 }
 
-Add-Type @"
+$consoleModeSource = @"
 using System;
 using System.Runtime.InteropServices;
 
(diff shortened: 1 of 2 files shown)


src/ConsoleMode.ps1
```diff
@@ -1,14 +1,19 @@
 # Hack! https://gist.github.com/lzybkr/f2059cb2ee8d0c13c65ab933b75e998c
 
-if ($IsWindows -eq $false) {
+# Always skip setting the console mode on non-Windows platforms.
+# On Windows, *if* we are running in PowerShell Core *and* the host is the "ConsoleHost",
+# rely on PS Core to manage the console mode.  This speeds up module import on standard
+# PowerShell Core on Windows.
+if (($IsWindows -eq $false) -or (($PSVersionTable.PSVersion.Major -ge 6) -and ($host.Name -eq 'ConsoleHost'))) {
```
dahlbyk: This seems like a future extensibility point for other hosts that don't need `Set-ConsoleMode`, but this is good for now.

musm:
when will this be available in the beta?

dahlbyk:
I still have a few lingering PRs to catch up with, but I'll try to ship a new beta in the next week or so.

dahlbyk:
> I'll try to ship a new beta in the next week or so.

So, uh. Beta 3 shipped yesterday. 🙄

src/ConsoleMode.ps1
```diff
@@ -85,19 +90,29 @@ function Set-ConsoleMode
         $StandardInput
     )
 
-    if ($ANSI)
-    {
-        $outputMode = [NativeConsoleMethods]::GetConsoleMode($false)
-        $null = [NativeConsoleMethods]::SetConsoleMode($false, $outputMode -bor [ConsoleModeOutputFlags]::ENABLE_VIRTUAL_TERMINAL_PROCESSING)
+    begin {
```
dahlbyk: @rkeithhill What's the advantage of splitting this into `begin`/`end`?

rkeithhill: In this particular case, there is no advantage.  Then again, there is no harm either.  But I can simplify this if you'd like.

dahlbyk: No change necessary unless you find the alternative easier to read. Just curious if I was missing something.

PR #641, opened 2018-11-04
Add last checkouts to tabcompletion

mpawelski:
Hi, thanks for this great project!
I think this small feature will make it even better ;-)

I really like git [`@{-N}` syntax](https://stackoverflow.com/questions/7206801/is-there-any-way-to-git-checkout-previous-branch) to checkout last used branches. I thought that if posh-git could suggest this last used branches first we would have nice user experience. You could just `Tab` and `Shift+Tab` and quickly choose the recent branch you need to work on again. 

It would be very useful when you work on a lot of branches that have long names and you don't remember the name correctly (for example because it has Jira key at beginning, and who remembers that ;-)).

It looks that git [source code](https://github.com/git/git/blob/cd69ec8cde54af1817630331fc441f493866f0d4/sha1-name.c#L1255) does something very similar to what I have done here to get last checkout commits.

### DIFF
--- a/src/GitTabExpansion.ps1
+++ b/src/GitTabExpansion.ps1
@@ -101,6 +101,13 @@ function script:gitCommands($filter, $includeAliases) {
     $cmdList | Sort-Object
 }
 
+function script:lastCheckouts($filter){
+    git reflog |
+    Where-Object { $_ -match ': checkout: moving from (.*) to (.*)' } |
+    ForEach-Object { $Matches[1] } |
+    Where-Object { $_ -like "$filter*" }
+}
+
 function script:gitRemotes($filter) {
     git remote |
         Where-Object { $_ -like "$filter*" } |
@@ -387,6 +394,7 @@ function GitTabExpansionInternal($lastBlock, $GitStatus = $null) {
         # Handles git checkout <ref>
         "^(?:checkout).* (?<ref>\S*)$" {
             & {
+                lastCheckouts $matches['ref']
                 gitBranches $matches['ref'] $true
                 gitRemoteUniqueBranches $matches['ref']
                 gitTags $matches['ref']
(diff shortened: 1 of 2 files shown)


src/GitTabExpansion.ps1
```diff
@@ -101,6 +101,13 @@ function script:gitCommands($filter, $includeAliases) {
     $cmdList | Sort-Object
 }
 
+function script:lastCheckouts($filter){
+    git reflog |
```
dahlbyk: Two suggestions here:

1. Specify an explicit `--format=%gs` (I think?) to get only the log output you need. There might be other flags you can also specify to optimize the operation (e.g. `--no-decorate`).
2. `reflog` can get quite long under active development, so I would add a `--max-count=<number>` to avoid scanning thousands of references.

mpawelski: I came up with two options and I'm not sure which will be better 🤔
1) ` git log --walk-reflogs --max-count=100 --format=%gs`
get more number of reflogs and then search for "'checkout: moving from (.*) to (.*)'" pattern from result.
2) `git log --walk-reflogs --max-count=10 --grep-reflog='checkout: moving from ' --fixed-strings --format=%gs`
get less number of reflogs but already filtered to one with data that we need (containing "checkout: moving from" in reflog subject).

For me option 2) seems more tempting because you get less data from git command. However the performance can depend on how big is you reflog and how many entries it has that are not what we are looking for (we want only reflogs with "checkout: moving from" subject). In other words you don't know how many reflogs git will traverse to get you list of 10 reflogs you want.

I think 2) is fine. How many crazy operations (like rebases of huge number of commits) you would have to do between switching branch 10 times to have huge reflog that can affect performance?

mpawelski:
I ended up with implementing seconds options as it seams "good enough" for me (for example for my repo I work on now I have reflog with over 600 entries and without `--max-count=10` I get more that 130 entries back, so having `--max-count` and `--grep-reflog` seems like good approach).

BTW. I broke and fixed unit test that I wrote before, but I actually don't know why previous quoting broke unit test... When I executed both quote variants in terminal they run fine... 🤔

PR #649, opened 2018-12-06
Add manual install instructions to readme.

### LINKED ISSUE #648: How do you manually install this?
How do you manually install this on windows?

All of the instructions instruct me to use some package manager, but i see no reason to install a package manager for 1 package, and I have no reason to use them for any other package as this is the only mod to power shell i'm gonna want, as i do everything else in bash shell. (git doesn't work for me in bash shell for windows because it leaves files behind when you control+c out of an interactive commit and then further git operations fail until you manually remove those files, posh-git on github shell classic didn't have this issue and i'm hoping the same is true for posh-git proper)

MrStonedOne:
Adds the manual install instructions given by @rkeithhill on #648 to the readme.

closes #648

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,7 +148,7 @@ choco install poshgit
 
 If you need to test/debug changes prior to contrubiting here, or would otherwise prefer to install post-git without the aid of a package manager you can run `Import-Module POST-GET-PATH` where `POST-GET-PATH` is the path to the `src\posh-git.psd1` file.
 
-Say you have posh-git located at `$home\documents\github\posh-git`, you can install posh-git by running `Import-Module $home\documents\github\posh-git\src\posh-git.psd1`.
+Say you have posh-git located at `$home\documents\github\posh-git`, you can import posh-git by running `Import-Module $home\documents\github\posh-git\src\posh-git.psd1`.
 
 ## Using posh-git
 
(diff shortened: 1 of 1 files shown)


rkeithhill:
@dahlbyk It looks like the test failures are related to a change in PS Core where it add `[username@hostname]: ` to the prompt when remoted into a PSCore session via SSH.  Guess we could test for that somehow and modify the string compare to a regex match.

README.md
```diff
@@ -144,6 +144,12 @@ With Chocolatey installed, execute the following command to install posh-git:
 choco install poshgit
 ```
 
+### Installing post-git Manually
+
+If you need to test/debug changes prior to contrubiting here, or would otherwise prefer to install post-git without the aid of a package manager you can run `Import-Module POST-GET-PATH` where `POST-GET-PATH` is the path to the `src\posh-git.psd1` file.
```
rkeithhill: There's a typo on this line - `contrubiting` should be `contributing`.  Also rather than say "run" it might be better to say "execute".

README.md
```diff
@@ -144,6 +144,12 @@ With Chocolatey installed, execute the following command to install posh-git:
...
+
+Say you have posh-git located at `$home\documents\github\posh-git`, you can install posh-git by running `Import-Module $home\documents\github\posh-git\src\posh-git.psd1`.
```
rkeithhill: I wouldn't say "install posh-git by running".  Better to say "import posh-git by executing".

PR #652, opened 2018-12-17
Tests: Fix host/user prompt match over SSH

dahlbyk:
https://github.com/dahlbyk/posh-git/pull/649#issuecomment-445485454


test/DefaultPrompt.Tests.ps1
```diff
@@ -1,10 +1,14 @@
 . $PSScriptRoot\Shared.ps1
 
+$sshPrefix = '(\[[^\]+\]: )?'
```
rkeithhill: I think this should be `'(\[[^\]]+\]: )?'`

rkeithhill:
Fixed by PR #658

PR #663, opened 2019-02-19
Fix #79 - add Remove-MergedGitBranch command

### LINKED ISSUE #79: Remove remote merged branches command

rkeithhill:
Not sure if this command should require confirmation by default.

Perhaps we also need a Remove-GitBranch command which would need
confirmation.

Fix #79

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@
 
 ### Added
 
-- Added `Remove-GitBranch` and `Remove-MergedGitBranch` commands. Partially addresses #79.
+- Added `Remove-GitBranch` command. Partially addresses #79.
   These commands currently only delete local branches.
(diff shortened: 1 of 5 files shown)


rkeithhill:
Also, if using `--merged` I don't think you'd ever need to use `-D` (force delete), would you?

rkeithhill:
Also, wondering if this should be just `Remove-GitBranch` and then add a `-Merged` switch to change its behavior to remove only merged branches.  Or we could default to only deleting merged and add a `-All` switch to delete all branches except the default exclude pattern (excludes the current branch as well as master and develop).

PR #673, opened 2019-04-03
fix(status): Only reset changed colors

ExE-Boss:
This fixes a bug with using **posh‑git** inside a [**PowerLine** block](https://github.com/Jaykul/PowerLine/blob/master/README.md), which would result in the background colour being reset to the PowerShell window’s background colour, breaking the **PowerLine** block.

review?(@dahlbyk): I’ve been using this for the last 15 days, and it works fine for me.

### DIFF
--- a/src/AnsiUtils.ps1
+++ b/src/AnsiUtils.ps1
@@ -47,6 +47,12 @@ function Test-VirtualTerminalSequece([psobject[]]$Object, [switch]$Force) {
 }
 
 function Get-VirtualTerminalSequence ($color, [int]$offset = 0) {
+    # Don't output ANSI escape sequences if the `$color` parameter is `$null`,
+    # they would be broken anyway
+    if ($null -eq $color) {
+        return $null;
+    }
+
     if ($color -is [byte]) {
         return "${AnsiEscape}$(38 + $offset);5;${color}m"
     }
--- a/src/PoshGitTypes.ps1
+++ b/src/PoshGitTypes.ps1
@@ -157,16 +157,16 @@ class PoshGitTextSpan {
         [System.Collections.Generic.List[string]]$reset = @()
         $e = [char]27 + "["
 
-        $bg = $this.BackgroundColor
-        if (($null -ne $bg) -and !(Test-VirtualTerminalSequece $bg)) {
-            $bg = Get-BackgroundVirtualTerminalSequence $bg
-            $reset.Add('49')
-        }
-
         $fg = $this.ForegroundColor
         if (($null -ne $fg) -and !(Test-VirtualTerminalSequece $fg)) {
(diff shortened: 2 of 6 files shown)


rkeithhill:
That build failure seems to be because apt-get is failing to install PowerShell on Ubuntu 14.04.  Any chance we could migrate the build to 16.04 since 14.04 is EOL in 10 days?  Install instructions for 16.04 are here - https://docs.microsoft.com/en-us/powershell/scripting/install/installing-powershell-core-on-linux?view=powershell-6#installation-via-package-repository---ubuntu-1604

src/PoshGitTypes.ps1
```diff
@@ -153,25 +153,33 @@ class PoshGitTextSpan {
     # $GitPromptSettings.AnsiConsole is $true.  It is also used by the default ToString()
     # implementation to display any ANSI seqs when AnsiConsole is $true.
     [string] ToAnsiString() {
+        # If we know which colors were changed, we can reset only these and leave others be.
+        [System.Collections.Generic.List[string]]$reset = @()
```
rkeithhill: Minor nit, since we're on v5 and higher, I'd probably just call the ctor here:
```
$reset = [System.Collections.Generic.List[string]]::new()
```

src/AnsiUtils.ps1
```diff
@@ -47,6 +47,12 @@ function Test-VirtualTerminalSequece([psobject[]]$Object, [switch]$Force) {
 }
 
 function Get-VirtualTerminalSequence ($color, [int]$offset = 0) {
+    # Don't output ANSI escape sequences if the `$color` parameter is `$null`,
```
rkeithhill: Minor nit - probably don't need the markdown syntax in a comment.

src/PoshGitTypes.ps1
```diff
@@ -153,25 +153,33 @@ class PoshGitTextSpan {
...
 
         $txt = $this.Text
         $str = "${fg}${bg}${txt}"
 
         # ALWAYS terminate a VT sequence in case the host supports VT (regardless of AnsiConsole setting),
         # or the host display can get messed up.
-        if (Test-VirtualTerminalSequece $str -Force) {
-            $str += "${e}0m"
+        if (Test-VirtualTerminalSequece $txt -Force) {
+            $reset = @('0')
```
rkeithhill: Minor nit -
```
$reset.Clear()
$reset.Add('0')
```

src/AnsiUtils.ps1
ExE-Boss: Yeah, but I feel like
```pwsh
# Don't output ANSI escape sequences if the `$color` parameter is `$null`,
```
looks better than
```pwsh
# Don't output ANSI escape sequences if the $color parameter is $null,
```

PR #674, opened 2019-04-10
fix rebase step count

skajfes:
Fixed a small issue when rebase count wasn't displayed in prompt


rkeithhill:
Yeah, that looks like a copy/paste bug.

dahlbyk:
I distinctly remembering fixing the file names after copy/paste, but I clearly missed the folder. Thanks!

cc #599

PR #684, opened 2019-07-02
Add GitPromptValues global to allow error status display in prompt

rkeithhill:
It is useful to display the error status of the last command.  But in
order to do that we must preserve the value of $? and make the
original $LASTEXITCODE available to string expansion in the prompt.


dahlbyk:
Belated 👍

This is great!

rkeithhill:
Thanks!

Issue #695, opened 2019-09-10
Azure CLI replaces VSTS CLI


pinkfloydx33:
The VSTS CLI has been deprecated and replaced with the Azure CLI (`az repos`). 

The aliases that are added via [`--use-git-aliases`](https://docs.microsoft.com/en-us/azure/devops/cli/auto-detect-and-git-aliases?view=azure-devops) vary slightly from those added by the VSTS CLI (`pr` vs `vsts.pr`). Additionally, some of the [sub-commands](https://docs.microsoft.com/en-us/cli/azure/ext/azure-devops/repos/pr?view=azure-cli-latest) have changed; many of them are pluralization (`reviewer` vs `reviewers`) but some options have been removed and a few added.

Any chance the module will be updated to officially support the aliases added by the Azure CLI?

I've updated my local copy of the module and would gladly submit a PR if you'd like.

dahlbyk:
> I've updated my local copy of the module and would gladly submit a PR if you'd like.

That would be great!

Is it reasonable to retain support for `vsts` and add parallel support for `az devops` aliases?

pinkfloydx33:
Yes we can support both aliases. I had to make a few tweaks to my local after you said that, but its easily done. The alias contents are different so we can key off KnownAliases and it works fine. You'll never have both installed, but if you did for some reason they shouldn't conflict.

I also added support for those commands that have sub-commands; for example `git pr policy list - -params` with policy having sub-commands (list, etc). There are a few others: reviewers and work-items. I did not back-port that to the Vsts version however, so azure cli only.

I'll have a PR likely later today. I've also added support for #691. Should I submit two seperate PRs?

dahlbyk:
> Should I submit two seperate PRs?

Definitely separate commits, but a single PR is fine.

PR #696, opened 2019-09-12
Add Support for Azure CLI #695

pinkfloydx33:
Adds Support for Azure CLI #695 
- Includes Support for 
   - Sub-Command Expansion
   - Target/Source Branch expansion
   - Parameter=Value expansion
- Updates VSTS CLI Support
   - Fills in gaps for missing commands
   - Feature parity with AZ (branch/parameter expansion)
- Both AZ/VSTS can be installed at the same time (with different aliases)
- Added Tests for all cases


bergmeister:
Awesome! For the old `vsts` CLI I've already created a module but adding the `az` CLI would be great (although I'm not sure if it should rather be its own standalone module)
https://github.com/bergmeister/posh-vsts-cli

Issue #692, opened 2019-09-01
Consider using module builder pattern to halve module import time


bergmeister:
Whilst I like `posh-git`, it also adds a big portion (500-1000ms) to the load time of PowerShell because of the call to `Import-Module posh-git` in the $PROFILE.
One possibility of halving (I've done an experiment) the module import time, is to have all the source code in one monolithic file instead of dot sourcing them in. This does not mean the source code of the module needs to be stored in a monolithic file but one would need to add a build process before publishing the module, one does not need to re-invent the wheel though as one could e.g. use the [ModuleBuilder](https://github.com/PoshCode/ModuleBuilder) module for that.
Would you be happy with that?
Or would it be useful to look into alternative ways of having a more lightweight module import (some user's might not use all features, so maybe we could defer some actions?)?

rkeithhill:
I don't see a problem with adding a **build**  for the module particularly if it can halve the module import time.  As for as "optional" features, that might be tough as this module doesn't offer that much in the way of optional features.  That said, I suppose there might be some folks who want only the tab-completion support.  Doubt there are many that want only the prompt status and not tab-completion though.

dahlbyk:
> but one would need to add a build process before publishing the module, one does not need to re-invent the wheel though as one could e.g. use the [ModuleBuilder](https://github.com/PoshCode/ModuleBuilder) module for that. Would you be happy with that?

We've discussed elsewhere that a pre-publish build process would be useful in other ways, so I'm not opposed to this as a related optimization. I would want to make sure the module can still be loaded as-is from the Git repo, as many folks use their own fork of the project.

> As for as "optional" features, that might be tough as this module doesn't offer that much in the way of optional features. That said, I suppose there might be some folks who want only the tab-completion support. Doubt there are many that want only the prompt status and not tab-completion though.

We could conceivably skip loading all the prompt/status infrastructure if `$GitPromptSettings.EnablePromptStatus -eq $false`, though technically that setting only disables the Git portion of the prompt, not the rest of it (path, etc).

bergmeister:
Good feedback.
Should I aim to make a PR against the `v0` or `master` branch (or both?). Personally, I still use v0 (accidentally) sometimes because I often forget the `-AllowPreRelease` switch...

rkeithhill:
I'd say `master` as I'd like to nudge folks towards v1.  :-)

bergmeister:
[...]

PR #697, opened 2019-09-15
v0: Add build script for creating module with scripts being merged into it

bergmeister:
Related: #692
This targets the `v0` branch.
The AppVeyor build has a problem is a known issue where AppVeyor does not see that the target branch is not master and cannot deal with that....


bergmeister:
Will update this later once the main PR has been merged

PR #698, opened 2019-09-15
Add build module

bergmeister:
Related: #692
This targets the `master` branch.

It creates the module in an `out` folder by copying the content of the `src` folder into it first and then merging the `.ps1` files into the main `.psm1` files that are dot sourced by looking for the pattern `$PSScriptRoot\ScriptName.ps1` and replacing it with the content and wrapping it into a `#region` block.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -19,9 +19,10 @@ foreach ($powerShellScriptToBeMerged in $PowerShellScriptsToBeMerged) {
     if ($contentOfMainModule.Contains($replaceSearchText)) {
         $contentOfMainModule = $contentOfMainModule.Replace(". `$PSScriptRoot\$NameOfScriptToBeMerged",
             @"
-#region
+#region $($powerShellScriptToBeMerged.BaseName)
 $scriptContentToBeMerged
 #endregion
+
 "@)
         Remove-Item $powerShellScriptToBeMerged
     }
--- a/build.ps1
+++ b/build.ps1
@@ -1,0 +1,32 @@
+$outputFolder = (Join-Path $PSScriptRoot 'out')
+
+# Clean
+if (Test-Path $outputFolder) {
+    Remove-Item $outputFolder -Recurse
+}
+
+# Copy files to output folder
+Copy-Item -Path (Join-Path $PSScriptRoot 'src') -Destination $outputFolder -Recurse
+
+$mainModulePath = (Join-Path $outputFolder 'posh-git.psm1')
+$contentOfMainModule = Get-Content -Path $mainModulePath
+
+$PowerShellScriptsToBeMerged = Get-ChildItem -Path $outputFolder -Filter '*.ps1'
+foreach ($powerShellScriptToBeMerged in $PowerShellScriptsToBeMerged) {
+    $NameOfScriptToBeMerged = $powerShellScriptToBeMerged.Name
+    $scriptContentToBeMerged = Get-Content -Path $powerShellScriptToBeMerged -Raw
+    $replaceSearchText = ". `$PSScriptRoot\$NameOfScriptToBeMerged"
+    if ($contentOfMainModule.Contains($replaceSearchText)) {
+        $contentOfMainModule = $contentOfMainModule.Replace(". `$PSScriptRoot\$NameOfScriptToBeMerged",
+            @"
+#region
+$scriptContentToBeMerged
+#endregion
+"@)
+        Remove-Item $powerShellScriptToBeMerged
+    }
+}
+
+Set-Content -Path $mainModulePath -Value $contentOfMainModule
+
+Write-Verbose "Module created at '$outputFolder'" -Verbose


rkeithhill:
Rather than put this in a top-level `out` dir, can you read the module version number from the psd1 file and create/build into that dir e.g.  `.\1.0.0.0`.  That would satisfy @dahlbyk desire to make it possible to put the repo path in `$env:PSModulePath` and have import-module work.  You'll need to add this to the `.gitgnore` file:
```
# ignore release dirs
/*.*.*
```

bergmeister:
Awesome, I've applied both your suggestions. Shall I adapt the build as part of this PR as well to test against the built version?

rkeithhill:
That would be great.  Maybe add build script parameters along the lines of  `-BuildOnly` and `-TestOnly`?

bergmeister:
For the moment I make it conditionally use the built folder based on the existence of the `CI` environment variable for the moment just to see if it works.
Do we want to run tests twice now (before and after build)?

test/Shared.ps1
```diff
@@ -1,4 +1,9 @@
 $modulePath = Convert-Path $PSScriptRoot\..\src
+# Test against built folder in CI (or just set $env:CI locally for that)
+if ($env:CI) {
```
rkeithhill: This works on a build machine. Can we also add a PowerShell variable check here like:
``
if ($env:CI -or $PoshGitBuild)
```
That way if we build & test from the build.ps1 script on a dev machine, it will run the Pester tests against the release dir.  And because that $PoshGitBuild variable goes away after the build script exits, I can still run tests from the src folder module when I doing dev work.

**Comment by bergmeister** [comment]
Where is the definition of the Travis build?

**Comment by bergmeister** [comment]
@rkeithhill The PR is now ready to review, I've update the PR summary as well.

**Comment by bergmeister** [comment]
PR is ready to review now

**Comment by rkeithhill** [comment]
I'm seeing about a 176 out of 1242 mS improvement on startup or about a 14% improvement.  That's good but not as good as I'd hoped.  Here's how I tested:
```
C:\Users\Keith
10-13 13:58:21 4> $res = @(); for ($i=0; $i -lt 100; $i++) { $res += Measure-Command { pwsh -nop -wd $home\GitHub\rkeithhill\vscode-powershell -command "& { import-module $home\GitHub\dahlbyk\posh-git\src\posh-git.psd1}" }; Start-Sleep -Seconds 3 }; $res | Measure-Object -Property TotalMilliseconds -AllStats                                                
Count             : 100
Average           : 1242.651353
Sum               : 124265.1353
Maximum           : 2306.8213
Minimum           : 1155.3381
StandardDeviation : 174.263883117032
Property          : TotalMilliseconds


C:\Users\Keith
10-13 14:11:37 5> $res = @(); for ($i=0; $i -lt 100; $i++) { $res += Measure-Command { pwsh -nop -wd $home\GitHub\rkeithhill\vscode-powershell -command "& { import-module $home\GitHub\bergmeister\posh-git\1.0.0.0\posh-git.psd1}" }; Start-Sleep -Seconds 3 }; $res | Measure-Object -Property TotalMilliseconds -AllStats                                        
Count             : 100
Average           : 1066.09344
Sum               : 106609.344
Maximum           : 1642.2115
Minimum           : 1007.3209
StandardDeviation : 87.9550066181624
Property          : TotalMilliseconds
```
That said, I'm still in favor of merging this.  

@dahlbyk What do you think?  This impacts your build & release process.

bergmeister:
Yes, I also found the speed-up when the import happens in $PROFILE or directly after the startup is only 10-20%. When the import happened in an already open PowerShell session manually, then the the import was twice as fast. I can only guess that there are other threads that are still active immediately after PS startup that reduce the resources available for the speedup and therefore lead to a smaller speedup.
But since it is still quite a big improvement equivalent to the startup time of pwsh itself, it's worthwhile IMHO and with computers getting even more cores in the future, the relative speedup of this optimisation will probably increase more with the number of cores.

musm:
@bergmeister  thanks for improving startup times! It has been annoying with powershell 6, which is twice as to load than powershell 5.

rkeithhill:
Also helps with #635

musm:
bump? Would be nice to improve startup times if this helps with that.

eluchsinger:
Can we merge this in? I don't know why, but my PS Core takes long to start when I have posh-git installed.

bergmeister:
@rkeithhill @dahlbyk Can you please speedup the review process or if one of you two is busy I expect the other person to take ? It's been nearly a year now! I would not want to do it but otherwise I will consider publishing a forked version or have a separate repo with only the build/publishing process to avoid having to pull in upstream changes....

rkeithhill:
You know, some people actually do go on vacation.  :-)  But I'm back now.  That said, I don't do the build/final publish of the module.  That's something that only @dahlbyk can do AFAIK.  So I've left this PR up to him to approve/merge.

bergmeister:
> You know, some people actually do go on vacation. :-) But I'm back now. That said, I don't do the build/final publish of the module. That's something that only @dahlbyk can do AFAIK. So I've left this PR up to him to approve/merge.

Yes I appreciate and respect that people go on holiday or are being busy for a while due to personal/work reasons, especially since this is an OSS project without company backing. But a year is what is called a sabbatical not vacation ;-)

rkeithhill:
Understood. I've already approved this but @dahlbyk needs to as well and he's a very busy fellow these days. 🤷‍♂️

dahlbyk:
Burnout. Worse this year than in the past.

I have the few weeks off work, so hoping to dig back into posh-git during that time. Appreciate your contribution and patience.

bergmeister:
Sorry to hear that and recovery has priority in this case @dahlbyk. How would you feel about handing the maintenance/ownership over to @rkeithhill (assuming he'd be happy with that) to help you relieve from the Although you own the project, delegating work and increasing the trusting people is something that both you and the community would benefit from.

musm:
I do want to mention that the extremely slow pwsh load times already is a big grievance for powershell users (see this 2 year old issue https://github.com/PowerShell/PowerShell/issues/6443#issuecomment-659534629).  I use posh-git everyday, but it adds a painful 0.5s to 1s to my load time. For example see also this user with similar reports:  https://github.com/PowerShell/PowerShell/issues/6443#issuecomment-656282020

I do want to thank the work of the development team in trying to improve the situation.

rkeithhill:
@musm To set expectations if this PR is merged, the perf improvement is only in the range of 10-20%.

PR #699, opened 2019-09-16
Make SSH_AGENT_PID mapping to Windows PID

technic:
Fixes https://github.com/dahlbyk/posh-sshell/issues/21
PID and WINPID mismatch was introduced in the recent cygwin release,
see https://github.com/git-for-windows/git/issues/2274

### DIFF
--- a/src/GitUtils.ps1
+++ b/src/GitUtils.ps1
@@ -433,7 +433,7 @@ function Get-NativeSshAgent {
         # The ssh.exe binary version must include "OpenSSH"
         # The windows ssh-agent service must exist
         $service = Get-Service ssh-agent -ErrorAction Ignore
-        $executableMatches = Get-Command ssh.exe | ForEach-Object FileVersionInfo | Where-Object ProductVersion -match OpenSSH
+        $executableMatches = Get-Command ssh.exe -ErrorAction Ignore | ForEach-Object FileVersionInfo | Where-Object ProductVersion -match OpenSSH
         $valid = $service -and $executableMatches
         if ($valid) {
             return $service;
--- a/src/GitUtils.ps1
+++ b/src/GitUtils.ps1
@@ -401,14 +401,26 @@ function Get-SshAgent() {
     else {
         $agentPid = $Env:SSH_AGENT_PID
         if ($agentPid) {
-            $sshAgentProcess = Get-Process | Where-Object { ($_.Id -eq $agentPid) -and ($_.Name -eq 'ssh-agent') }
-            if ($null -ne $sshAgentProcess) {
-                return $agentPid
+            # Convert cygwin PID to Windows PID
+            $ps = Find-Ssh('ps')
+            if (!$ps) {
+                Write-Warning 'Could not find ps'
+                return 0
             }
-            else {
-                setenv 'SSH_AGENT_PID' $null
-                setenv 'SSH_AUTH_SOCK' $null
+            $pidMap = @{ }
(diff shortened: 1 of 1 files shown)


technic:
Hi, for my personal use I am fine with 0.7 branch, but would you like me to submit this to the 1.0 aka master branch?

rkeithhill:
@technic SSH functionality has been removed from the 1.0 branch (`master`) and placed in a separate `posh-sshell` [repo](https://github.com/dahlbyk/posh-sshell).

technic:
> 
> 
> @technic SSH functionality has been removed from the 1.0 branch (`master`) and placed in a separate `posh-sshell` [repo](https://github.com/dahlbyk/posh-sshell).

But 0.7 needs fix as well, because it is the stable branch now

rkeithhill:
I'd like to close this now that v1 has shipped. Any objections?

technic:
@rkeithhill Hey, since you didn't merge it 1.5 years ago, I don't think it makes sense to close it now. However, for me this patch still works relatively fine.

rkeithhill:
Thanks for the reply!

hazendaz:
I'm so haunted.  I finally was able to upgrade at work to the most recent software here (posh-git 1.1.0 and posh-sshell 0.3.1) and this bit me hard.  Spent few hours trying to figure out what was going on.  Finally realized the PID was not seen in windows.  My git bash equivalent continued working like a champ.  I realize the original author here from some PRs doesn't use or have time for this any more but some of us don't have modern enough windows even still in 2024 (love corporate!) to not have this and have to rely on git's ssh client.  This should be reopened and merged here on 0.7.x line and it should also be raised on posh-sshell and applied there.  Its a simple port to the 'Agent.ps1' identically.  Thankfully after running back into this my memory came back and I went and looked and I had documented exactly applying this patch back in March of 2021 pointing to this very PR.  So for 4 years, few hundred devs have been able to work smoothly in powershell so I think its worth having!  And anyone that does have same situation that tries to upgrade is going to run into this and waste hours for something so easily addressed.  And shame on me, I had actually documented this in a readme really well as well as the code but 3 years later I forgot about it.

However, that all said, this is not solely enough.

Two additional lines need patched

```
$sshLocation = join-path $gitItem.directory.parent.fullname bin/$program
$sshLocation = join-path $gitItem.directory.parent.fullname usr/bin/$program
```

Both in 2021 and now in 2024 - using windows 2016, those lines return 'ps' path twice.  This fix then and now worked.  The fix needs to be on posh-sshell as well.

```
$sshLocation = join-path $gitItem.directory.parent.fullname[0] bin/$program
$sshLocation = join-path $gitItem.directory.parent.fullname[0] usr/bin/$program
```

dahlbyk:
I don't use SSH anymore, but glad to merge/release PRs here and in posh-sshell.

hazendaz:
@dahlbyk Perfect!  I'll get a PR up to the posh-sshell this weekend and I'll create a new PR off this one to add the extra two lines that needed fixed.  I'll likely drop the ignore item added here as I didn't take that in 2021 and never missed it.

hazendaz:
Note: Both my older usage off 0.7.x and usage from this past week off posh-sshell where based off the released versions.  I understand releasing is unlikely and for our usage since we rebundle, we don't have users installing the normal way anyways.  So I plan to re-take latest masters off both and work forwards.  I can drop back and also do a bit more testing with 0.7.x.  Longer plan there will be to start reviewing merit of any open PRs that are not yet applied to see what else might be beneficial given there are a good number of them.

PR #702, opened 2019-09-21
Add tab expansion support for git restore / switch

### LINKED ISSUE #691: Support completion for new `git switch` and `git restore` commands
Git 2.23 brings some new features which include two new commands: `git switch` and `git restore`, which are meant to eventually provide a better interface for `git checkout`.

https://github.blog/2019-08-16-highlights-from-git-2-23/#experimental-alternatives-for-git-checkout

rkeithhill:
Fix #691

### DIFF
--- a/src/GitParamTabExpansion.ps1
+++ b/src/GitParamTabExpansion.ps1
@@ -28,12 +28,14 @@ $shortGitParams = @{
     rebase = 'm s X S q v n C f i p x'
     remote = 'v'
     reset = 'q p'
+    restore = 's p W S q m'
     revert = 'e m n S s X'
     rm = 'f n r q'
     shortlog = 'n s e w'
     stash = 'p k u a q'
     status = 's b u z'
     submodule = 'q b f n N'
+    switch = 'c C d f m q t'
     tag = 'a s u f d v n l m F'
     whatchanged = 'p'
 }
@@ -70,6 +72,7 @@ $longGitParams = @{
     reflog = 'stale-fix expire= expire-unreachable= all updateref rewrite verbose'
     remote = 'verbose'
     reset = 'patch quiet soft mixed hard merge keep'
+    restore = 'source= patch worktree staged quiet progress no-progress ours theirs merge conflict= ignore-unmerged ignore-skip-worktree-bits overlay no-overlay'
     revert = 'edit mainline no-edit no-commit gpg-sign signoff strategy= strategy-option continue quit abort'
     rm = 'force dry-run cached ignore-unmatch quiet'
     shortlog = 'numbered summary email format='
@@ -77,6 +80,7 @@ $longGitParams = @{
     stash = 'patch no-keep-index keep-index include-untracked all quiet index'
     status = 'short branch porcelain long untracked-files ignore-submodules ignored column no-column'
     submodule = 'quiet branch force cached files summary-limit remote no-fetch checkout merge rebase init name reference recursive depth'
+    switch = 'create force-create detach guess no-guess force discard-changes merge conflict= quiet no-progress track no-track orphan ignore-other-worktrees recurse-submodules no-recurse-submodules'
     tag = 'annotate sign local-user force delete verify list sort column no-column contains points-at message file cleanup'
     whatchanged = 'since'
 }
@@ -173,6 +177,14 @@ $gitParamValues = @{
     rebase = @{
         strategy = 'resolve recursive octopus ours subtree'
     }
+    restore = @{
+        conflict = 'merge diff3'
+        source = {
+            param($ref)
+            gitBranches $matches['ref'] $true
+            gitTags $matches['ref']
+        }
+    }
     revert = @{
         strategy = 'resolve recursive octopus ours subtree'
     }
@@ -184,5 +196,8 @@ $gitParamValues = @{
     status = @{
         'untracked-files' = 'no normal all'
         'ignore-submodules' = 'none untracked dirty all'
+    }
+    switch = @{
+        conflict = 'merge diff3'
     }
 }
--- a/src/GitTabExpansion.ps1
+++ b/src/GitTabExpansion.ps1
@@ -54,8 +54,8 @@ function script:gitCmdOperations($commands, $command, $filter) {
 $script:someCommands = @('add','am','annotate','archive','bisect','blame','branch','bundle','checkout','cherry',
                          'cherry-pick','citool','clean','clone','commit','config','describe','diff','difftool','fetch',
                          'format-patch','gc','grep','gui','help','init','instaweb','log','merge','mergetool','mv',
-                         'notes','prune','pull','push','rebase','reflog','remote','rerere','reset','revert','rm',
-                         'shortlog','show','stash','status','submodule','svn','tag','whatchanged', 'worktree')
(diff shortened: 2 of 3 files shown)


src/GitTabExpansion.ps1
```diff
@@ -374,8 +380,15 @@ function GitTabExpansionInternal($lastBlock, $GitStatus = $null) {
             gitAddFiles $GitStatus $matches['files']
         }
 
-        # Handles git checkout -- <path>
-        "^checkout.* -- (?<files>\S*)$" {
+        # Handles git restore -s <ref> - must come before the next regex case
+        "^restore.* -s\s*(?<ref>\S*)$" {
+            gitBranches $matches['ref'] $true
+            gitTags $matches['ref']
+            break
```
rkeithhill: I wonder if more of these `switch` cases should be using `break`.  Once we've found a "match" do we need to continue with every subsequent regex match?

pinkfloydx33: I was actually running into issues with this case myself and ended up matching --staged somehow, and kinda gave up and forgot to look back at it.  That said, I think this looks semi-correct. I think this will match "-S"  (which is the short-form of staged) followed by anything since we aren't using a case-insensitive match

src/GitTabExpansion.ps1
```diff
@@ -374,8 +380,15 @@ function GitTabExpansionInternal($lastBlock, $GitStatus = $null) {
...
+        }
+
+        # Handles git checkout -- <path> and git restore <path>
+        "^(?:checkout.* --|restore.*) (?<files>\S*)$" {
             gitCheckoutFiles $GitStatus $matches['files']
```
pinkfloydx33: I don't think this is correct; unless I'm reading things wrong it's missing support for `--staged`/`-S` . Even in that case,  files which qualify for a restore are more than just those that qualify for the "old" `checkout` -- `restore` is a bit more complex, enough-so that I pulled out a new function `gitRestoreFiles` for it. I don't have my laptop with me and unfortunately didn't push my latest updates to my fork, but best I can recall both with and without --staged could restore a *mixture* of Index and Working files. I originally had something very similar until I realized the files being auto-completed didn't match what git was suggesting. I'll check my local when I get into the office tomorrow.

src/GitTabExpansion.ps1
pinkfloydx33: @rkeithhill  re: `break` ... when working on my version of this as well as #696 I realized that the fall through would end up matching some of the later cases per the regex, but then return nothing based on the inner function calls. (lots of Write-Host debugging :-p) and was going to suggest we throw in breaks. With that said, I swear I encountered at least one instance where the fall-through actually caught an edge-case.

rkeithhill: >  I swear I encountered at least one instance where the fall-through actually caught an edge-case.

That's why I'm nervous about throwing in `break` everywhere.  But in this new case, it should break IMO.

RE the above code matching on `-S`, good catch.  I'll make this regex be case-sensitive.

src/GitTabExpansion.ps1
rkeithhill: Unless I'm missing something, according to these docs - https://git-scm.com/docs/git-restore/2.23.0 - `-S` and `--staged` appear to behave like switch parameters (taking no value).  As far as files to restore the docs mention only `<pathspec>` which led me to believe it expects file paths.

pinkfloydx33: You are correct that those are switch parameters. You are also correct that `git restore` takes a `<pathspec>`, however what _qualifies_ as a valid pathspec seems to vary based on the presence/absence of the `--staged`/`-S`  switch. This is based on my somewhat extended use of the feature thus far:

- Presence of the `--staged` switch: 
  - Index.Modified
  - Index.Added
  - Index.Deleted
- Absence of the `--staged` switch: 
  - Working.Unmerged
  - Working.Modified
  - Working.Deleted

Maybe I'm missing something (or perhaps over-complicating it). However, I wasn't able to get the tab expansion of actual files in my tree to match the options git was presenting until I had limited to the above.

rkeithhill: Interesting.  I'm not sure we've done such a "stateful" tab-completion with this setup. I've done it using the built-in `Register-ArgumentCompleter` facilities.  But it should be doable.  Basically the difference is that with `--staged` you add in the "Added" files and without you add in the "Unmerged" files.  I think I can do that by capturing the `.*` in this case and examining it for the presence of `--staged|-S`.

pinkfloydx33: We do something similar when calling `gitDiffFiles` which also has a staged parameter. This is what I had come up with:

```
function script:gitRestoreFiles($GitStatus, $filter, $staged) {
    if ($staged) {
        gitFiles $filter (@($GitStatus.Index.Modified) +  @($GitStatus.Index.Added)+@($GitStatus.Index.Deleted))
    } else {
        gitFiles $filter (@($GitStatus.Working.Unmerged) + @($GitStatus.Working.Modified) + @($GitStatus.Working.Deleted))
    }
}
```
and

```
        "^restore(?:.* (?<staged>(?:--staged))|.*) (?<files>\S*)$" {
            gitRestoreFiles $GitStatus $matches['files'] $matches['staged']
        }
```

Though the regex might need tweaking, as I said I was having some difficulties with the -s and -S and I can't remember now if I fixed it completely.

rkeithhill:
@pinkfloydx33 Updated.  Can you take another peek?

src/GitTabExpansion.ps1
```diff
@@ -380,16 +389,21 @@ function GitTabExpansionInternal($lastBlock, $GitStatus = $null) {
...
             gitBranches $matches['ref'] $true
             gitTags $matches['ref']
             break
         }
 
-        # Handles git checkout -- <path> and git restore <path>
-        "^(?:checkout.* --|restore.*) (?<files>\S*)$" {
-            gitCheckoutFiles $GitStatus $matches['files']
+        # Handles git restore <path>
+        "^restore(?:.* (?<staged>(?:-S|--staged))|.*) (?<files>\S*)$" {
```
pinkfloydx33: This needs the case sensitive check but only for the `-S` which would require an additional non-capturing group. Not a big deal other than readability. Other than that LGTM

[ I posted this comment over a week ago, but apparently I hit 'start a review' instead of add comment. I noticed this morning that GitHub was telling me that my comment was still pending(?). I cancelled it and re-posted as a comment. I admit that I don't use GitHub much (we use Azure git) so I'm not too familiar with the UI. Was that the *right* thing to do? ]

rkeithhill:
@pinkfloydx33 I'm going to merge this.  Can you pull master and try it out.  If you find any issues, we can fix them before the final release.

src/GitParamTabExpansion.ps1
```diff
@@ -173,6 +177,14 @@ $gitParamValues = @{
     rebase = @{
         strategy = 'resolve recursive octopus ours subtree'
     }
+    restore = @{
+        conflict = 'merge diff3'
+        source = {
+            param($ref)
+            gitBranches $matches['ref'] $true
```
dahlbyk: I don't always review code, but when I do it's six months later. 💪 

Did you mean to use `$ref` here?

rkeithhill: Yup, good catch.  I'll submit a PR shortly.

PR #711, opened 2019-11-06
Use Register-ArgumentCompleter to >= PS 6

rkeithhill:
A slightly different approach to PR #609


rkeithhill:
I'm going to merge this so I can use it for a while to see how it behaves.

Issue #719, opened 2019-11-25
Feature request: abbreviate path from git root


elsassph:
### System Details

- posh-git version/path: 1.0.0 beta3 ~/.local/share/powershell/Modules/posh-git/1.0.0
- PowerShell version: 6.2.3
- git version 2.21.0 (Apple Git-122.2)
- OS: Unix 18.7.0.0

### Issue Description

My abbreviated path can be pretty long and maybe there could be an option to abbreviate based on the closest git root (including the folder name), so it would essentially show the path within the project I'm working.

e.g.
```
~/company/work/projects/client/project-name/src [master] >
```
Could be further shortened to:
```
-/project-name/src [master] >
```

elsassph:
What are the contributions guidelines? Has this feature a chance to be accepted?
cc @dahlbyk @rkeithhill

rkeithhill:
We should probably have some contribution guidelines.  Short version, following the existing coding style in the function/script you're modifying.  If there are tests for the function you are modifying, please add new Pester tests to cover the changed/new functionality.  All tests must pass before PR would be accepted.  And yes, IMO I would accept this feature.  My only question is would it be enough to add an optional parameter to the `Get-PromptPath` function which would require the user to have to change the `$GitPromptSettings.DefaultPromptPath` setting.  Or should we also add yet another setting to enable this feature?  Could go either way.

elsassph:
@rkeithhill well I also did a PR adding an extra option for it. CI is green, but I'll look into adding a test specific to it. 

Maybe the path manipulations would be worth extracting into more easily testable functions BTW, unless that would add noise (I'm not familiar with all subtleties of pwsh).

dahlbyk:
Closed by #720. Thanks for the contribution!

PR #720, opened 2019-11-25
Add option to abbreviate based path based on the git root

elsassph:
Implement #719. Added an option to abbreviate the prompt path based on the git root.

When enabled
```
$GitPromptSettings.DefaultPromptAbbreviateGitDirectory = $true
```
When the currentpath is under a git repository, e.g.
```
~/company/work/projects/client/project-name/src [master] >
```
Prompt shows a path shortened to:
```
# using an arbitrary token to indicate the shortening
-/project-name/src [master] >
# or maybe using git `:/` root convention
project-name:/src [master] >
```

### DIFF
--- a/src/PoshGitTypes.ps1
+++ b/src/PoshGitTypes.ps1
@@ -283,6 +283,7 @@ class PoshGitPromptSettings {
     [PoshGitTextSpan]$DefaultPromptSuffix       = '$(">" * ($nestedPromptLevel + 1)) '
 
     [bool]$DefaultPromptAbbreviateHomeDirectory = $true
+    [bool]$DefaultPromptAbbreviateGitDirectory  = $false
     [bool]$DefaultPromptWriteStatusFirst        = $false
     [bool]$DefaultPromptEnableTiming            = $false
     [PoshGitTextSpan]$DefaultPromptTimingFormat = ' {0}ms'
--- a/src/Utils.ps1
+++ b/src/Utils.ps1
@@ -301,15 +301,14 @@ function Get-PromptPath {
         if ($gitPath) { $gitPath = Split-Path $gitPath -Parent }
     }
 
-    # Abbreviate path under a git repository
+    # Abbreviate path under a git repository as "<repo-name>:<relative-path>"
     if ($abbrevGitDir -and $gitPath -and $currentPath.StartsWith($gitPath, $stringComparison)) {
-        # Up another level to keep repo name in the abbreviated path
-        # $removePath = Split-Path $gitPath -Parent
         $gitName = Split-Path $gitPath -Leaf
-        $relPath = if ($currentPath -eq $gitPath) { "/" } else { $currentPath.SubString($gitPath.Length) }
+        $relPath = $currentPath.SubString($gitPath.Length)
+        if ($relPath -eq "") { $relPath = [System.IO.Path]::DirectorySeparatorChar }
         $currentPath = "$gitName`:$relPath"
     }
-    # Abbreviate path under the user's home dir
+    # Abbreviate path under the user's home dir as "~<relative-path>"
     elseif ($abbrevHomeDir -and $currentPath.StartsWith($Home, $stringComparison)) {
         $currentPath = "~" + $currentPath.SubString($Home.Length)
     }
--- a/src/Utils.ps1
+++ b/src/Utils.ps1
@@ -293,6 +293,7 @@ function Get-PromptPath {
     # Look up the git root
     if ($abbrevGitDir) {
         $gitPath = $(Get-GitDirectory)
+        # Up one level from `.git`
         if ($gitPath) { $gitPath = $(Split-Path $gitPath -Parent) }
     }
 
@@ -301,6 +302,7 @@ function Get-PromptPath {
     # Abbreviate path by replacing beginning of path with - *iff* the path is under a git repository
     if ($abbrevGitDir -and $currentPath -and $gitPath -and
         $currentPath.StartsWith($gitPath, $stringComparison)) {
+        # Up another level to keep repo name in path
         $removePath = $(Split-Path $gitPath -Parent)
         $currentPath = "-" + $currentPath.SubString($removePath.Length)
     }
--- a/src/Utils.ps1
+++ b/src/Utils.ps1
@@ -280,8 +280,7 @@ function Get-PathStringComparison {
 
 function Get-PromptPath {
     $settings = $global:GitPromptSettings
-    $abbrevHomeDir = $settings -and $settings.DefaultPromptAbbreviateHomeDirectory
-    $abbrevGitDir = $settings -and $settings.DefaultPromptAbbreviateGitDirectory
+    $stringComparison = Get-PathStringComparison
 
     # A UNC path has no drive so it's better to use the ProviderPath e.g. "\\server\share".
     # However for any path with a drive defined, it's better to use the Path property.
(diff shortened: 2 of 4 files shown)


elsassph:
@dahlbyk @rkeithhill now with tests :)

src/Utils.ps1
```diff
@@ -289,10 +290,24 @@ function Get-PromptPath {
     $pathInfo = $ExecutionContext.SessionState.Path.CurrentLocation
     $currentPath = if ($pathInfo.Drive) { $pathInfo.Path } else { $pathInfo.ProviderPath }
 
+    # Look up the git root
+    if ($abbrevGitDir) {
+        $gitPath = $(Get-GitDirectory)
```
rkeithhill: Simplify to just:
```
        $gitPath = Get-GitDirectory
```

src/Utils.ps1
```diff
@@ -289,10 +290,24 @@ function Get-PromptPath {
...
+        # Up one level from `.git`
+        if ($gitPath) { $gitPath = $(Split-Path $gitPath -Parent) }
```
rkeithhill: Simplify to 
```
        if ($gitPath) { $gitPath = Split-Path $gitPath -Parent }
```

src/Utils.ps1
```diff
@@ -289,10 +290,24 @@ function Get-PromptPath {
...
+    }
+
     $stringComparison = Get-PathStringComparison
```
rkeithhill: Let's simplify the following logic a bit by returning early if $currentPath is null or empty e.g.:
```
if (!$currentPath) { return $currentPath }
```
This is semantically equivalent and you can get rid of the `$currentPath` term in both the following `if` and `elseif`.

src/Utils.ps1
```diff
@@ -289,10 +290,24 @@ function Get-PromptPath {
...
 
+    # Abbreviate path by replacing beginning of path with - *iff* the path is under a git repository
+    if ($abbrevGitDir -and $currentPath -and $gitPath -and
+        $currentPath.StartsWith($gitPath, $stringComparison)) {
+        # Up another level to keep repo name in path
+        $removePath = $(Split-Path $gitPath -Parent)
```
rkeithhill: Don't need the `$()` here.

src/Utils.ps1
```diff
@@ -289,10 +290,24 @@ function Get-PromptPath {
...
+        $currentPath = "-" + $currentPath.SubString($removePath.Length)
```
rkeithhill: Hmm, what character should be used to indicate the rooted in the Git repo? I could see some folks wanting to use `$/` as opposed to `-/`. Or maybe `./` or even an emoji.  I wind up replacing `~` with 🏠.  I might be tempted to use 🎪 or 💩for the root of Git repos.  :-)

test/Shared.ps1
```diff
@@ -60,6 +60,30 @@ function GetHomeRelPath([string]$Path) {
     }
 }
 
+function GetGitRelPath([string]$Path) {
+    $gitPath = $(Get-GitDirectory)
```
rkeithhill: Don't need `$()` here or on line 69

elsassph: Pardon my noob-ness :)

src/Utils.ps1
elsassph: Fair enough, `-/` is no convention.

OMG apparently `:/` is the git convention for the repository root! https://stackoverflow.com/a/22049939/27219

Additionally would you suggest adding a new configuration option like `$GitPromptSettings.DefaultPromptGitRootPrefix`?

rkeithhill: Cool.  Let's go with that then.  How about we call it `DefaultPromptGitDirectoryRootPrefix` to try to tie it a bit closer to the other setting?

elsassph: @rkeithhill having tried this `:/` I'm hesitating a bit.

- `:/` is supposed to be the root,
- `:/myproject/src` is incorrect based on this convention
- maybe `myproject:/src` or `[myproject]:/src` would be more correct?

Using a function (passing project name and sub-path) would allow to format that the way people want.

elsassph: image

rkeithhill: That seems reasonable. Of course, whether it's `reponame:\` or `reponame:/` will depend on the platform.  

> Using a function (passing project name and sub-path) would allow to format that the way people want.

What do you mean?  Keep in mind folks can already do this:
```
$GitPromptSettings.DefaultPromptPath.Text = '$(CallMyCustomPromptFunction)'
```

elsassph: Good point about the directory separator direction!
And you're right - one may as well just replace the function.

All fixed, but I didn't add any option for the `:` as it sounds like we are using a correct convention

src/Utils.ps1
```diff
@@ -280,21 +280,36 @@ function Get-PathStringComparison {
...
     # A UNC path has no drive so it's better to use the ProviderPath e.g. "\\server\share".
     # However for any path with a drive defined, it's better to use the Path property.
     # In this case, ProviderPath is "\LocalMachine\My"" whereas Path is "Cert:\LocalMachine\My".
     # The latter is more desirable.
     $pathInfo = $ExecutionContext.SessionState.Path.CurrentLocation
     $currentPath = if ($pathInfo.Drive) { $pathInfo.Path } else { $pathInfo.ProviderPath }
+    if (!$settings -or !$currentPath -or $currentPath.Equals($Home, $stringComparison)) {
+        return $currentPath
+    }
+    $abbrevHomeDir = $settings.DefaultPromptAbbreviateHomeDirectory
```
rkeithhill: Extremely minor nit, sorry but could you add a blank line above this one to separate the previous `if` statement from this set of variable assignments.

src/PoshGitTypes.ps1
```diff
@@ -283,6 +283,7 @@ class PoshGitPromptSettings {
     [PoshGitTextSpan]$DefaultPromptSuffix       = '$(">" * ($nestedPromptLevel + 1)) '
 
     [bool]$DefaultPromptAbbreviateHomeDirectory = $true
+    [bool]$DefaultPromptAbbreviateGitDirectory  = $false
```
rkeithhill: This is looking pretty good to me.  The only thing I wonder about is if users will understand the "interplay" between these two settings i.e. that GitDirectory effectively overrides HomeDirectory?  It's probably OK as is but if were willing to take on another breaking change, I'd almost consider an enum setting here like `$DefaultPromptPathAbbreviation` with values of `None`, `Home`, `BelowHome`, and `GitRoot`.  The deal with `Home` and `BelowHome` is that some folks want the Linux behavior that if you're in your home dir you literally see only `~`.  The current implementation was done to "cushion" the change for Windows users such that in their home dir, they see the full path `C:\Users\keith` but below that, the home dir gets swapped with `~` e.g. `~\GitHub\dahlbyk\posh-git`.

elsassph: Well the git abbreviation goal is to shorten further than home abbreviation. There shouldn't be conflicts, although you can get the git abbreviation outside the home dir (which is desirable as far as I'm concerned).

rkeithhill: OK, let's go with the these two settings then.

rkeithhill:
@elsassph Thanks!

elsassph:
@rkeithhill I stumbled onto something more about the colon:
```
# this is correct
git show :README.md
git show :src/GitUtils.ps1

# this is wrong
git show :/README.md 
git show :/src/GitUtils.ps1
```

So it sounds like it should be adjusted before being worth releasing and maybe this would be a better formatting:
image

src/PoshGitTypes.ps1
dahlbyk: Belated agreement with the two settings.

src/Utils.ps1
```diff
@@ -280,21 +280,37 @@ function Get-PathStringComparison {
...
+    $abbrevGitDir = $settings.DefaultPromptAbbreviateGitDirectory
 
-    # Abbreviate path by replacing beginning of path with ~ *iff* the path is under the user's home dir
-    if ($abbrevHomeDir -and $currentPath -and !$currentPath.Equals($Home, $stringComparison) -and
-        $currentPath.StartsWith($Home, $stringComparison)) {
+    # Look up the git root
+    if ($abbrevGitDir) {
+        $gitPath = Get-GitDirectory
+        # Up one level from `.git`
+        if ($gitPath) { $gitPath = Split-Path $gitPath -Parent }
```
dahlbyk: Two edge cases to consider here:

1. If the current repo is bare, this returns its parent directory.
2. `Get-GitDirectory` inside a worktree currently returns the path to `.git/worktrees/[name]`, which won't match the current path. I can think of two ways to resolve this:
    1. Revise `Get-GitDirectory` to return the contents of `.git/worktrees/[name]/gitdir`. No idea what the downstream effects of this would be.
    2. Check here if `$gitPath` matches `.git/worktrees/`; if it does, set `$gitPath` to the contents of `$gitPath/gitdir`

elsassph: Never used worktrees, but git abbreviation just won't happen in that case.
An issue should be created to track it.

dahlbyk: > Never used worktrees, but git abbreviation just won't happen in that case.

Not a problem. It doesn't break, it just won't work as expected in all cases.

> An issue should be created to track it.

#746

PR #722, opened 2019-11-30
Fix #718 do not create HOME env var

### LINKED ISSUE #718: posh-git messes with the HOME environment variable
This happens with posh-git 1.0.0 beta, and possibly earlier. We have the following lines in the psm1:

```powershell
if (!$Env:HOME) { $Env:HOME = "$Env:HOMEDRIVE$Env:HOMEPATH" }
if (!$Env:HOME) { $Env:HOME = "$Env:USERPROFILE" }
```

This is a problem if the HOME variable wasn't previously set, as it may affect various unrelated programs.

In our particular case we have USERPROFILE pointing to the local C:\Users\name as usual, and HOMEDRIVE set to a mapped network drive. Some programs might default to HOME, then USERPROFILE then HOMEDRIVE, which is now reversed after posh-git has been imported.

posh-git shouldn't mess with important environment variables that affect other things than Git.

rkeithhill:
Windows does not normally define a `HOME` env var so we are always defining that env var.  We should be using PowerShell's built-in variable `$HOME` instead.

Fixes #718

### DIFF
--- a/src/Utils.ps1
+++ b/src/Utils.ps1
@@ -291,11 +291,14 @@ function Get-PromptPath {
 
     $stringComparison = Get-PathStringComparison
 
-    # Abbreviate path by replacing beginning of path with ~ *iff* the path is under the user's home dir
-    if ($abbrevHomeDir -and $currentPath -and !$currentPath.Equals($Home, $stringComparison) -and
-        $currentPath.StartsWith($Home, $stringComparison)) {
+    # $HOME is defined by PowerShell on all platforms. If for some reason it isn't defined, use a fallback.
+    $homePath = if ($HOME) {$HOME} else {[System.Environment]::GetFolderPath([System.Environment+SpecialFolder]::UserProfile)}
 
-        $currentPath = "~" + $currentPath.SubString($Home.Length)
+    # Abbreviate path by replacing beginning of path with ~ *iff* the path is under the user's home dir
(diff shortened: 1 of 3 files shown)


rkeithhill:
@dahlbyk  Ran across this which explains why we did this (and is another complaint requesting we don't do this) #153.  The git.exe wrapper that comes with Git for Windows sets the HOME env var for its process.  Is this sufficient? Or do other Git impls for Windows (msys, mingw, etc) require this?  If so, I think we should make creation of the HOME env var conditional (or maybe we warn as you suggest in the other issue).

cyanite:
Creating the HOME variable is problematic for entirely unrelated programs (such as Azure DevOps command line utilities), where it overrides its normal placement of config files. So enabling posh git might make things stop working. I'd prefer that posh-git never set that variable; it's not needed for git now. See also #718.

rkeithhill:
@cyanite I lean that way myself but I also don't use any of the alternate Git impls on Windows.  OTOH if those are missing `$env:HOME`, seems like the user could define that env var to make Git work.  @dahlbyk ?

rkeithhill:
@cyanite Going to go with not setting this env var at all. I agree that it doesn't seem like something that `posh-git` should be doing.

dahlbyk:
I'm good with removing this. As I noted forever ago in https://github.com/dahlbyk/posh-git/issues/153#issuecomment-53531647, this was in imitation of Git shims that are no longer relevant.

Issue #733, opened 2020-01-20
Tab auto-complete regression


ExE-Boss:
## System Details

- posh-git version: <https://github.com/dahlbyk/posh-git/commit/efed853aa5c5ab235ded5c1a191fac189470eb28>
- PowerShell version: v6.2.3

## Issue Description

[**PR&nbsp;#711**](https://github.com/dahlbyk/posh-git/pull/711) caused&nbsp;a&nbsp;regression that&nbsp;broke tab&nbsp;completion on&nbsp;**PowerShell&nbsp;>=&nbsp;6**,&nbsp;causing:
```
git <command> \t
```

To&nbsp;incorrectly expand&nbsp;to:
```
git <command> <command>
```

Instead&nbsp;of&nbsp;what it&nbsp;expanded&nbsp;to in&nbsp;commit&nbsp;<https://github.com/dahlbyk/posh-git/commit/89dafc1f37d9a44917d5ce6a84e889e54c6f98b9>, which&nbsp;is&nbsp;what I&nbsp;was&nbsp;forced to&nbsp;roll&nbsp;back&nbsp;to.

rkeithhill:
What do you get from the built-in argument completion for `cd ~; gci \w`?

rkeithhill:
Also, do you have a specific repro because I haven't been able to repro what you are seeing?

eXpl0it3r:
Not sure if it's actually an regression, currently the new git command `git switch <branch>` doesn't work with tab-completion though.

rkeithhill:
If you git clone the repo, it has support for git switch.  Has had it for a few months now.  But we haven't published a new beta in almost a year and I'm not sure when the next beta will be published.  I recommend that you git clone this repo and import the module by path.

theaquamarine:
Would you be open to an update to 0.x that adds support for switch and restore?

rkeithhill:
Yes and no.  Obviously adding features to 0.x would help folks but I'd really like to "nudge" folks to 1.0.  That said, 1.0 is still beta so that's hard to do until we release a stable version of v1.  Supporting multiple versions is not ideal when we have such a small set of folks trying to maintain this module.  Whining aside ;-), I would accept a PR to add that functionality to 0.x but who knows when a new version of 0.x would be published.

cdonnellytx:
[...]

PR #737, opened 2020-02-08
Add switch and restore completion to v0

theaquamarine:
As discussed in https://github.com/dahlbyk/posh-git/issues/733#issuecomment-583513199, this would allow v0 to support "new" usage for git at least until v1 updates/releases.

Uses the code from #702 for `restore` and `switch` support, plus a couple of lines from `Get-GitDirectoryTests.ps1` to fix some tests being broken by git asking for user details.

### DIFF
--- a/src/GitTabExpansion.ps1
+++ b/src/GitTabExpansion.ps1
@@ -165,6 +165,10 @@ function script:gitCheckoutFiles($GitStatus, $filter) {
     gitFiles $filter (@($GitStatus.Working.Unmerged) + @($GitStatus.Working.Modified) + @($GitStatus.Working.Deleted))
 }
 
+function script:gitDeleted($GitStatus, $filter) {
+    gitFiles $filter $GitStatus.Working.Deleted
+}
+
 function script:gitDiffFiles($GitStatus, $filter, $staged) {
     if ($staged) {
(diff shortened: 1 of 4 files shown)


src/GitTabExpansion.ps1
```diff
@@ -218,10 +227,16 @@ function script:expandShortParams($cmd, $filter) {
...
 
 function script:expandParamValues($cmd, $param, $filter) {
-    $gitParamValues[$cmd][$param] -split ' ' |
-        Where-Object { $_ -like "$filter*" } |
-        Sort-Object |
-        ForEach-Object { -join ("--", $param, "=", $_) }
+    $paramValues = $gitParamValues[$cmd][$param]
+
+    $completions = if ($paramValues -is [scriptblock]) {
+        & $paramValues $filter | Where-Object { $_ -like "$filter*" }
```
dahlbyk: Let's bring in fixes from #743

rkeithhill:
@theaquamarine Given that beta4 is out now with support for these two new git commands, can we close this?  Or if you'd like to get this merged, could you apply the fixes from #743 and see if that gets the tests to pass?  Thanks.

rkeithhill:
I'd like to close this now that v1 has shipped.  Any objections?

PR #753, opened 2020-03-30
Fix typo in README: GitPromtpValues

douglaswth:


rkeithhill:
Good catch.  Thanks for the PR to fix it!

Issue #755, opened 2020-04-02
Some repositories fail to get status when using GitStatusCache


tmercswims:
### System Details

- posh-git version/path: 1.0.0 beta4 ~\Documents\PowerShell\Modules\posh-git\1.0.0
- PowerShell version: 7.0.0
- git version 2.26.0.windows.1 hub version 2.14.2
- OS: Microsoft Windows NT 10.0.18363.0

### Issue Description

I have [git-status-cache-posh-client](https://github.com/cmarcusreid/git-status-cache-posh-client) installed and working. I have posh-git configured to use it (`$GitPromptSettings.EnableFileStatusFromCache = $true`).

I also have posh-git integrated into my prompt. I use it to print status information to a [PowerLine](https://github.com/Jaykul/PowerLine) block in my PowerShell profile like this:
```powershell
# ...
{ New-PromptText { & { Write-VcsStatus } 6>&1 } -BackgroundColor Yellow -ForegroundColor Black },
# ...
```
Side note: I'm unsure if there is a better way to do this, as this was the only way I could see to get the repository information in the "prompt format" while also taking advantage of the caching. Either way, the end result, most of the time, looks like this (name blacked because it's a private org repo, sorry 😬):
![image](https://user-images.githubusercontent.com/1624639/78295069-55406000-74e0-11ea-84ce-bd7571725d59.png)

However, if I clone or create a new repository on my system, then it looks like this:
![image](https://user-images.githubusercontent.com/1624639/78295863-a0a73e00-74e1-11ea-9e68-b9a42416e147.png)

If I do `Get-GitStatus` then I get a little bit more detail:
```powershell
$ Get-GitStatus
InvalidOperation: C:\Users\thomas.mercurio\Documents\PowerShell\Modules\posh-git\1.0.0\GitUtils.ps1:179
Line |
 179 |              $hash[$path] = 1
     |              ~~~~~~~~~~~~~~~~
     | Index operation failed; the array index evaluated to null.

InvalidOperation: C:\Users\thomas.mercurio\Documents\PowerShell\Modules\posh-git\1.0.0\GitUtils.ps1:179
Line |
 179 |              $hash[$path] = 1
     |              ~~~~~~~~~~~~~~~~
     | Index operation failed; the array index evaluated to null.

InvalidOperation: C:\Users\thomas.mercurio\Documents\PowerShell\Modules\posh-git\1.0.0\GitUtils.ps1:179
Line |
 179 |              $hash[$path] = 1
[...]

PR #759, opened 2020-04-07
Put 'CACHE ERROR' in prompt when cache returns failure.

cmarcusreid:
Follow-up from https://github.com/dahlbyk/posh-git/issues/755

image

This diff is a bit unfortunate; the only new bit is the leading if block. The rest was indented in the new else block.

### DIFF
--- a/src/GitUtils.ps1
+++ b/src/GitUtils.ps1
@@ -252,32 +252,42 @@ function Get-GitStatus {
             if ($settings.EnableFileStatusFromCache) {
                 dbg 'Getting status from cache' $sw
                 $cacheResponse = Get-GitStatusFromCache
-                dbg 'Parsing status' $sw
 
-                $indexAdded.AddRange($castStringSeq.Invoke($null, (,@($cacheResponse.IndexAdded))))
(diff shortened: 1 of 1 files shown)


rkeithhill:
Fortunately, you can diff this in VSCode with the GitHub Pull Requests extension.  And because VSCode diffs allow you to ignore whitespace, this diff simplifies to:

image

cmarcusreid:
TIL! Thanks!

rkeithhill:
I really wish GitHub would offer the option to ignore whitespace changes.  This is one area where Bitbucket has a leg up on GH.

rkeithhill:
Thanks for the PR!  I've asked the other Keith to review it before merging it.

dahlbyk:
Thanks for the fix, @cmarcusreid!

> I really wish GitHub would offer the option to ignore whitespace changes in diffs. This is one area where Bitbucket has a leg up on GH.

You have been able to add `?w=1` for years, but it has been in the UI for a while, too.

Screenshot_20200407-212532

rkeithhill:
Thanks for pointing that out! That has been bugging me for a while. I was looking in the wrong place - in the `...` menu to the right of the `Viewed` button.

Issue #657, opened 2019-02-13
Documentation and improved Get-GitStatus for powershell script integration


arberg:
### System Details

- posh-git version/path: 0.7.3.1  C:\Tools\poshgit\dahlbyk-posh-git-9bda399\src
- PowerShell version: 5.1.17134.407
- git version 2.20.1.windows.1
- OS: Microsoft Windows NT 10.0.17134.0

### Issue Description

I was developing our a powershell script for some simple custom management of our git-submobules, with nice overview status of the state of each submodule, and got tired of parsing git status. In came posh-git in my mind, and its absolute awesome for getting git info using Get-GitStatus. However I noticed it could easily be made even slightly more awesome. The Readme lacks some info for using posh-git as powershell functions and not just prompt-management. The Get-GitStatus could use a boolean for isOnBranch rather that manually having to parse if branchName is on form "(xfz)". The Get-GitStatus could also use some documentation for the different attributes. Like what does 'HasWorkning' mean? 

I might be inspired to invest a few hours adding this. Would you be interested in such a pull-request?

rkeithhill:
Since `Get-GitStatus` runs every time the PowerShell prompt is processed we try to make sure perf is as good as it can be.  Processing the branch name text to determine if we're "on a branch" isn't something we need for processing the prompt string.  So it's a bit hard to justify adding this property.  Regarding documentation, yes this command's comment-help could use a lot of, well, help.  :-)  We'd gladly accept a PR to improve the help.  BTW 'HasWorking' indicates whether there are workspace changes and we should display the `EFGH W` section of the prompt - `[{HEAD-name} S +A ~B -C !D | +E ~F -G !H W]`.

elsassph:
Could `Get-GitStatus` accept extra parameters?

I don't like seeing the git status in the prompt, so I set `EnableFileStatus=$false`, however I would really like to be able to call like `Get-GitStatus` to obtain the repository status.

rkeithhill:
What if we use the existing `-Force` parameter and have that override `EnableFileStatus` as well as `EnablePromptStatus`?  Another option you could use for now is to not set `EnableFileStatus` to false and define your own prompt function in your profile.  If you import posh-git after your prompt function def, posh-git will not override your prompt.  Then `Get-GitStatus` will return the info you want and then you can display it however you want.

elsassph:
If `-Force` would override `EnableFileStatus` that would be most suitable.

[...]

PR #760, opened 2020-04-11
Update help doc comments on Get-GitStatusFix #657

rkeithhill:


rkeithhill:
After adding all the other settings that impact Get-GitStatus, I wonder if `-Force` should also override `EnableStashStatus` as well?

dahlbyk:
> After adding all the other settings that impact `Get-GitStatus`, I wonder if `-Force` should also override `EnableStashStatus` as well?

IMO no, since posh-git defaults to not including stash status. Maybe `Get-GitStatus -IncludeStashStatus`?

rkeithhill:
Another option would be to pass in a GitPromptSettings object via a `-GitPromptSettings` parameter and use the global GitPromptSettings as the default value for this parameter when it's not provided e.g.:
```
$s = New-GitPromptSettings
$s.EnableFileStatusFromCache = $false
$s.EnableStashStatus = $true
$s.UntrackedFilesMode = 'All'
Get-GitStatus -GitPromptSettings $s
```
That would give a user of `Get-GitStatus` full control but yeah, the usage is less than ideal.  And adding this would conflict with `-Force`.  I suppose it could the `-GitPromptSettings` parameter could be added in another parameter set - exclusive of `-Force`.  Thoughts?

PR #762, opened 2020-04-12
Be sure to honor EnableStashStatus when using cache

### LINKED ISSUE #761: When using git-status-cache stash settings are not respected
### System Details

- posh-git version/path: latest from master (a160c9a)
- PowerShell version: 7
- Git version: 2.26.0.windows.1
- Operating system name and version: Microsoft Windows [Version 10.0.18363.752]
- posh-git version/path: 1.0.0.0 beta4x C:\p\WindowsPowerShell\Modules\posh-git\src
- PowerShell version: 7.0.0

### Issue Description

Even when `$global:GitPromptSettings.EnableStashStatus` is set to false, if git-status-cache is installed, `StashCount` is reported on `Get-GitStatus`.

rkeithhill:
Fix #761

### DIFF
--- a/src/GitUtils.ps1
+++ b/src/GitUtils.ps1
@@ -285,8 +285,13 @@ function Get-GitStatus {
                     $aheadBy = $cacheResponse.AheadBy
                     $behindBy = $cacheResponse.BehindBy
 
-                    if ($cacheResponse.Stashes) { $stashCount = $cacheResponse.Stashes.Length }
-                    if ($cacheResponse.State) { $branch += "|" + $cacheResponse.State }
+                    if ($settings.EnableStashStatus -and $cacheResponse.Stashes) {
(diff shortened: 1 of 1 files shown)


src/GitUtils.ps1
```diff
@@ -285,8 +285,13 @@ function Get-GitStatus {
                     $aheadBy = $cacheResponse.AheadBy
                     $behindBy = $cacheResponse.BehindBy
 
-                    if ($cacheResponse.Stashes) { $stashCount = $cacheResponse.Stashes.Length }
-                    if ($cacheResponse.State) { $branch += "|" + $cacheResponse.State }
+                    if ($settings.EnableStashStatus -and $cacheResponse.Stashes) {
```
dahlbyk, line 288: If @cmarcusreid gets bored, the API could conceivably receive request options like this to prevent the cache from doing background work the user doesn't care.

Issue #758, opened 2020-04-07
Prompt breaks if you execute a script with strict mode


gsscoder:
First of all thank you for this very useful and well done project!

If I execute a script with this strict mode enabled:
```powershell
Set-StrictMode -Version Latest
```
**posh-git** prompt breaks:
```powershell
$ [PoshGitVcsPrompt error: The property 'Length' cannot be found on this object. Verify that the property exists.]> 
```

It seems that **strict mode** is extended to loaded modules. Do you think it's better to post this issue in *PowerShell* repo?

Regards,
Giacomo S. S.

dahlbyk:
We've fixes strict mode issues before…seems like something we should just fix. 


Maybe we should always run tests in strict mode?

rkeithhill:
We definitely should fix the strict mode error.  Setting strict mode to latest caused 4 failures but not the missing Length error:
```
   [-] Tab completes when there is one alias of a given name 74ms (72ms|2ms)
    PropertyNotFoundException: The property 'Count' cannot be found on this object. Verify that the property exists.
    at <ScriptBlock>, C:\Users\Keith\GitHub\dahlbyk\posh-git\test\TabExpansion.Tests.ps1:270

 Context Get-PromptConnectionInfo
   [-] Returns null if Env:SSH_CONNECTION is not set 8ms (6ms|2ms)
    RuntimeException: The variable '$ssh_connection' cannot be retrieved because it has not been set.
    at <ScriptBlock>, C:\Users\Keith\GitHub\dahlbyk\posh-git\test\Utils.Tests.ps1:116
   [-] Returns null if Env:SSH_CONNECTION is empty 4ms (2ms|2ms)
    RuntimeException: The variable '$ssh_connection' cannot be retrieved because it has not been set.
    at <ScriptBlock>, C:\Users\Keith\GitHub\dahlbyk\posh-git\test\Utils.Tests.ps1:116
   [-] Returns "[username@hostname]: " if Env:SSH_CONNECTION is set 5ms (3ms|2ms)
    RuntimeException: The variable '$ssh_connection' cannot be retrieved because it has not been set.
    at <ScriptBlock>, C:\Users\Keith\GitHub\dahlbyk\posh-git\test\Utils.Tests.ps1:116
```
FYI, the line numbers are for my Pester v5 port branch.

rkeithhill:
Nevermind. Need to set strict mode in the module and yeah, that shows plenty (37) of potential issues.

[...]

PR #783, opened 2020-07-26
Enables strict mode testing of module, fix git restore --source= tab bug

rkeithhill:
Also fixes bug in git flow feature tab completion.

### DIFF
--- a/src/posh-git.psm1
+++ b/src/posh-git.psm1
@@ -1,5 +1,10 @@
 param([bool]$ForcePoshGitPrompt, [bool]$UseLegacyTabExpansion)
 
+if (Test-Path Env:\POSHGIT_ENABLE_STRICTMODE) {
+    # Set strict mode to latest to help catch scripting errors in the module. This is done by the Pester tests.
+    Set-StrictMode -Version Latest
+}
+
 . $PSScriptRoot\CheckRequirements.ps1 > $null
 
 . $PSScriptRoot\ConsoleMode.ps1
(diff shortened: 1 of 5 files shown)


src/posh-git.psm1
```diff
@@ -1,5 +1,10 @@
 param([bool]$ForcePoshGitPrompt, [bool]$UseLegacyTabExpansion)
 
+if (Test-Path Env:\POSHGIT_ENABLE_STRICTMODE) {
+    # Set strict mode to latest to help catch scripting errors in the module. This is done by the Pester tests.
+    Set-StrictMode -Version Latest
+}
```
dahlbyk, line 6: I assume `Set-StrictMode` is best done here rather than in `test/Shared.ps1` to scope strictness to the module, rather than all tests too?

Hi, hello, yes it's me I still exist. Barely. 👋

rkeithhill: Yeah, I found when I set it in Shared.ps1 that A) it complained about some of our "lose" scripting in the tests and B) it wasn't finding issues in the module.  That goes counter to the claim in #758 but it makes sense that enabling set-strictmode in the module just for testing is the way to go - IMO.   

BTW I don't think we want to leave strict mode on for ship because it will gen errors with user injected code like this:
```
$global:GitPromptSettings.DefaultPromptSuffix.Text = ' $((Get-History -Count 1).id + 1)$(">" * ($nestedPromptLevel + 1)) '
```
For example, when you first start PowerShell there is no history so `Get-HIstory` returns `$null` and strict-mode doesn't like the null de-ref to get the `id`. But for my prompt string it works out - `$null + 1 ` is `1`.

Issue #708, opened 2019-10-20
Prompt error in remote PSSession


fullenw1:
### System Details

- posh-git version/path: 0.7.3  C:\Program Files\WindowsPowerShell\Modules\posh-git\0.7.3
- PowerShell version: 5      1      14409  1018 
- Git version: 1.9.5.msysgit.0
- Operating system name and version: Win32NT             6.3.9600.0 Microsoft Windows NT 6.3.9600.0

### Issue Description

When I am in a remote PSSession, the prompt displays an error message:

`C:\Repo [Error: Exception setting "OutputEncoding": "The handle is invalid.`
`"][ComputerName]: >`

When I am in a local PowerShell session on the same computer, the prompt works fine.

`C:\Repo [master ≡]>`

rkeithhill:
Can you repro that and then execute `$error[0] | Format-List * -Force` and paste the output here.

fullenw1:
PSMessageDetails      : 
Exception             : System.Management.Automation.SetValueInvocationException: Exception setting "OutputEncoding": "The handle is invalid.
                        " ---> System.IO.IOException: The handle is invalid.
                        
                           at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
                           at System.Console.set_OutputEncoding(Encoding value)
                           at CallSite.Target(Closure , CallSite , Type , Object )
                           --- End of inner exception stack trace ---
                           at System.Management.Automation.ExceptionHandlingOps.CheckActionPreference(FunctionContext funcContext, Exception exception)
                           at System.Management.Automation.Interpreter.ActionCallInstruction`2.Run(InterpretedFrame frame)
                           at System.Management.Automation.Interpreter.EnterTryCatchFinallyInstruction.Run(InterpretedFrame frame)
                           at System.Management.Automation.Interpreter.EnterTryCatchFinallyInstruction.Run(InterpretedFrame frame)
TargetObject          : 
CategoryInfo          : NotSpecified: (:) [], SetValueInvocationException
FullyQualifiedErrorId : ExceptionWhenSetting
ErrorDetails          : 
InvocationInfo        : System.Management.Automation.InvocationInfo
ScriptStackTrace      : at Invoke-Utf8ConsoleCommand, C:\Program Files\WindowsPowerShell\Modules\posh-git\0.7.3\Utils.ps1: line 42
[...]

PR #796, opened 2020-09-30
Fixed non-ascii git branch name garbled.

LittleboyHarry:
By 2020, The utf-8 encoding beats the market for any other specified non-single byte encodings for internationalization. In fact, most of Asian developer would prefer utf-8 as the first-class citizens while remaining specified old encoding like gbk, big5, jis as default page code only for windows system compatibly.

#397 @dahlbyk you introduced a bug that it wont't change to utf-8 for non-single byte while you actually want to do it!

After fixed, It will avoid changing output encoding which causing error for single-byte charset users. (but it will return get non utf8 encoding string which make no problem in fact )  #708

### DIFF
--- a/src/Utils.ps1
+++ b/src/Utils.ps1
@@ -32,16 +32,16 @@ function Invoke-Utf8ConsoleCommand([ScriptBlock]$cmd) {
         # A native executable that writes to stderr AND has its stderr redirected will generate non-terminating
         # error records if the user has set $ErrorActionPreference to Stop. Override that value in this scope.
         $ErrorActionPreference = 'Continue'
-        if ($currentEncoding.IsSingleByte) {
-            try { [Console]::OutputEncoding = [Text.Encoding]::UTF8 } catch [System.IO.IOException] {}
+        if (!$currentEncoding.IsSingleByte) {
+            [Console]::OutputEncoding = [Text.Encoding]::UTF8
+            & $cmd
+            [Console]::OutputEncoding = $currentEncoding
         }
-        & $cmd
+        else {
+            & $cmd
+        }
     }
     finally {
-        if ($currentEncoding.IsSingleByte) {
-            try { [Console]::OutputEncoding = $currentEncoding } catch [System.IO.IOException] {}
-        }
-
         # Clear out stderr output that was added to the $Error collection, putting those errors in a module variable
         if ($global:Error.Count -gt $errorCount) {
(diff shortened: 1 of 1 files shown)


LittleboyHarry:
@fullenw1 will you get IOException using new version in your machine?

src/Utils.ps1
```diff
@@ -32,16 +32,16 @@ function Invoke-Utf8ConsoleCommand([ScriptBlock]$cmd) {
...
-        if ($currentEncoding.IsSingleByte) {
-            try { [Console]::OutputEncoding = [Text.Encoding]::UTF8 } catch [System.IO.IOException] {}
+        try { 
+            [Console]::OutputEncoding = [Text.Encoding]::UTF8
+            & $cmd
+            [Console]::OutputEncoding = $currentEncoding
+        }
+        catch [System.IO.IOException] {
+            & $cmd
         }
```
dahlbyk, line 42: Seems like wrapping `& $cmd` in the `try` could result in reexecuting `$cmd` if it happens to `throw IOException`.

Is there a reason not to keep the previous pattern, just drop the `IsSingleByte` checks?
```suggestion
        try { [Console]::OutputEncoding = [Text.Encoding]::UTF8 } catch [System.IO.IOException] {}
        & $cmd
```

LittleboyHarry: > Is there a reason not to keep the previous pattern, just drop the IsSingleByte checks?

GBK and other Asian encoding aren't single byte, it will cause returing garbled until I switch [Console]::OutputEncoding to utf-8.

BTW, I use powershell 7.0.3, on windows 10 x64 2004, machine default encoding is gb2312 not utf8.

LittleboyHarry:
How about the new version?

src/Utils.ps1
dahlbyk: Sorry, I meant to drop the `if ($currentEncoding.IsSingleByte)` but otherwise keep the previous structure, i.e.

```pwsh
        try { [Console]::OutputEncoding = [Text.Encoding]::UTF8 } catch [System.IO.IOException] {}
        & $cmd
    }
    finally {
        try { [Console]::OutputEncoding = $currentEncoding } catch [System.IO.IOException] {}
```

Specifically `finally` seems preferable to running `$cmd` in a `catch`.

dahlbyk:
If this still solves the issue you observe, I'm good with it. Thanks!

PR #797, opened 2020-10-03
If abbreviating git dir, display correct path on a bare repo

RexTremendae:


src/Utils.ps1
```diff
@@ -295,8 +295,8 @@ function Get-PromptPath {
     # Look up the git root
     if ($abbrevGitDir) {
         $gitPath = Get-GitDirectory
-        # Up one level from `.git`
-        if ($gitPath) { $gitPath = Split-Path $gitPath -Parent }
+        # Up one level from `.git` if inside git directory
+        if ($gitPath -and $gitPath.EndsWith(".git")) { $gitPath = Split-Path $gitPath -Parent }
```
dahlbyk, line 299: I suppose we should clarify what's expected here. As far as I can tell from my testing, this doesn't actually change behavior - I get this same display before and after: 
image

I could argue that it should show `posh-git.git: [BARE:master]>` instead of `Temp:posh-git.git`.

dahlbyk (commented):
Sorry for the delayed review. Appreciate the contribution!

Thoughts on my comment?

src/Utils.ps1
RexTremendae: Yes, that's the idea of the PR. I get a different behavior when I compare current master (just updated today) with this PR (which now is rebased onto master):

image

dahlbyk: Ah, the difference is my bare repo has a `.git` suffix but yours does not:
image

Maybe we should check `(Split-Path $gitDir -Leaf) -eq '.git'` instead of `EndsWith()`?

RexTremendae: Right you are...

Thinking a little bit more about it, I think my solution is a little bit over-simplified. Maybe the correct solution would be to base the logic on the same information that casues the "BARE:" label to appear? However, I cannot see that information being stored anywhere other than in the branch specifier after the function `Get-GitBranch` has completed.

I really don't have the full overview of how all the functions work together in the code base; maybe it is possible to receive this information somehow?

dahlbyk: > Thinking a little bit more about it, I think my solution is a little bit over-simplified. Maybe the correct solution would be to base the logic on the same information that casues the "BARE:" label to appear? However, I cannot see that information being stored anywhere other than in the branch specifier after the function `Get-GitBranch` has completed.

You're right, we only store `BARE`/`GIT_DIR` in the branch name at the moment. We could definitely tack those onto `$GitStatus` with a bit of refactoring, but we'd also have to rearrange quite a bit to make `$GitStatus` available at the point where we're calling `Get-PromptPath`.

Let's just check for a `.git` exact match for now. If we get feedback that it's insufficient, we can revisit.

> I really don't have the full overview of how all the functions work together in the code base; maybe it is possible to receive this information somehow?

I don't know that we have an overview written up anywhere, but that would definitely be good to have. In short:

1. Unless it has already been customized, we replace `prompt` with `$GitPromptScriptBlock` https://github.com/dahlbyk/posh-git/blob/b0cbffb7549c485514518dc0872708996835c6c8/src/posh-git.psm1#L31
2. `$GitPromptScriptBlock` builds the actual prompt mostly by expanding strings defined in `$global:GitPromptSettings` https://github.com/dahlbyk/posh-git/blob/b0cbffb7549c485514518dc0872708996835c6c8/src/PoshGitTypes.ps1#L225
    - This is how `Get-PromptPath` is called https://github.com/dahlbyk/posh-git/blob/b0cbffb7549c485514518dc0872708996835c6c8/src/PoshGitTypes.ps1#L280 https://github.com/dahlbyk/posh-git/blob/5ce5422e5ab8d6998cf5e51f472b3f7c05e236a6/src/posh-git.psm1#L61
3. Git status comes into the prompt from `Write-GitStatus` via `Write-VcsStatus` (added to support posh-hg long ago) https://github.com/dahlbyk/posh-git/blob/b0cbffb7549c485514518dc0872708996835c6c8/src/GitPrompt.ps1#L202

RexTremendae: Great, thanks for the walk-through!

Issue #746, opened 2020-03-16
Abbreviate Git Directory Edge Cases


dahlbyk:
Two edge cases not considered by original implementation (#720):

1. If the current repo is bare, returns its parent directory.
2. `Get-GitDirectory` inside a worktree currently returns the path to `.git/worktrees/[name]`, which won't match the current path. I can think of two ways to resolve this:
    1. Revise `Get-GitDirectory` to return the contents of `.git/worktrees/[name]/gitdir`. No idea what the downstream effects of this would be.
    2. Check here if `$gitPath` matches `.git/worktrees/`; if it does, set `$gitPath` to the contents of `$gitPath/gitdir`

RexTremendae:
Submitted a PR (#797) to address nr 1 above.

I also found another thing which I don't really understand, Utils.ps1 line 288:
`if (!$settings -or !$currentPath -or $currentPath.Equals($Home, $stringComparison))`

This means that when in the home directory root, the path won't be abbreviated, this will only when I go down one or more levels. Isn't a more standard behavior to also abbreviate the home dir to only '~'?

rkeithhill:
> Isn't a more standard behavior to also abbreviate the home dir to only '~'?

On Linux yes but after using it on Windows, it seemed a bit too much of an abbreviation for the typical Windows user.  We could qualify that last condition with a test for the OS so that we'd only do this on Windows.  Alternatively, we could add a new setting for this that defaults differently for Windows.

RexTremendae:
Alright, then I vote for a setting so that people who use both Windows and Linux can feel welcome at ~ :D

rkeithhill:
After thinking about this some more, I'm really tempted to simplify and make Windows behave like Linux and just show `~` when in the home dir.

dahlbyk:
Is it reasonable to default to whatever PowerShell defaults to without posh-git?

rkeithhill:
Well, PowerShell does not abbreviate paths displayed in the prompt.  Maybe we should not enable this option `DefaultPromptAbbreviateHomeDirectory` by default.  And if the user does enable it, make it behave like a typical Linux prompt - always display `~` in place of the home path even when directly in the home path.

dahlbyk:
> And if the user does enable it, make it behave like a typical Linux prompt - always display `~` in place of the home path even when directly in the home path.

This matches Git Bash behavior, too.

rkeithhill:
OK, I'll submit a PR to make this change later tonight!

PR #801, opened 2020-10-20
Only default home dir abbreviation to $true on Linux/macOS

rkeithhill:
If enabled, always substitute ~ for the home dir path on all OSs.

Address issue brought up in #746

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,13 @@
   ([PR #779](https://github.com/dahlbyk/posh-git/pull/779))
   Thanks @csc027
 
+### Changed
+
+- `$GitPromptSettings.DefaultPromptAbbreviateHomeDirectory` defaults to `$false` on Windows. When enabled, `~` is
+   always substituted for the user's home path. Before this change, when directly in your home dir, this setting would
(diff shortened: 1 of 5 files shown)


rkeithhill:
@dahlbyk Well, this looks a little odd now:
image

dahlbyk:
> @dahlbyk Well, this looks a little odd now:
> 
> image
> 
> Can we change that separator character or maybe just get rid of it?

The `~` was introduced in 1ed6d4d3c26059c5b0abbd33708dde2be70ad8f2 due to a Console2 bug that's either long fixed or likely irrelevant—and it's customizable. Let's change to a `-` by default, or maybe `>` to roughly mirror what's visible in the prompt?

dahlbyk (approved):
Leaving open in case you want to tack on the `~` change. Otherwise feel free to merge.

rkeithhill:
Here's what a `>` looks like:

image

image

vs a `-`

image

image

I dunno.  Having `>` in the title bar seems a bit weird to me while `-` seems more "titlebar like".

dahlbyk:
> Having `>` in the title bar seems a bit weird to me while `-` seems more "titlebar like".

Concur. Let's go with `-`.

PR #802, opened 2020-10-21
Autocomplete user-defined pretty formats

### LINKED ISSUE #786: Support completion for user-defined formats
### System Details

- posh-git version/path: 1.0.0 beta4 ~\Documents\WindowsPowerShell\Modules\posh-git\1.0.0
- PowerShell version: 5.1.19041.1
- git version 2.28.0.windows.1
- OS: Microsoft Windows NT 10.0.19041.0

### Issue Description

Git offers the possibility to [define your own formats](https://git-scm.com/docs/git-config#Documentation/git-config.txt-prettyltnamegt) to use with `--format` flag for `git log`. I defined a custom one:
```gitconfig
[pretty]
        birdseye = %C(auto)%h%d% s %C(bold blue)<%aN> %C(green)(%cr)%Creset
```
but it doesn't show up in the list of completions when I type `--format=`:

image

rdnlsmith:
Enumerates any `pretty.*` keys in the user's Git config and includes
them alongside built-in options like `oneline`, `short`, `full`, etc.
for the `--pretty=` and `--format=` parameters of `git log` and
`git show`.

The key enumeration is done generically via function `gitConfigKeys`,
which in theory could be used for enumerating other config values with a
similar structure.

Fixes #786.

### DIFF
--- a/src/GitParamTabExpansion.ps1
+++ b/src/GitParamTabExpansion.ps1
@@ -150,8 +150,24 @@ $gitParamValues = @{
     log = @{
         decorate = 'short full no'
         'no-walk' = 'sorted unsorted'
-        pretty = 'oneline short medium full fuller email raw'
-        format = 'oneline short medium full fuller email raw'
+        pretty = {
+            param($ref)
+            $completions = @()
+            gitConfigKeys $ref 'pretty' | ForEach-Object { $completions += $_ }
+            'oneline short medium full fuller email raw' -split ' ' |
+                Where-Object { $_ -like "$filter*" } |
+                ForEach-Object { $completions += $_ }
+            return $completions | Sort-Object
+        }
+        format = {
+            param($ref)
+            $completions = @()
+            gitConfigKeys $ref 'pretty' | ForEach-Object { $completions += $_ }
+            'oneline short medium full fuller email raw' -split ' ' |
+                Where-Object { $_ -like "$filter*" } |
+                ForEach-Object { $completions += $_ }
+            return $completions | Sort-Object
+        }
         encoding = 'UTF-8'
         date = 'relative local default iso rfc short raw'
     }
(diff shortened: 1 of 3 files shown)


src/GitParamTabExpansion.ps1
```diff
@@ -150,8 +150,24 @@ $gitParamValues = @{
     log = @{
         decorate = 'short full no'
         'no-walk' = 'sorted unsorted'
-        pretty = 'oneline short medium full fuller email raw'
-        format = 'oneline short medium full fuller email raw'
+        pretty = {
+            param($ref)
```
dahlbyk, line 154: Rename variable to `$format`?

src/GitParamTabExpansion.ps1
```diff
@@ -150,8 +150,24 @@ $gitParamValues = @{
...
+            $completions = @()
+            gitConfigKeys $ref 'pretty' | ForEach-Object { $completions += $_ }
+            'oneline short medium full fuller email raw' -split ' ' |
```
dahlbyk, line 157: Would be nice to reduce some of this duplication. Two ideas:

1. Extract a `$defaultPrettyFormats` variable (with new options `reference` and `mboxrd`; [docs](https://git-scm.com/docs/pretty-formats)), probably pre-`split`
2. Update `gitConfigKeys` to accept an optional third parameter with additional values to include (i.e. `gitConfigKeys $format 'pretty' $defaultPrettyFormats`), so we can filter/sort all at once

And IMO the `$section` parameter looks better first, since that specifies which keys we're about to filter. We'd end up with:

```pwsh
pretty = {
    param($format)
    gitConfigKeys 'pretty' $format $defaultPrettyFormats
}
```

dahlbyk (commented):
Thanks for the contribution! I have a few suggestions - thoughts?

It would be nice to have tests to verify this works as expected with and without custom `pretty.*` config. For alias setup/teardown, see https://github.com/dahlbyk/posh-git/blob/d94feed10170b737693136367f1056e16964d7cc/test/TabExpansion.Tests.ps1#L198-L207

rdnlsmith:
Updated! I totally agree with the code changes—the way I had it was a case of me trying a little too hard to emulate existing nearby code.

Hopefully the tests I wrote are appropriate!

dahlbyk (approved):
Nailed it. Nice work, and thanks for the contribution!

Issue #804, opened 2020-10-23
Incorrect git checkout completion for branches checked out in other worktrees


SpecLad:
### System Details

- posh-git version/path: 1.0.0 beta4 ~\Documents\PowerShell\Modules\posh-git\1.0.0
- PowerShell version: 7.0.3
- git version 2.29.0.windows.1
- OS: Microsoft Windows NT 10.0.19041.0

### Issue Description

Do this:

```
> git init repo
> cd repo
> git commit --allow-empty -m "Test"
> git worktree add -b test ../wt
```

Now type `git checkout ` and start pressing Tab. One of the completions will be `'+ test'`, which is incorrect. It should either be `test`, or it should not be suggested at all, because you can't check out a branch that's already checked out in another worktree.

dahlbyk:
Great bug report! I'm curious what Git Bash completion does for this case. Even though checking it out is an error, it seems preferable to include it in the list. For example, this is valid: `git checkout test -b branch-from-test`.

SpecLad:
> I'm curious what Git Bash completion does for this case.

It includes the `test` branch in the list of completions.

dahlbyk:
Marking this as up-for-grabs, if anyone wants to take a crack at it. Pretty sure we just need to update a regex here:

https://github.com/dahlbyk/posh-git/blob/e4231b89a13485cbf0c5374a749141be8448d3dc/src/GitTabExpansion.ps1#L138

wallybh:
Hi guys. I can do this. I'll open a PR soon.

dahlbyk:
This should be fixed in `master` and will be released with beta5 🔜

PR #805, opened 2020-10-24
Fix tab completion of branches in different worktrees

wallybh:
When a branch is created in another worktree, the tab complection show
this branch incorrectly.

Added the plus sign character to the regex on the 'gitBranches' function.

#804


rkeithhill (approved):

dahlbyk:
Thanks!

Issue #771, opened 2020-05-20
Prompt only renders in ansi 


tvfischer:
### System Details

- posh-git version/path: 1.0.0 beta4 D:\tvfischer\PowerShell\Modules\posh-git\1.0.0
- PowerShell version: 7.0.1
- git version 2.21.0
- OS: Microsoft Windows NT 10.0.18363.0

### Issue Description

I am experiencing a problem with with the rendering of the prompt. It is constantly displaying what looks like the ansi version of the prompt information. See picture below.
![image](https://user-images.githubusercontent.com/439019/82462669-c6999980-9ab3-11ea-9df2-506e16d14bc4.png)

Terminal code page is utf8:
```
[console]::inputencoding

Preamble          :
BodyName          : utf-8
EncodingName      : Unicode (UTF-8)
HeaderName        : utf-8
WebName           : utf-8
WindowsCodePage   : 1200
IsBrowserDisplay  : True
IsBrowserSave     : True
IsMailNewsDisplay : True
IsMailNewsSave    : True
IsSingleByte      : False
EncoderFallback   : System.Text.EncoderReplacementFallback
DecoderFallback   : System.Text.DecoderReplacementFallback
IsReadOnly        : True
CodePage          : 65001
```

If i try version 0.7.3 it renders a bit better but doesn't quite work and fails with oh-my-posh.

GoatHead:
Same here. the prompt is not showing Hangul.
![image](https://user-images.githubusercontent.com/39069051/82519008-9c031b80-9b5b-11ea-8ac9-ddcc580ef45f.png)

rkeithhill:
I have a few questions.  What is PS running in?  PowerShell console, Windows Terminal, ConEmu, Cmder, etc?  What do you get for `$host.ui.SupportsVirtualTerminal`?  Also, the code page I see is 437.  Are you running on a non-English system?  Finally, what do you get if you start pwsh like so `pwsh -noprofile` then execute, `ipmo posh-git` and the cd into a Git repo?

tvfischer:
For my part, I am running it in the default pwsh console under W10 w/cmdline `"C:\Program Files\PowerShell\7\pwsh.exe" -WorkingDirectory ~`
I also ran it in the recent final release of windows terminal. Both show the problem.

- pwsh window: `$host.ui.SupportsVirtualTerminal` = True
- windows terminal: `$host.ui.SupportsVirtualTerminal` = True

[...]

PR #806, opened 2020-10-27
Avoid try/catch for 'DarkYellow'

dahlbyk:
Error log in https://github.com/dahlbyk/posh-git/issues/771#issuecomment-716622281 pointed out using `'DarkYellow'` ends up polluting `$Errors` with:

```text
Exception calling "FromHtml" with "1" argument(s): "DarkYellow is not a valid value for Int32."
At C:\Dev\OSS\posh-git\src\AnsiUtils.ps1:70 char:17
+                 $color = $ColorTranslatorType::FromHtml($color)
```

Since `DarkYellow` is the only `[ConsoleColor]` that's not also represented in HTML, a hard-coded exception seems appropriate.

cc #536

### DIFF
--- a/src/AnsiUtils.ps1
+++ b/src/AnsiUtils.ps1
@@ -64,6 +64,11 @@ function Get-VirtualTerminalSequence ($color, [int]$offset = 0) {
         return "${AnsiEscape}$(38 + $offset);2;${r};${g};${b}m"
     }
 
+    # Force 'DarkYellow' to ConsoleColor, since it is not an HTML color
+    if ($color -eq [System.ConsoleColor]::DarkYellow) {
+        $color = [System.ConsoleColor]::DarkYellow
+    }
+
     if ($color -is [String]) {
         try {
             if ($ColorTranslatorType) {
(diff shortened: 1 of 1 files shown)


src/AnsiUtils.ps1
```diff
@@ -64,6 +64,11 @@ function Get-VirtualTerminalSequence ($color, [int]$offset = 0) {
         return "${AnsiEscape}$(38 + $offset);2;${r};${g};${b}m"
     }
 
+    # Force 'DarkYellow' to ConsoleColor, since it is not an HTML color
+    if ($color -eq [ConsoleColor]::DarkYellow) {
```
rkeithhill, line 68: Recommend using `[System.ConsoleColor]` here and on the next line. Except for the native types `[string]`, we've tried to specify the full type name so PS doesn't use its "look twice" for the type algorithm.

src/AnsiUtils.ps1
```diff
@@ -64,6 +64,11 @@ function Get-VirtualTerminalSequence ($color, [int]$offset = 0) {
...
+        $color = [ConsoleColor]::DarkYellow
+    }
+
```
rkeithhill, line 71: Might as well join this to the next condition with an `elseif`

rkeithhill (approved):
Minor knits otherwise LGTM.

dahlbyk:
> Minor knits otherwise LGTM.

Better?

rkeithhill:
Yup. LGTM.

PR #807, opened 2020-10-28
Add warning for non-G4W on Windows

dahlbyk:
#771 suggests we don't want folks using `git` from Cygwin. I think we can just check for `.windows` in `git --version`?

Set `$Env:POSHGIT_CYGWIN_WARNING = 'false'` to hide the warning.

image

### DIFF
--- a/src/CheckRequirements.ps1
+++ b/src/CheckRequirements.ps1
@@ -2,6 +2,7 @@ $global:GitMissing = $false
 
 $requiredVersion = [Version]'1.7.2'
 $script:GitVersion = $requiredVersion
+$script:GitCygwin = $false
 
 if (!(Get-Command git -TotalCount 1 -ErrorAction SilentlyContinue)) {
     Write-Warning "git command could not be found. Please create an alias or add it to your PATH."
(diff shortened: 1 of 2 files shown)


src/CheckRequirements.ps1
```diff
@@ -2,15 +2,26 @@ $global:GitMissing = $false
...
-if ([string](git --version 2> $null) -match '(?<ver>\d+(?:\.\d+)+)') {
+if ([string](git --version 2> $null) -match '(?<ver>\d+(?:\.\d+)+)(?<g4w>\.windows)?') {
     $script:GitVersion = [System.Version]$Matches['ver']
+
+    if (($PSVersionTable.PSVersion.Major -le 5) -or $IsWindows) {
+        if (!$Matches['g4w']) {
+            $script:GitCygwin = $true
+
+            if (!$Env:POSHGIT_CYGWIN_WARNING) {
+                Write-Warning "posh-git recommends Git for Windows. You appear to have a different distribution."
```
rkeithhill, line 21: Can we mention why we recommend Git for Windows?  Maybe `... recommends Git for Windows for proper ANSI/VT support. ....` or something like that.

dahlbyk: > You appear to have an unsupported Git distribution; setting $GitPromptSettings.AnsiConsole = $false. posh-git recommends Git for Windows.

rkeithhill (approved):

PR #815, opened 2020-11-21
Allow checkout path expansion without double dash

### LINKED ISSUE #814: git checkout should also autocomplete filepaths without the need for a double dash "--"
The git manual page also states the "--" are optional for git checkout:

> `git checkout [-f|--ours|--theirs|-m|--conflict=<style>] [<tree-ish>] [--] <pathspec>…​`

https://git-scm.com/docs/git-checkout

The path expansion for git reset does also work with and without "--".

dannoe:
Fixes #814

### DIFF
--- a/src/GitTabExpansion.ps1
+++ b/src/GitTabExpansion.ps1
@@ -414,7 +414,7 @@ function GitTabExpansionInternal($lastBlock, $GitStatus = $null) {
         }
 
         # Handles git checkout -- <path>
-        "^checkout.* -- (?<files>\S*)$" {
+        "^checkout.*(?:\s+--)? (?<files>\S*)$" {
             gitCheckoutFiles $GitStatus $matches['files']
         }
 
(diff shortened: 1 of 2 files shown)


rkeithhill:
My one concern here is that folks used to doing `git checkout <tab>` via posh-git might be a bit surprised to see modified files showing up in the completion list.  But if you're intent it to switch branches you could use `git switch` AND it is unlikely you'd switch branches when you have modified files.  @dahlbyk Penny for your thoughts?

dahlbyk:
Appreciate the quick PR to address your issue. I do think we should to avoid changing this behavior for existing users, but would love your input on https://github.com/dahlbyk/posh-git/issues/814#issuecomment-1084100397.

dahlbyk:
Appreciate the quick PR to address your issue. I do think we should to avoid changing this behavior for existing users, but would love your input on https://github.com/dahlbyk/posh-git/issues/814#issuecomment-1084100397.

PR #868, opened 2021-07-23
Fix branch regex to handle symbolic refs

### LINKED ISSUE #867: Tab completion of branches doesn't handle manually created symbolic refs appropriately
### System Details

- posh-git version/path: 1.0.0  ~\Documents\PowerShell\Modules\posh-git\1.0.0
- PowerShell version: 7.1.3
- git version 2.32.0.windows.2
- OS: Microsoft Windows NT 10.0.19042.0

### Issue Description

Tab completion of branches doesn't handle manually created symbolic refs appropriately. The suggested completion for a symbolic ref `symref` that points to `really/long/and/awkward/branchname` is completed as `'symref -> really/long/and/awkward/branchname'`, when it should probably just be `symref`.

### Reproduction

In a git repo

```
# make a branch
git branch really/long/and/awkward/branchname
# point symref to really/long/and/awkward/branchname
git symbolic-ref refs/heads/symref refs/heads/really/long/and/awkward/branchname
# try to tab complete
git switch sym<TAB>
``` 

### Solution

I think this can be fixed with a change to the regex in [`gitBranches` in GitTabExpansion.ps1](https://github.com/dahlbyk/posh-git/blob/b79c2dc39c9387847642bc3b38fa2186b29f6113/src/GitTabExpansion.ps1#L139), either by recognizing the ref with `\S*` instead of `.*` or by recognizing and discarding the ` -> .+` portion as done for the remote branches.

dmringo:
This should fix #867

### DIFF
--- a/src/GitTabExpansion.ps1
+++ b/src/GitTabExpansion.ps1
@@ -136,7 +136,7 @@ function script:gitBranches($filter, $includeHEAD = $false, $prefix = '') {
         $filter = $matches['to']
     }
 
-    $branches = @(git branch --no-color | ForEach-Object { if (($_ -notmatch "^\* \(HEAD detached .+\)$") -and ($_ -match "^[\*\+]?\s*(?<ref>.*)")) { $matches['ref'] } }) +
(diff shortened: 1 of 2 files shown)


dahlbyk:
Very belated thanks!

PR #874, opened 2021-10-01
Fix ignoring .github path

### LINKED ISSUE #873: poshgit doesn't report change values when inside the .github directory
### System Details

- posh-git version/path: 1.0.0  ~\Documents\PowerShell\Modules\posh-git\1.0.0
- PowerShell version: 7.1.4
- git version 2.31.1.windows.1
- OS: Microsoft Windows NT 10.0.19043.0

### Issue Description

I am experiencing a problem with seeing the git status in the prompt when I am in the .github directory.

Here is what I see:
image

jethas-bennettjones:
Closes #873


dahlbyk:
Belated thanks! The substring check was missing an exact match, so I added some tests in #893.

PR #864, opened 2021-07-06
Use `--list` for `branch` and `tag` tab expansion

AndreasHassing:
To improve performance of those tab expansions, which are currently sub-optimal in git repos with a lot of branches.

* for `branch`, `--list` has been available since 1.7.8
  * https://github.com/git/git/blob/ecbdaf0899161c067986e9d9d564586d4b045d62/Documentation/RelNotes/1.7.8.txt#L25
* for `tag`, `--list` has been available since 1.7.10
  * https://github.com/git/git/blob/142430338477d9d1bb25be66267225fb58498d92/Documentation/RelNotes/1.7.10.txt#L128

This is a major performance improvement I've been using locally. Seeing that posh-git has moved to require git v2 or higher, this feature will be in all git versions using posh-git and I thought to share it.

Often times in monolithic repositories at work I used to wait for multiple seconds for branch and tag name completion to succeed, this offloads most of the work to the git engine.

PR #878, opened 2021-12-10
Add a GitHub workflow

dscho:
GitHub workflows provide the fastest and most convenient way to perform Continuous Testing on Windows, macOS and Linux.

Note: currently, there are a few test failures (not introduced by this PR!) that we will have to resolve before this becomes useful.


dscho:
I fear that the test failures ([some of which have been reported before](https://github.com/dahlbyk/posh-git/pull/864#issuecomment-875378865)) are not new, and I fear it is necessary to fix them. However, I am clueless how the `script:`-scoped functions are seemingly no longer visible from the module functions, I am not a PowerShell expert. Help?

dahlbyk:
Thanks for this! I had been meaning to get off Travis. Entering the final week of the course I've been teaching, so expect more attention from me soon.

dahlbyk:
Will troubleshoot failures in a different PR.

dscho:
> Will troubleshoot failures in a different PR.

Thank you! Could you let me know which PR once it is opened?

PR #884, opened 2022-01-10
Added check for GVFS git

### LINKED ISSUE #860: Git with GVFS flagged as unsupported git distribution
### System Details

- posh-git version/path: 1.0.0
- PowerShell version: 7.1.3
- Git version: git version 2.31.0.vfs.0.1
- Operating system name and version: Windows 10

### Issue Description

In Windows development, we use a copy of Git with GVFS enabled, which returns a version "git version 2.31.0.vfs.0.1". This gets flagged as a cygwin git distribution, even though it's not, and fully supports AsciConsole, so shouldn't be flagged as a cygwin version of git.

shmuelie:
Fixes #860

### DIFF
--- a/src/CheckRequirements.ps1
+++ b/src/CheckRequirements.ps1
@@ -8,7 +8,7 @@ if (!(Get-Command git -TotalCount 1 -ErrorAction SilentlyContinue)) {
     return
 }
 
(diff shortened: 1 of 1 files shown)


dahlbyk:
Belated thanks!

Issue #845, opened 2021-03-14
After installing Git "rc" version, PowerShell reports an error


DustinWin:
**Git version:** 2.31.0-rc2 (all rc version)
**Operating system version:** Windows 10 21H1
**PowerShell version:** comes with the system
**oh-my-posh version:** 3.111.2
**posh-git version:** 1.0.0
**Steps to reproduce:**
1. Run PowerShell as an administrator and run the following command:
```
Set-ExecutionPolicy RemoteSigned
Install-Module posh-git
Install-Module oh-my-posh
Import-Module posh-git
Import-Module oh-my-posh
Set-PoshPrompt -Theme PowerLine
```
2. Enter the folder C:\Users\ [your username] \Documents\WindowsPowerShell, edit the text document, and enter the following:
```
Import-Module posh-git
Import-Module oh-my-posh
Set-PoshPrompt -Theme PowerLine
```
And save as Microsoft.PowerShell_profile.ps1
3. Close PowerShell, and open PowerShell again
**Expected results:**
PowerShell will not report errors
**Actual results:**
PowerShell reports an error, and the content of the error:
Warning: You appear to have an unsupported Git distribution; setting $GitPromptSettings.AnsiConsole = $false. posh-git recommends Git for Windows.
![111055583-49ac3600-84b2-11eb-879b-b244d468566f](https://user-images.githubusercontent.com/45238096/111063685-0fab5600-84eb-11eb-99a7-ce611bb9e56f.png)
**Remarks:**
If you install Git v2.30.2 version, no error will be reported. I installed the 2.31.0 rc1, 2.31.0 rc0, and 2.30.0 rc2 versions, and all encountered this problem. I installed the 2.30.1 and 2.30.2 version again and did not encounter this problem. After I reported this problem in Git, I was told that it was a posh-git problem. Please see: https://github.com/git-for-windows/git/issues/3111#issuecomment-798873710

dahlbyk:
The regex use to validate the Git version lives here: https://github.com/dahlbyk/posh-git/blob/5a87e8a196d13395f978d774f80dedb40d43b85e/src/CheckRequirements.ps1#L11

The warning is from past issued with folks using the Cygwin build of Git which doesn't include `.windows` in it's `--version`.

You can set `$Env:POSHGIT_CYGWIN_WARNING = "off"` to disable the warning, or we can try to figure out a better way to identify Git for Windows vs other builds.
[...]

PR #885, opened 2022-01-17
Avoid warning when using -rc versions of Git for Windows

dscho:
This fixes https://github.com/dahlbyk/posh-git/issues/845.


test/CheckRequirements.Tests.ps1
```diff
@@ -0,0 +1,45 @@
+BeforeAll {
+    . $PSScriptRoot\Shared.ps1
+}
+
+Describe 'Test-GitVersion' {
+    InModuleScope 'posh-git' {
+        It 'Returns true for Git for Windows newer than 2.15 (<_>)' -ForEach @(
+            'git version 2.33.0-rc2.windows.1',
```
dahlbyk, line 7: @dscho can you confirm this is the correct version format? Your commit mentions `-rc` but the regex was `\.rc`.

dahlbyk (approved):

test/CheckRequirements.Tests.ps1
dahlbyk: https://github.com/dahlbyk/posh-git/issues/860#issuecomment-846617267 has `git version 2.32.0.rc1.windows.1`. Pushed a commit to support either. 🤷‍♂️

PR #892, opened 2022-03-26
Fix Tests

dahlbyk:
Following on from #878, trying to get back to all green.

- [ ] **Correctly reverts the Window Title back to original state** is unreliable; might just skip
- [x] `Expected: 'igf ' But was:  'git checkout '` seems to be a change in how either Pester or PowerShell handles `-Scope Script`; changing to `-Scope Global` seems to fix. (@dscho you were close!)
- [x] **Tab completes add file in working dir with special char as quoted** suggests `core.quotepath=false` (from b689b5e2974ef8365e16108eb4f7e0edcccfdab6 in #403) no longer affects `git status -sb` as expected - a path with special characters is coming back in `""` regardless. This feels like a regression, but I haven't gone looking for release notes.

### DIFF
--- a/src/GitUtils.ps1
+++ b/src/GitUtils.ps1
@@ -334,20 +334,22 @@ function Get-GitStatus {
                     '^(?<index>[^#])(?<working>.) (?<path1>.*?)(?: -> (?<path2>.*))?$' {
                         if ($sw) { dbg "Status: $_" $sw }
 
+                        $path1 = $matches['path1']
+
                         switch ($matches['index']) {
-                            'A' { $null = $indexAdded.Add($matches['path1']); break }
-                            'M' { $null = $indexModified.Add($matches['path1']); break }
-                            'R' { $null = $indexModified.Add($matches['path1']); break }
-                            'C' { $null = $indexModified.Add($matches['path1']); break }
-                            'D' { $null = $indexDeleted.Add($matches['path1']); break }
-                            'U' { $null = $indexUnmerged.Add($matches['path1']); break }
+                            'A' { $null = $indexAdded.Add($path1); break }
+                            'M' { $null = $indexModified.Add($path1); break }
+                            'R' { $null = $indexModified.Add($path1); break }
+                            'C' { $null = $indexModified.Add($path1); break }
+                            'D' { $null = $indexDeleted.Add($path1); break }
+                            'U' { $null = $indexUnmerged.Add($path1); break }
                         }
                         switch ($matches['working']) {
-                            '?' { $null = $filesAdded.Add($matches['path1']); break }
-                            'A' { $null = $filesAdded.Add($matches['path1']); break }
-                            'M' { $null = $filesModified.Add($matches['path1']); break }
-                            'D' { $null = $filesDeleted.Add($matches['path1']); break }
-                            'U' { $null = $filesUnmerged.Add($matches['path1']); break }
+                            '?' { $null = $filesAdded.Add($path1); break }
+                            'A' { $null = $filesAdded.Add($path1); break }
+                            'M' { $null = $filesModified.Add($path1); break }
+                            'D' { $null = $filesDeleted.Add($path1); break }
+                            'U' { $null = $filesUnmerged.Add($path1); break }
                         }
                         continue
                     }
--- a/src/GitUtils.ps1
+++ b/src/GitUtils.ps1
@@ -336,6 +336,13 @@ function Get-GitStatus {
 
                         $path1 = $matches['path1']
 
+                        # Even with core.quotePath=false, paths with spaces are wrapped in ""
+                        # https://github.com/git/git/commit/dbfdc625a5aad10c47e3ffa446d0b92e341a7b44
+                        # https://github.com/git/git/commit/f3fc4a1b8680c114defd98ce6f2429f8946a5dc1
+                        if ($path1 -like '"*"') {
+                            $path1 = $path1.Substring(1, $path1.Length - 2)
+                        }
+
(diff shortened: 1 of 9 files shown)


dahlbyk:
> * a path with special characters is coming back in `""` regardless

Turns out it's only spaces? https://github.com/git/git/commit/f3fc4a1b8680c114defd98ce6f2429f8946a5dc1

dahlbyk:
> > * a path with special characters is coming back in `""` regardless
> 
> Turns out it's only spaces? [git/git@f3fc4a1](https://github.com/git/git/commit/f3fc4a1b8680c114defd98ce6f2429f8946a5dc1)

I'm genuinely confused how this test ever passed. https://github.com/git/git/commit/dbfdc625a5aad10c47e3ffa446d0b92e341a7b44 is from 2010. 🤔

src/GitUtils.ps1
```diff
@@ -334,20 +334,26 @@ function Get-GitStatus {
                     '^(?<index>[^#])(?<working>.) (?<path1>.*?)(?: -> (?<path2>.*))?$' {
                         if ($sw) { dbg "Status: $_" $sw }
 
+                        $path1 = $matches['path1']
+
+                        if ($path1 -like '"*"') {
+                            $path1 = $path1.Substring(1, $path1.Length - 2)
```
rkeithhill, line 340: A comment here on why the path is being adjusted would other maintainers.  :-)

test/DefaultPrompt.Tests.ps1
```diff
@@ -531,17 +533,12 @@ M test/Baz.Tests.ps1
         Context 'Removing the posh-git module' {
             It 'Correctly reverts the Window Title back to original state' {
                 Set-Item function:\prompt -Value ([Runspace]::DefaultRunspace.InitialSessionState.Commands['prompt']).Definition
+                $originalTitle = & $module Get-Variable OriginalWindowTitle -ValueOnly
```
rkeithhill, line 536: Why not?
```
$originalTitle = & $module { $OriginalWindowTitle }
```

rkeithhill: But do it after importing posh-git

test/DefaultPrompt.Tests.ps1
```diff
@@ -531,17 +533,12 @@ M test/Baz.Tests.ps1
...
-                if ($Host.Name -eq 'RemoteHostImplementation') {
-                    $title | Should -eq $originalTitle
-                }
-                elseif ($PSVersionTable.PSVersion.Major -lt 6) {
-                    $title | Should -match '^Windows PowerShell|:\\Windows\\System32\\WindowsPowerShell\\v1.0\\powershell.exe$'
-                }
-                else {
-                    $title | Should -match '^(Administrator: )?(Windows )?PowerShell'
-                }
+                $title | Should -eq $originalTitle
```
rkeithhill, line 541: Yeah, that's better.  😁

test/TabExpansion.Tests.ps1
```diff
@@ -357,7 +358,7 @@ Describe 'TabExpansion Tests' {
             $result | Should -BeExactly "'v2.0.0'''"
         }
         It 'Tab completes add file in working dir with special char as quoted' {
-            $filename = 'foo{bar} (x86).txt';
+            $filename = 'foo{bar} ��(x86).txt';
```
rkeithhill, line 361: What's with the unrecognized chars?  Is that part of the "special char" test?  Might be worth adding as another test for emoji chars or whatever that is.  IIRC this test was meant to check for tab completion with PowerShell and/or regex special chars/syntax e.g. `{}` and `()` in the completed text.

rkeithhill (approved):
Just a few suggestions otherwise LGTM

test/DefaultPrompt.Tests.ps1
rkeithhill: Ah, now I see you're capturing this variable in the test fixture before the test runs.

test/TabExpansion.Tests.ps1
dahlbyk: Was just being lazy, but yeah the emoji is to ensure we're setting `core.quotePath` (and it's working as expected). Will break out as a separate test.

test/DefaultPrompt.Tests.ps1
dahlbyk: > Why not?
> 
> ```
> $originalTitle = & $module { $OriginalWindowTitle }
> ```

I don't use this syntax often, so I refer to https://github.com/dahlbyk/posh-git/blob/5a87e8a196d13395f978d774f80dedb40d43b85e/src/Utils.ps1#L7

test/TabExpansion.Tests.ps1
dahlbyk: > > What's with the unrecognized chars? Is that part of the "special char" test? Might be worth adding as another test for emoji/unprintable chars or whatever that is. IIRC this test was meant to check for tab completion with PowerShell and/or regex special chars/syntax e.g. `{}` and `()` in the completed text.
>
> Was just being lazy, but yeah the emoji is to ensure we're setting `core.quotePath` (and it's working as expected). Will break out as a separate test.

Turns out we already had a test for this: https://github.com/dahlbyk/posh-git/blob/a62f90cb1ff1b66bc05da5787eb561bce64e3da6/test/TabExpansion.Tests.ps1#L222-L226

Relocated into the appropriate `Context`.

dahlbyk (commented):
Thanks for the review, @rkeithhill

PR #905, opened 2022-04-27
Fix tests: Modifying git init initial branch to always be master.

cdonnellytx:
If `init.defaultBranch` is set (2.28.0 or later), then test setup explicitly initializes the repo with "master" as the default branch.
This argument will not be added to the command unless `init.defaultBranch` is set.

This ensures the tests always use "master" and so aren't affected by that setting.


rkeithhill (approved):
LGTM. Thanks!

PR #906, opened 2022-05-23
refactor: exclude PSUseSingluarNouns

mcdonaldjc:
I became irritated by this rule showing up for functions like New-GitPromptSettings. Instead of refactoring all similar methods when the noun seems just fine or refactoring so that each function implements ShouldProcess, I think it best just to suppress the rule.


dahlbyk:
Good idea - thanks!

PR #916, opened 2022-07-24
Tab completion for partial clones

bgold09:
* Tab completion for `clone` parameters that support partial clones.
   * `--filter=<filter spec>` for "blobless" and "treeless" clones.
   * `--sparse` to employ a sparse-checkout, with only files in the 
      top-level directory initially being present.


bgold09:
@dahlbyk would you mind reviewing this?

PR #945, opened 2023-02-13
VS Code tasks improvments

Ra0to:
Fixed error with Pester tests launch.  
```
The property 'WriteVSCodeMarker' cannot be found on this object. Verify that the property exists and can be set.
```

Added new task for running tests only in current file.


7assq:
Kkdk

dahlbyk (approved):
LGTM, thanks!

PR #981, opened 2024-12-26
Tab completion if dashed parameter for git is used

dodmi:
This fixes tab completion if a dashed parameter directly after git command is used, e.g. `git -P branch`


ali0935nim (commented):
5a5cedb2010b5110be3fe35f97e6f8a54d8c4292https://github.com/dahlbyk/posh-git/blob/master/README.md#customization-variables

